PR #1582, opened 2023-06-16
Number input

benjaminleonard:
Uses [`useNumberField`](https://react-spectrum.adobe.com/react-aria/useNumberField.html) from `react-aria` and adds a `NumberField` for the `react-hook-form` hookup.

Has a bunch of useful features like steps, min and max values, units and keyboard input.

CleanShot 2023-10-05 at 15 03 34


benjaminleonard:
Struggling with the types of `defaultValue` and `onChange` within `NumberField` if you get a second to pitch in @david-crespo

PR #1741, opened 2023-08-17
Update `<Modal.Docs>` for consistency

benjaminleonard:
Constraining the input of the `<Modal.Docs>` links seems a bit neater. Updated the technical preview banner to use this along with some small visual tweaks.

image


david-crespo:
Looks nice. Might need the title customizable — they're repos, not docs. I also don't like the array API — I'd rather have a component you call for each link as children of the wrapper. Those two changes go together though — could split the monolithic component into a wrapper with a configurable title and a child component you call three times.

benjaminleonard:
Sounds good. Should we scope that to the modal, something like `<Modal.Link>` to distinguish from the other link component?

david-crespo:
Yeah, I think that makes sense.

PR #1750, opened 2023-09-05
Add loading state to ok button in confirm delete modal

### LINKED ISSUE #1748: Instance delete spinner
Instance delete can take a non-negligible amount of time (~5s) on Rack3. When clicking the "Confirm" button on the delete modal there should probably be some kind of loading spinner on the modal to confirm that the delete is occurring. Currently nothing happens and my immediate reaction was that the confirm button was not working.

This should probably be a general pattern, as presumably there are other resources that require non-zero time to delete.

david-crespo:
Closes #1748. This covers all deletes, not just instance create.

2023-09-05-confirm-delete-spinner

### DIFF
--- a/libs/ui/lib/modal/Modal.tsx
+++ b/libs/ui/lib/modal/Modal.tsx
@@ -118,6 +118,7 @@ Modal.Footer = ({
   onAction,
   actionType = 'primary',
   actionText,
+  actionLoading,
   cancelText,
   disabled = false,
 }: {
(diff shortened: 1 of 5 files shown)


libs/ui/lib/modal/Modal.tsx
```diff
@@ -126,6 +127,7 @@ Modal.Footer = ({
   onAction: () => void
   actionType?: 'primary' | 'danger'
   actionText: React.ReactNode
+  actionLoading?: boolean
   cancelText?: string
   disabled?: boolean
```
david-crespo, line 132: Adding even more props here to pass through to the children is a clear sign this should be refactored into a compositional API.

PR #1753, opened 2023-09-05
Try to reduce safari E2E flake

david-crespo:
make e2e image upload take longer, add another retry


david-crespo:
WELL THAT DIDN'T WORK

PR #1755, opened 2023-09-06
Fix image upload modal cancel loop

### LINKED ISSUE #1745: Cancel modal loop
Encountered a loop when clicking off of the image upload modal. When the right hand upload image form it seems that attempting to cancel out of the alert will retrigger it again. Clicks on other areas of the screen do not seem to exhibit this behavior.

[Screencast from 2023-08-30 21-47-53.webm](https://github.com/oxidecomputer/console/assets/1377730/1ed8573a-39c9-473d-9ba3-3bd9c7dd2ef2)

(looks like cursor didn't get captured, I can recapture if needed)

Running omicron commit ab3591753bc2546bab6f666985dd9e26afa7d3dd

david-crespo:
Fixes #1745. Will make a separate issue for the remaining part of the problem.

I believe the issue is due to both the click outside and focus outside triggering `onDismiss`, which triggers the `confirm()`. My theory so far is that when you cancel out of the confirm, focus returns to the dismissable layer overlay, which triggers `onFocusOutside` again.

This change fixes the infinite loop of confirms that fully breaks the page, but it does not fix the double confirm when you click elsewhere (this is shown at the end of the recording). Another issue shown in the recording is that when you click the background over the inputs in the side modal, it seems to focus them. This may be directly caused by the `preventDefault`. In any case, it's an improvement on the current broken state.

2023-09-06-modal-loop-fix


david-crespo:
I need a break from Safari failing my E2Es before I throw my computer out the window. But I will fix it.

PR #1757, opened 2023-09-07
Try running E2E against Vercel

david-crespo:


david-crespo:
Interesting, the tests that use a cookie to pose as a less-privileged user didn't work.

PR #1758, opened 2023-09-07
Tweak playwright config

david-crespo:
We were actually overriding the config with `--workers=2` in the command in the workflow config, but I prefer to centralize that in the config. We now tell it to use `'100%'` of available cores ([which is 2 as of right now](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners)).


david-crespo:
Interesting! With worker count unspecified, it used 1 and took 8 minutes instead of 5. Turns out the default is 50% of whatever is there.

PR #1765, opened 2023-10-04
Move to `@oxide/design-system` icons

### LINKED ISSUE #1654: Metrics pages should use metrics icon in header and sidebar
Needs https://github.com/oxidecomputer/design-system/issues/35

Sidebar links should also use `Metrics16Icon` but we should change them all together so they don't mismatch.

benjaminleonard:
The design system now exports icons in any format that a repo might use; React components, SVGs, spritesheet so we no longer need to have a build process within the console – and it benefits us to have a single source of truth so this PR imports from the design system instead.

The default index generator from SVGR exports an icon in the following format: `export { default as Close8Icon } from './Close8Icon.tsx';`. The type checker is displeased with this because it includes the file extension. The easiest solution is to set `allowImportingTsExtensions` to true in the `tsConfig`.

Also switching to the new metrics icon on the utilization pages. Fixes #1654


david-crespo:
Very cool, love to get these icons out of here. Seems like Vite is pissed off about the imports, though. Blank screen and 

```
Uncaught ReferenceError: can't access lexical declaration 'Zr' before initialization
```

in console. When I click the sourcemap I see this in `Message.tsx`:

```ts
import { Error12Icon, Success12Icon, Warning12Icon } from '../../'
```

https://console-j93an1226-oxidecomputer.vercel.app/

david-crespo:
The error reproduces in local dev so it should be pretty easy to figure out. I'll take a look.

david-crespo (approved):
GOODBYE CODE

PR #1772, opened 2023-10-05
Scaling utilization Y axis

### LINKED ISSUE #1640: Using rack capacity for system utilization max y is bad when stats are small
I think dynamic max would be fine but then we cap it at the rack capacity.

image

benjaminleonard:
Fixes #1640 

There's a few small changes in here.

- We use the largest data point +20% for the graph scale clamping at `maxValue` (if set) which is usually overall capacity
- Rounding the `calculatedMaxValue` to be divisible by 6 so we get even ticks
- Normalizing the mock metrics data to be within max capacity (previously the walker would exceed 100%)

On the `maxValue` rounding – we could do it the other way round where we find a number of ticks that fits but it might be confusing to have it change between graphs. The uneven tick lines was most obvious on disk space because we are using TiB.

### DIFF
--- a/app/components/TimeSeriesChart.tsx
+++ b/app/components/TimeSeriesChart.tsx
@@ -100,7 +100,7 @@ function renderTooltip(props: TooltipProps<number, string>, unit?: string) {
   )
 }
 
-type Props = {
+type TimeSeriesChartProps = {
   className?: string
   data: ChartDatum[] | undefined
   title: string
@@ -109,8 +109,14 @@ type Props = {
   interpolation?: 'linear' | 'stepAfter'
   startTime: Date
   endTime: Date
-  maxValue?: number
   unit?: string
+}
+
+const TICK_COUNT = 6
+
+/** Round `value` up to nearest number divisible by `divisor` */
+function roundUpToDivBy(value: number, divisor: number) {
+  return Math.ceil(value / divisor) * divisor
 }
 
 export default function TimeSeriesChart({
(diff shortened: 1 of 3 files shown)


app/components/TimeSeriesChart.tsx
```diff
@@ -125,13 +125,26 @@ export default function TimeSeriesChart({
   maxValue,
   unit,
 }: Props) {
+  // We use the largest data point +20% for the graph scale
+  // clamping at `maxValue` (if set) which is usually overall capacity
+  const calculatedMaxValue = useMemo(() => {
+    if (!maxValue) return null
```
david-crespo, line 131: Don't you still want a reasonable max on the chart (like `dataMax * 1.2`) if none is specified from outside?

app/components/TimeSeriesChart.tsx
```diff
@@ -125,13 +125,26 @@ export default function TimeSeriesChart({
...
+    if (!rawData) return maxValue
+    const dataMax = Math.max(...rawData.map((datum) => datum.value))
+    const unroundedValue = Math.min(maxValue, dataMax * 1.2)
```
david-crespo, line 134: As long as we don't have the capacity number nailed down, I'm wary of picking `maxValue` as the max if `dataMax` is actually bigger. In the unfortunate case where `dataMax` is bigger than `maxValue`, we have to set the chart max to at least `dataMax`, otherwise you're not seeing all the data. Ideally in that situation we would also show a dotted flat line or something at `maxValue` too, though that can be a separate change. We'd have to explain it somehow too.

I came up with this after banging my head against this for bit. Does it make any sense? 

|Situation| Exact range | Desired max |
|---|---|---|
| Data way less than cap | `dataMax * 1.2 < maxValue` | `dataMax * 1.2` |
| Data close to cap but still under | `dataMax < maxValue < dataMax * 1.2 ` | `maxValue` |
| Data over cap (pathological but possible) | `maxValue < dataMax` | `dataMax * 1.2` |

The middle one is the odd one out. I think that's what's messing me up. What if we _always_ used `dataMax * 1.2` as the top edge of the chart, and then if `maxValue` happens to be inside the range of the chart, we show it as a dotted line? Or maybe not always `dataMax * 1.2` — maybe if `dataMax` is anywhere near `maxValue`, like at least half of it, we make it so the `maxValue` line is shown.

benjaminleonard: That sounds about right for me. Setting the top of the graph to the capacity doesn't really clearly communicate that if it's not being applied regardless of usage. So some other indicator is probably best.

Whilst the max capacity is still fuzzy, we could probably just omit it and the logic around keeping the `maxValue` line visible. Happy to just change to `dataMax * 1.2` for now.

david-crespo: I think that would work. We'd scrap `maxValue` altogether in that case, right?

benjaminleonard: For now yeah, we can revisit when we have updated capacity values. Looks like we can use Recharts reference lines when thats ready: https://recharts.org/en-US/examples/LineChartWithReferenceLines

app/components/TimeSeriesChart.tsx
```diff
@@ -122,20 +128,23 @@ export default function TimeSeriesChart({
...
   // is the maximum, we also use our own function as recharts
-  // doesn't fill the whole domain (just upto the data max)
-  const yTicks = maxValue
-    ? {
-        domain: [0, maxValue],
-        ticks: getVerticalTicks(6, maxValue),
-      }
-    : {
-        tickSize: 6,
-      }
```
david-crespo, line 138: I got rid of `tickSize` here — I thought it was weird to only specify it when there's a max, but also [6 is the default anyway](https://recharts.org/en-US/api/YAxis#tickSize) so we're not doing anything.

Issue #1211, opened 2022-10-06
Snapshot create confirmation/feedback


david-crespo:
> Really there are two issues, I think. In practice the snapshotting will not be instant (you should probably add a couple of seconds delay to the mock endpoint), so there's the feedback on completion, but I also think there's a general "ok what is happening here, what is being created, what is it called, where does it end up?" that might not be answered by the completion confirmation. Or at least, even if we do give some of that info at completion time, and make it easy, for example, to click through to wherever the snapshot lives, I think the user wants some of that info _before_ they click the button so they can feel confident it's what they want to do.

_Originally posted by @david-crespo in https://github.com/oxidecomputer/console/pull/1204#discussion_r989212083_

askfongjojo:
I think even an acknowledgement of the request made (including in the message the name of the snapshot being created) will be very helpful. Besides a better user experience, it'll likely help reduce the number of unnecessary retries when user thinks the request has not been received.  Currently, repeated snapshot requests may cause sagas to fail as explained in RFD 419.

david-crespo:
Plopping an extra toast in there when we start is very easy but also terrible. @benjaminleonard @paryhin help!

```diff
@@ -95,6 +95,7 @@ export function DisksPage() {
     {
       label: 'Snapshot',
       onActivate() {
+        addToast({ content: 'Creating snapshot', variant: 'info' })
         createSnapshot.mutate({
           query: projectSelector,
           body: {
```

![2023-10-05-creating-snapshot-toast](https://github.com/oxidecomputer/console/assets/3612203/85cf3009-bdb3-4ae0-968e-d321adc03418)

david-crespo:
Couple of thoughts after doing #1774. One, obviously whatever we do here should apply to both disks list and instance disks list. Two, I think there are three different stages here that we can give the user feedback about. There's completion, which we already have, but before that there is confirmation prior to initiating the snapshot, _and_ there is a loading/in progress state indicating that the snapshot is happening. I could go either way on whether we need a confirmation modal — I think the bigger problem is that there's no feedback that anything is happening. But I do lean toward confirmation. It could be a little form where we show the generated name and give them a chance to change it, for example.

benjaminleonard:
[...]

PR #1774, opened 2023-10-05
Add snapshot action to instance disks list

### LINKED ISSUE #1771: Create snapshot from instance disks list
I ran into this while using the colo rack. When I'm working with an instance, I think of its disks primarily through the view on the instance detail page's storage tab. It was pretty surprising that I would have to go to the project disks list in order to make a snapshot, and it took me a minute to figure out what to do. And I work here!

image

david-crespo:
Closes #1771 

Confirmation toast continues to be unimpressive. That will be addressed with #1211.

2023-10-05-instance-disk-snapshot

### DIFF
--- a/app/test/e2e/instance/disks.e2e.ts
+++ b/app/test/e2e/instance/disks.e2e.ts
@@ -5,7 +5,15 @@
  *
  * Copyright Oxide Computer Company
  */
-import { expect, expectNotVisible, expectVisible, stopInstance, test } from '../utils'
+import {
+  clickRowAction,
+  expect,
+  expectNotVisible,
+  expectRowVisible,
+  expectVisible,
+  stopInstance,
+  test,
+} from '../utils'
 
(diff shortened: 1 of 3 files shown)


app/test/e2e/instance/disks.e2e.ts
```diff
@@ -57,3 +65,23 @@ test('Attach disk', async ({ page }) => {
...
+
+  await clickRowAction(page, 'disk-2', 'Snapshot')
+
+  await expect(successMsg).toBeVisible() // we see the toast!
+
+  // now go see the snapshot on the snapshots page
+  await page.getByRole('link', { name: 'Snapshots' }).click()
+  const table = page.getByRole('table')
+  await expectRowVisible(table, {
+    name: expect.stringMatching(/^disk-2-/),
```
david-crespo, line 84: had to add this to be able to match on a partial cell content

Issue #1777, opened 2023-10-06
Image name looks bad when OS and/or version are empty


david-crespo:
![Screenshot 2023-10-06 at 1 46 00 PM](https://github.com/oxidecomputer/console/assets/3612203/6678fd30-a4c0-4d05-84e5-c2000de786b2)
![Screenshot 2023-10-06 at 1 45 49 PM](https://github.com/oxidecomputer/console/assets/3612203/4a4af8f1-d164-4376-b548-e4cfd4e8855f)

benjaminleonard:
We should figure out a graceful solution for this, but I think including the OS and version is decent UX improvement over the label alone. It reduces the pressure on a really good naming convention on the label to find the image you are looking for. We are also working with a very small number of images currently, and I think we'd find it much more useful if we had to filter and search (when the feature exists) to find the image you are looking for.

I'm open on requiring OS and version. It's feasible let's say when you're creating an image from a snapshot that you not know that information.

Issue #1775, opened 2023-10-06
Copy button on instance external IP


david-crespo:
It's the only thing I want to do on that page! Might even be worth adding to the main details table for the instance.

neeleshbisht99:
Hi David,
Creating another PR for the Copy icon in the system-design repo.

PR #1782, opened 2023-10-10
Added instance ip copy button

neeleshbisht99:
Fixes [#1777](https://github.com/oxidecomputer/console/issues/1775)

### DIFF
--- a/app/pages/project/instances/instance/tabs/NetworkingTab.tsx
+++ b/app/pages/project/instances/instance/tabs/NetworkingTab.tsx
@@ -23,10 +23,12 @@ import { useQueryTable } from '@oxide/table'
 import {
   Badge,
   Button,
+  Clipboard16Icon,
   EmptyMessage,
   Networking24Icon,
   Spinner,
   Success12Icon,
+  useTimeout,
 } from '@oxide/ui'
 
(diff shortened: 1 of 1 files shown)


benjaminleonard:
Thanks for the PR Neelesh – but I'd like to hold off on this until we figure out exactly how we want to handle inline copying of table values. It's a pattern we'll likely re-use in a few places so I want to approach this holistically. Or if indeed a copy button is the right solution, as this is an IP, we might want to just wrap it in an anchor tag and the use can simply right click -> copy link address within their browser.

david-crespo:
Thanks Neelesh. We're going to close this for now while we figure it out.

neeleshbisht99:
Hi @david-crespo @benjaminleonard , thanks for the heads up. 
Can you point me to another issue where I can be helpful?

david-crespo:
Unfortunately, not really. As we say in [CONTRIBUTING.md](https://github.com/oxidecomputer/console/blob/main/CONTRIBUTING.md), we're a very small team and it's hard enough to coordinate and prioritize work amongst ourselves even with the full context we have from working here. Issues labeled [`bug`](https://github.com/oxidecomputer/console/issues?q=is%3Aissue+is%3Aopen+label%3Abug) are an ok starting point, though again, I'd caution you that something like the above could easily happen with those.

Issue #1289, opened 2022-11-07
Don't validate on blur if field is not dirty


david-crespo:
This validation is a little aggro.

![2022-11-07-aggro-validation](https://user-images.githubusercontent.com/3612203/200353735-b78e091a-6e9f-463a-8139-a4acf0cc2b9f.gif)

You shouldn't get screamed at if all you do is blur the field (e.g., to look at the rest of the form if the form is longer). My first thought is that we can simply not validate on blur until the field (or the form?) is empty. I kind of doubt we can implement this on the form itself using [`mode`](https://react-hook-form.com/api/useform) because the options are pretty flat — submit, blur, change, touched, all. However, it would be pretty do at the _display_ layer: regardless of whether there is an error on the field, only show it under certain conditions, which we can pull from the field and form state.

https://github.com/oxidecomputer/console/blob/d41e40f13ebb13f51f28a17b80307747402b22ab/app/components/form/fields/TextField.tsx#L83-L100

david-crespo:
I'm not sure there's a nice way to do this, so it might be prudent to decide we can live with it.

PR #1784, opened 2023-10-10
Add TLS Cert input

### LINKED ISSUE #1736: Add TLS certs to silo create form
There is an optional field in the POST body that takes a list of certs. We should add file inputs for that.

benjaminleonard:
Fixes #1736 

Uses the mini table and modal create pattern to add TLS certificates. I've omitted `service` as there's only one option currently `external_api`.

As specced in the API, `key` and `cert` are plain text — added `readBlobAsText` to read that from the File object.

tls-cert

Running into a couple of nits around the required validation. If you try to click cancel after opening the modal, the first click will be interrupted by the error on the "Name" field. Similarly, when you click on the file upload you immediately get the required error message. Should probably handle in a separate PR.


app/components/form/index.ts
```diff
@@ -22,3 +22,5 @@ export * from './fields/NetworkInterfaceField'
 export * from './fields/RadioField'
 export * from './fields/SubnetListbox'
 export * from './fields/TextField'
+export * from './fields/TlsCertsField'
+export * from './fields/FileField'
```
benjaminleonard, line 26: We had most the fields already in here, so I migrated the remainder over.

app/components/form/fields/TlsCertsField.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+      {showAddCert && (
+        <AddCertModal
+          onDismiss={() => setShowAddCert(false)}
+          onSubmit={async (values) => {
+            const keypair = {
+              cert: await readBlobAsText(values.cert),
+              key: await readBlobAsText(values.key),
+            }
+            onChange([...items, { ...values, ...keypair }])
```
benjaminleonard, line 74: `onChange` accepts any here, do we want to do `[...items. {...values, ...keypair}] as TlsCertificate[]` to get a typecheck on this.

app/components/form/fields/TlsCertsField.tsx
```diff
@@ -0,0 +1,137 @@
...
+              {/* For remove button */}
+              <MiniTable.HeadCell className="w-12" />
+            </MiniTable.Header>
+            <MiniTable.Body>
+              {items.map((item, index) => (
+                <MiniTable.Row
+                  tabIndex={0}
+                  aria-rowindex={index + 1}
+                  aria-label={`Name: ${item.name}, Description: ${item.description}`}
+                  key={item.name}
```
benjaminleonard, line 45: Currently we aren't checking for uniqueness around name. Do we want to do so? If not, we should use something like `${item.name}-${index}` to ensure the keys do not clash.

benjaminleonard:
Validated working against dogfood earlier today.

app/components/form/fields/TlsCertsField.tsx
david-crespo: I bet there’s a way to get a better type on `onChange`, will try it later.

app/components/form/fields/TlsCertsField.tsx
david-crespo: Yeah, we probably should enforce uniqueness. I guess we’d check that on submit in the modal? Refuse to submit if there’s a duplicate.

app/components/form/fields/TlsCertsField.tsx
```diff
@@ -0,0 +1,137 @@
...
+          Add TLS certificate
+        </Button>
+      </div>
+
+      {showAddCert && (
+        <AddCertModal
+          onDismiss={() => setShowAddCert(false)}
+          onSubmit={async (values) => {
+            const keypair = {
+              cert: await readBlobAsText(values.cert),
```
david-crespo, line 71: turns out you should be able to do `await values.cert.text()`

https://developer.mozilla.org/en-US/docs/Web/API/Blob/text

david-crespo: Confirmed that works, that's fun. I also noticed while testing that if the file dropped on the input has newlines, `readBlobAsText` seems to only get the last line.

app/components/form/fields/TlsCertsField.tsx
david-crespo: Or not. I tried.

```ts
const certCreate: (typeof items)[number] = {
  ...values,
  cert: await values.cert.text(),
  key: await values.key.text(),
}
onChange([...items, certCreate])
```

Nothing I tried felt perfect, but this feels pretty good — only assert that the value matches the items in `items`. This is very close to your idea, except 

a) `as` is a cast, so it doesn't really check. `satisfies` would work better in that spot, and
b) because we're inferring from the type of `items`, the type of `items` can change and we will only see an error if `certCreate` doesn't match it

app/components/form/fields/TlsCertsField.tsx
```diff
@@ -0,0 +1,172 @@
...
+  allNames,
+  ...textFieldProps
+}: Omit<TextFieldProps<TFieldValues, TName>, 'validate'> & {
+  label?: string
+  allNames: string[]
+}) {
+  return (
+    <TextField
+      validate={(name) =>
+        allNames.includes(name) ? 'Certificate with this name already exists' : true
```
benjaminleonard, line 120: I did attempt to get this going in the `onSubmit` of `AddCertModal` but for whatever reason nothing I put in there was firing. Guessing `react-hook-form` is intercepting the default `onSubmit`.

But the benefit of this approach is we don't need to mess around with `setError` or anything like that. I did consider memoizing `allNames` but figured it might be overboard for the number of certs we'll have here.

libs/ui/lib/file-input/FileInput.tsx
```diff
@@ -62,6 +62,10 @@ export const FileInput = forwardRef(
           onDragEnter={() => setDragOver(true)}
           onDragLeave={() => setDragOver(false)}
           onDrop={() => setDragOver(false)}
+          /* Validating this onBlur causes the required 
+              error message to appear on click. Instead we can
+              validate on form submit by overriding onBlur */
+          onBlur={() => {}}
```
benjaminleonard, line 68: Any thoughts on this @david-crespo? We might want to validate `onChange` if we have anything like validating filenames or sizes – but it might give you an error when you clear the file, which we may or may not want. Presently we are only really validating required or not.

This fixes the issue where the error appears as soon as you click on the file input.

david-crespo: As discussed in chat, I think validate on blur only if dirty (#1289) would be our ideal behavior, though I don't think RHF supports it yet. I will look into patching or PRing it so we can test that out. On the other hand, I don't think that would solve #1786 because you could still trigger by typing something invalid in the name field and then trying to change the radio input.

benjaminleonard: Yeah – though this specifically is targeting the error appearing when you click on the file input. It seems like the `onBlur` is triggered whenever the file input opens.

david-crespo:
Merging main with #1787 fixed the issues with the file input error on blur, plus the one (same as #1786) where if you scroll down and click the add cert button without typing in the name field, you'd get a validation error on name due to the blur that prevents your click on the button from going through (probably because it gets scrolled out of the way, ugh).

I noticed name was not being validated as a required field, fixing that.

app/components/form/fields/TextField.tsx
```diff
@@ -53,7 +53,6 @@ export interface TextFieldProps<
   description?: string
   placeholder?: string
   units?: string
-  // TODO: think about this doozy of a type
```
david-crespo, line 56: one year of thinking is enough

app/util/validate.ts
```diff
@@ -1,23 +0,0 @@
...
-  if (value.length === 0) {
-    return 'A name is required'
-  } else if (!/^[a-z]/.test(value)) {
-    return 'Must start with a lower-case letter'
-  } else if (!/[a-z0-9]$/.test(value)) {
-    return 'Must end with a letter or number'
-  } else if (!/^[a-z0-9-]+$/.test(value)) {
-    return 'Can only contain lower-case letters, numbers, and dashes'
-  }
-}
```
david-crespo, line 23: this wasn't used. apparently a leftover from when we switched from Formik to RHF

david-crespo (approved):
This looks good to me. I want to give the code and the E2Es another once-over later and then merge.

Something we should follow up on: displaying the certs on the silo detail page where we currently list IdPs. I'm a little confused about the API for this, will need to look into it. I don't see a way to list these certs.

PR #1795, opened 2023-10-18
Silo utilization table

benjaminleonard:
Presently a user can filter by silo to see its individual utilization. There presently is no real way to easily compare metrics across silos. This is partly due to the API not returning this aggregated data. In the short-term we can stub this out by batching a whole bunch of requests and constructing the table from the merged results – this might also give us some guidance on the shape of the API that we need.

We're working on an idealised design, including quotas – but the details of this are still being worked on, and this PR has the simplest implementation possible right now.

This adds a new hook `useApiQueries` which is our version of `useQueries`, but with the key difference that all queries in a given call are using the same API method, and therefore all have the same request and response (`Params` and `Result`) types.


david-crespo:
This is looking pretty good. Couple of things I'm still thinking about:

* Could take "Capacity" out of the page title, just call it Utilization
* Tab names of `Metrics` and `Usage` are not very informative, but we had a hard time coming up with more concrete names
* "Usage" in general I don't like because it's the same as Utilization with no special connotation
* I'll add an e2e test. should actually be pretty easy — will need to use regexes for the table values because the data is randomized
* Loading state for the table

paryhin:
CleanShot 2023-10-20 at 12 08 00@2x

Looks good, a couple of things that caught my attention:

1. It took me a while to find the table switch, even though I'm familiar with the UI 🙂. We should make the tabs more visible. 
2.  On wider screens, the chart and related data seem too far apart, and based on the proximity of elements is looks like it is grouped with the next value, not the actual one on the left.

It'd be great to address these for future improvements.

david-crespo:
Current (and possibly final) state.

2023-10-23-utilization-table

PR #1797, opened 2023-10-19
Delete button on silo images table

### LINKED ISSUE #1796: Add image delete button to silo images page
We have it on project images but not silo images. It seems from https://github.com/oxidecomputer/console/pull/1574 that this decision was based on the belief (possibly correct at the time) that project images had delete working but silo images didn't. As of https://github.com/oxidecomputer/omicron/pull/4296 it should be working for both (and before that it was not implemented for either one).

There is a permissions question we should think about — silo images require stronger permissions to delete. In general we have not put much effort yet into conditional UI based on the user's roles, and that's probably smart because the actual role required for something may be a result of fairly complex logic on the backend that it would be risky and error-prone to try to recreate. So I am inclined for now to put the button in unconditionally and just make sure we're handling the 403 nicely.

david-crespo:
Closes #1796 

- Add delete button to silo images
- Remove `onError` on project image delete to fix double toast (delete confirm handles error toast for us)
- Make user role check in mock server more elaborate — let us check for any role on any resource, not just fleet viewer

image

### DIFF
--- a/libs/api-mocks/msw/util.ts
+++ b/libs/api-mocks/msw/util.ts
@@ -10,16 +10,18 @@ import type { RestRequest } from 'msw'
 
 import type {
   DiskCreate,
+  RoleKey,
   Sled,
   SystemMetricName,
   SystemMetricQueryParams,
   User,
 } from '@oxide/api'
-import { MAX_DISK_SIZE_GiB, MIN_DISK_SIZE_GiB, totalCapacity } from '@oxide/api'
+import { FLEET_ID, MAX_DISK_SIZE_GiB, MIN_DISK_SIZE_GiB, totalCapacity } from '@oxide/api'
 import type { Json } from '@oxide/gen/msw-handlers'
 import { json } from '@oxide/gen/msw-handlers'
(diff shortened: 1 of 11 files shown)


libs/api-mocks/msw/util.ts
```diff
@@ -292,24 +294,45 @@ export function currentUser(req: RestRequest): Json<User> {
...
-  const actorsWithFleetRole = db.roleAssignments
-    .filter((ra) => ra.resource_type === 'fleet')
+  /** All actors with *at least* the specified role on the resource */
+  const actorsWithRole = db.roleAssignments
+    .filter(
+      (ra) =>
+        ra.resource_type === resourceType &&
+        ra.resource_id === resourceId &&
+        roleOrStronger[role].includes(ra.role_name)
+    )
```
david-crespo, line 331: Honestly this was very easy. Not sure why I didn't do this before?

Well.... this does not account for parent-child relationships, like if you have collab on silo you also have it on all projects in silo. I need to either account for that or add a comment about the limitation.

PR #1802, opened 2023-10-25
Fix sled instances prefetch

david-crespo:
I noticed on rack3 that when navigating from sleds list to sled detail (instances tab), there was a delay before the instances table would show up. We were prefetching it, so I figured out it's because the sled detail link goes to `/sleds/abc123`, and then we do a little client-side redirect from there to `/sleds/abc123/instances`, and the loader is on the latter. Strictly speaking I think this might be a bug in React Router, but in the meantime we can fix this either by having the link from the sled list go directly to `/sleds/abc123/instances` or by adding the loader to that redirect route (as I've done here). I vaguely remember there being a problem with this when I tried it before on another page, so I'll leave this up while we mull it over.

### DIFF
--- a/app/layouts/SystemLayout.tsx
+++ b/app/layouts/SystemLayout.tsx
@@ -65,7 +65,7 @@ export default function SystemLayout() {
     const systemLinks = [
       { value: 'Silos', path: pb.silos() },
       { value: 'Utilization', path: pb.systemUtilization() },
-      { value: 'Inventory', path: pb.sledInventory() },
+      { value: 'Inventory', path: pb.inventory() },
     ]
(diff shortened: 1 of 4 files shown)


david-crespo:
The problems discussed in the issues below don't seem to be a problem anymore, so I'm guessing they fixed the race conditions that were causing them. Of course it's always possible they're still there, but more likely to go how we prefer. In the short term I think that's ok, but eventually I'd like to understand why it was fixed.

https://github.com/oxidecomputer/console/pull/1267#discussion_r1014486191
https://github.com/oxidecomputer/console/pull/1125

PR #1805, opened 2023-10-26
typescript-eslint v6 + prettier 3.0

### LINKED ISSUE #1789: Upgrade typescript-eslint and rework config
Latest release has significant changes.

https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/#as-an-existing-user

david-crespo:
Closes #1789 

https://typescript-eslint.io/blog/announcing-typescript-eslint-v6

Not that interesting — honestly thought there would be more to it. I put us on `@typescript-eslint/strict` and `@typescript-eslint/stylistic` and only had to turn off two rules. All the comma noise is due to Prettier 3.0.


.prettierrc.js
```diff
@@ -6,33 +6,13 @@
...
-  parse: twPlugin.parsers.typescript.parse,
-}
-
-/** @type {import("prettier").Plugin}  */
-const combinedPlugin = {
-  parsers: {
-    typescript: combinedParser,
-  },
-}
-
```
david-crespo, line 30: Not an issue anymore, that's nice

.babelrc
```diff
@@ -1,14 +0,0 @@
...
-      {
-        "corejs": "3.11",
-        "useBuiltIns": "usage"
-      }
-    ],
-    "@babel/preset-react",
-    "@babel/preset-typescript"
-  ],
-  "plugins": ["@babel/plugin-transform-runtime", "@babel/plugin-proposal-class-properties"]
-}
```
david-crespo, line 14: 99% sure this was not being used for anything?

libs/api-mocks/msw/util.ts
```diff
@@ -69,11 +69,6 @@ export const paginated = <P extends PaginateOptions, I extends { id: string }>(
 export const repeat = <T extends { id: string; name: string }>(obj: T, n: number): T[] =>
   new Array(n).fill(0).map((_, i) => ({ ...obj, id: obj.id + i, name: obj.name + i }))
 
-export const clone = <T extends object>(obj: T): T =>
-  typeof structuredClone !== 'undefined'
-    ? structuredClone(obj)
-    : JSON.parse(JSON.stringify(obj))
-
```
david-crespo, line 76: This was kind of interesting — my switch to `Map` for `db.diskBulkImportState` would not work with this `JSON.parse(JSON.stringify(obj))` fallback. Fortunately I don't think there is any situation where this code would hit the fallback. Two-year-old browsers only, basically.

https://caniuse.com/mdn-api_structuredclone

PR #1809, opened 2023-11-01
Upgrade to MSW 2.0

zephraph:
The v1 to v2 migration guide can be found [here](https://mswjs.io/docs/migrations/1.x-to-2.x/).

This PR depends on https://github.com/oxidecomputer/oxide.ts/pull/215

Most of the changes here are fairly mechanical like renaming the `rest` export from `msw` to `http`. The biggest change is the removal of the compositional `req` API from MSW which was a pseudo request object with extra metadata associate with it and the ability to compose utility functions on top of it. It now uses the standard `Request` object which includes some immediate tradeoffs like not having `cookies` hanging off of it. Cookies are now passed to the handler and have to be piped down through the rest of the call chain. Given the removal of the compositional API, utilities like `delay` are now stand alone functions which necessitates actually being able to write async handlers. I've wrapped all the handlers in `type-fest`'s `Promisable` (`T | Promise<T>`) and we were actually already awaiting the results so no changes were needed there. 

Overall, not too bad. Just remember to use `HttpResponse` for normal responses. I've also included a `json` helper export from the generated API to cut down on the diff as suggested.

### DIFF
--- a/libs/api-mocks/msw/handlers.ts
+++ b/libs/api-mocks/msw/handlers.ts
@@ -18,7 +18,7 @@ import {
   type ApiTypes as Api,
   type SamlIdentityProvider,
 } from '@oxide/api'
-import { HttpResponse, makeHandlers, type Json } from '@oxide/gen/msw-handlers'
+import { json, makeHandlers, type Json } from '@oxide/gen/msw-handlers'
 import { GiB, pick, sortBy } from '@oxide/util'
 
 import { genCumulativeI64Data } from '../metrics'
@@ -64,7 +64,7 @@ export const handlers = makeHandlers({
     }
     db.projects.push(newProject)
 
-    return HttpResponse.json(newProject, { status: 201 })
+    return json(newProject, { status: 201 })
   },
   projectView: ({ path }) => {
     if (path.project.endsWith('error-503')) {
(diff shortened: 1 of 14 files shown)


libs/api-mocks/msw/handlers.ts
```diff
@@ -63,7 +64,7 @@ export const handlers = makeHandlers({
     }
     db.projects.push(newProject)
 
-    return json(newProject, { status: 201 })
+    return HttpResponse.json(newProject, { status: 201 })
```
david-crespo, line 67: I wouldn’t say no to cutting this diff down by doing `const json = HttpResponse.json`

zephraph: Took a big of finagling but I got it to work. Diff is definitely smaller now.

david-crespo (approved):

zephraph:
For historical context, @david-crespo and I spent all of yesterday trying to debug an issue where E2E tests were failing in safari due to this browser error

```
FetchEvent.respondWith received an error: TypeError: Response cannot have a body with the given status.
```

We traced it down to a bug in a safari work around in MSW which was causing an empty array buffer to be passed to the `Response` constructor on a 204 response code which expects no body. See the full report here: https://github.com/mswjs/msw/issues/1827

I've worked around the issue in the service worker here: https://github.com/oxidecomputer/console/pull/1809/commits/7a930f20f0861f3f9abffc633010d1512b211b23

Also, note to our future selves: console logs and debug output for service workers on safari are shuttled off to a different devtools instance hidden in the developer menu.

david-crespo (approved):
Awesome. The Firefox E2E failure is likely due to GH's current outages.

PR #1811, opened 2023-11-14
Fix issue where empty IP on a NIC is submitted as an empty string

### LINKED ISSUE #1438: Add NIC to instance: validation problems with IP address field
Two issues here:
1. When adding a network interface to an existing instance, user can simply specify the VPC and subnet without a specific IP address. However, the address k/v pair is still passed to the API call with an empty value, resulting in an unhandled 400 error (i.e. the operation failed silently and the popup remains open).
2. If an IP address is specified but is invalid (e.g. outside of the subnet range), the Nexus validation error is also unhandled.

zephraph:
Fixes #1438 

When creating a NIC without an IP in the current version of console, the request fails due to the IP being malformed (an empty string). I updated the IP's default to be `undefined`, meaning it'll be left out of the submitted object and not received by the API.

#1438 says that the IP address _isn't_ optional, but the API specifies that it is. If it's not provided it should be chosen by Nexus. If that's not happening, we should fix that upstream. 

---

This also updates the generated validators as per https://github.com/oxidecomputer/oxide.ts/pull/223 to help catch this error. With the validator update but without the fix included here, the tests would fail.


app/forms/network-interface-create.tsx
```diff
@@ -23,7 +23,7 @@ import { useForm, useProjectSelector } from 'app/hooks'
 const defaultValues: InstanceNetworkInterfaceCreate = {
   name: '',
   description: '',
-  ip: '',
+  ip: undefined,
```
david-crespo, line 26: I believe this will only work with the initial value, but not if you type something in and then delete it.

zephraph: Yep, you're right.

zephraph:
This fix is incomplete because as @david-crespo pointed out the submission error will reoccur if the field is populated and emptied. #1812 will include a full fix, but we're going to merge this for now to unblock the common case.

PR #1812, opened 2023-11-16
Use `NumberField` instead of `TextField` for inputs of type `number`

zephraph:
## Changes

Replaces all instances of `TextField` where `type=number` with `NumberField`. 

## Background

When investigating the fix for #1438, I began investigating our ability to do transformations on text input before form submissions. `react-hook-form`'s docs have a section on [transformation and parsing](https://react-hook-form.com/advanced-usage#Howtosharerefusage) in which they recommend either creating a high order controller component that receives a `transform` function that can operate via `value`/`onChange` or by using [`register` props](https://react-hook-form.com/docs/useform/register#options) like `valueAsDate`, `valueAsNumber` or`setValueAs`. The latter seems closer to what we'd actually want to do, but we're not actually using `register`. `register` hangs off of the form object returned via `useForm` and refactoring our inputs to use that approach would be a larger effort than seems warranted for just this functionality. The `Controller` component which we're using to define our components has a `rules` prop that takes arguments intended for `register` but it [explicitly omits the value transformation props](https://github.com/react-hook-form/react-hook-form/blob/c7b7eacb00cd11fe8fd075c958589b5b00b2c49d/src/types/controller.ts#L39-L42) . I timed boxed some explorations in seeing if I could have the `rules` prop resolve the value transformation fields upstream but wasn't able to get to a satisfactory outcome in the time allotted so I decided just to try my hand at a user land implementation. 

After that investigation I noticed we _were_ actually doing some userland transformation in the `TextField` component. It's specifically around the special handling of numbers though, which seemed odd to me given that we have a dedicated `NumberField` component. 

https://github.com/oxidecomputer/console/blob/ae8218df707360a902133f4a96b48a3b5a62a09e/app/components/form/fields/TextField.tsx#L146-L163

The presence of this normalization logic for numbers makes the implementation of a transformation slightly more complicated than it needs to be and (to me) points to a larger need for changes. Thus I decided to take the first step of untangling the number logic from the `TextField` by specifically relegating that to the `NumberField`

## Remaining work

- [x] Ensure `NumberField` has the same input handling logic as was present in `TextField`
- [x] Remove the number transformation logic from `TextField`
- [x] Do a UI pass to ensure the changes in field components didn't cause a visual regression


app/components/form/fields/TextField.tsx
```diff
@@ -28,12 +28,13 @@ import { capitalize } from '@oxide/util'
 import { ErrorMessage } from './ErrorMessage'
 
 export interface TextFieldProps<
+  Type,
```
zephraph, line 31: I don't actually like this change, it's a bit of a contagion. It's used to aid in typing the transform but I suspect there might be a better way to do this.

zephraph:
Supplanted by #1854 and #1926

PR #1820, opened 2023-11-26
Make bump-omicron script cooler with dax, which is zx for deno

david-crespo:
Was curious about [dax](https://github.com/dsherret/dax), wanted to try it out. It is cute.


zephraph (approved):
I dig it. It's nice to have a semi-consistent way to make these scripts.

PR #1821, opened 2023-11-27
Preview meta

benjaminleonard:
Adds an OG image that's included only on Vercel deployments: 
image

One downside maybe someone can chip in on is you can't import images within the `vite.config` file. That means that the OG image is included in the public directory and therefore even if the meta data isn't included, the image itself it. Not a big deal (only 90kb) – but certainly neater if we can remove it as part of the build process.


david-crespo:
I can rm the image as part of the build step in CI, but it might not be worth doing. I’ll think about it.

david-crespo:
I was thinking what if we conditionally included the OG tag inside index.html and somehow referenced the image file in a way that it would be shaken out when the OG tag is not included. However:

> For other use cases discussed in this issue, like conditionals we would suggest adopting a full-fledge HTML pre-processor or creating inline plugins for your projects.

https://github.com/vitejs/vite/issues/3105#issuecomment-1441947641

Like I said I could rm it from the build tarball in the GH action that does it, but it's not a big deal. I'd rather just leave it.

Though it occurs to me... what if we left in the OG tag in production with an appropriate description?

Issue #1728, opened 2023-08-10
Fix time positioning in the date-picker control


paryhin:
![image](https://github.com/oxidecomputer/console/assets/56555905/6659d64b-06b2-4f35-99c1-63b9b82ae0c2)

It looks like the time is being pushed to the second line in the date-picker control.

PR #1822, opened 2023-11-29
Update width of time picker to account for wider datetimes

charliepark:
This PR just updates the width of the datetime picker container so it can handle wider values. Although [the original bug report ](https://github.com/oxidecomputer/console/issues/1728) shows the split line when the local datetime uses a leading zero, we'd see this same bug for certain datetimes with "wider" two-digit numbers in locales, as in the screenshots here.

Broken / split:
Screenshot 2023-11-28 at 4 12 59 PM

Fixed / single-line:
Screenshot 2023-11-28 at 4 35 49 PM

### DIFF
--- a/libs/ui/lib/date-picker/DateRangePicker.tsx
+++ b/libs/ui/lib/date-picker/DateRangePicker.tsx
@@ -68,7 +68,7 @@ export function DateRangePicker(props: DateRangePickerProps) {
               : 'border-default ring-accent-secondary'
           )}
         >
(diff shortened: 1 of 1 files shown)


charliepark:
The field looks a little wide to me when using a shorter datetime, but that's maybe fine; open to thoughts.
Screenshot 2023-11-28 at 4 41 20 PM

(Removing the width altogether has the side effect of changing the position of the larger calendar datepicker.)

PR #1828, opened 2023-12-02
Regen API client to bring in fix for zod error response

david-crespo:
Blocked on #1811 because this brings in the same IP address validations being fixed there.


david-crespo:
I'll just do this in #1811, there's no point doing it separately.

Issue #1824, opened 2023-11-30
Show image size on image picker


david-crespo:
I'm always wondering about the image size when I look at these. It would be very easy to include. Related to #1777, could fix together.

<img width="567" alt="image" src="https://github.com/oxidecomputer/console/assets/3612203/09a89aae-1a89-4e49-90ed-84b33c7e5d59">

charliepark:
Is this about what we want for the disk size piece of it?
![Screenshot 2023-12-05 at 5 10 24 PM](https://github.com/oxidecomputer/console/assets/22547/fb826437-dbc5-4c50-98eb-301cd2200af3)

david-crespo:
That would do it for me, though seeing it makes me wonder how we should round. I think the mock images are exact GiB multiples, but obviously that’s not true in general. I’m thinking at least one decimal point, maybe two?

zephraph:
One decimal point should definitely be sufficient.

PR #1829, opened 2023-12-02
Fix auto-updating of boot disk size when selected image is larger than current disk size; add tests

charliepark:
In the instance creation form, when a user selects an existing image, we want the disk size for the new instance to automatically increase so that it's large enough to successfully use that image. This PR updates the code so that it successfully gets increased to the next-largest increment of 10 GiB (relative to the selected image). It also adds a validator message and an e2e test to cover this behavior.

A few small UI updates in this PR as well that are worth mentioning:
1. This PR switches the order of the image lists, so that the silo images tab comes before the project images tab.
2. This PR adds a warning for the unlikely edge case that there are neither silo images nor project images. (It *is* possible to create an instance without a disk image through the API, but creating an instance via the console requires an image to be present. Accordingly, with this PR, the form is disabled when no images are available.

Screenshot 2023-12-01 at 2 57 32 PM


app/components/form/fields/DiskSizeField.tsx
```diff
@@ -32,6 +39,12 @@ export function DiskSizeField<
       min={minSize}
       max={MAX_DISK_SIZE_GiB}
       validate={(diskSizeGiB) => {
+        if (Number.isNaN(diskSizeGiB)) {
+          return 'Disk size is required'
+        }
+        if (imageSize && diskSizeGiB < imageSize) {
+          return `Must be as large as selected image (min. ${imageSize} GiB)`
+        }
```
zephraph, line 47: Another thing we could do here is just update the `min` input threshold. We'd probably want to include an info icon with some details to indicate _why_ the user can't go below the threshold though.

david-crespo: Yeah, that’s why we went with two thresholds, the message saying it’s due to the image is pretty helpful. See also #1824

app/forms/instance-create.tsx
```diff
@@ -328,17 +343,7 @@ export function CreateInstanceForm() {
...
-          if (!image) return true
-          if (diskSizeGiB < image.size / GiB) {
-            const minSize = Math.ceil(image.size / GiB)
-            return `Must be as large as selected image (min. ${minSize} GiB)`
-          }
-          if (diskSizeGiB > MAX_DISK_SIZE_GiB) {
-            return `Can be at most ${MAX_DISK_SIZE_GiB} GiB`
-          }
-        }}
+        imageSize={image?.size ? Math.ceil(image.size / GiB) : undefined}
```
zephraph, line 346: I'm not quite sure that we should should push the validation logic down into the `DiskSizeField` component. Disks, as a concept, aren't strictly speaking tied to images. Certainly in this case the disk size is semi-dependent on the image size, but that doesn't impact other situations in which a disk size may be specified. 

We'll either continue pushing more context down into the `DiskSizeField` component or have to find some other way of composing the logic. I think my preference, for now, is to localize the validation logic to where it's used (similar to what we had here). If we need that logic to be more portable perhaps we could just have some reusable helper we could spread into the usage?

david-crespo: Yeah, even though the logic is currently the same between the two spots, we could model it as a shared `validate` function that’s passed in in both spots.

charliepark: Refactored this after chatting with @david-crespo … several of the validations are standard and will be needed for the DiskSizeField regardless of where we use it. But we also want to have the option of adding one-off validators. So we added an optional `validate` prop that we can pass in when calling the component; any validations passed in via that prop will get called after the standard validator functions are evaluated.

app/components/form/fields/DiskSizeField.tsx
```diff
@@ -17,12 +22,19 @@ interface DiskSizeProps<
   TName extends FieldPath<TFieldValues>,
 > extends TextFieldProps<TFieldValues, TName> {
   minSize?: number
+  validate?(diskSizeGiB: number): ValidateResult
```
david-crespo, line 25: This is more of a FYI about React Hook form, not a suggestion, because I can't bring myself to suggest something so much worse than what you have, and I don't think it's really more typesafe in practice. But if you hover on the validate prop, you can see it is using some helpers from RHF to look up the value of the field using the field name.

Screenshot 2023-12-06 at 5 00 41 PM

If you wanted to match that type on the prop, you could do this:

```ts
  validate?: Validate<PathValue<TFieldValues, TName>, TFieldValues>
```

This produces the same result as what you have (the argument is a number, of course) except it also has a second argument, which is all the form values together:

image

This could be handy in theory because you could validate one field value based on the value of another. But we're not doing that — well, we're doing it, but we're doing it at the call site so this field doesn't have to know about it. So we don't need it, and the resulting code is dramatically harder to understand, plus the second arg is required (lol) so it actually gets mad at you for leaving it out. 

image

All this is to say, I've already typed this so I might as well shoot it into the ether, but it is not a suggestion to use this horrible type.

charliepark: Ah, that's good to know, about, though! Thanks for writing that all out.

app/forms/disk-create.tsx
```diff
@@ -79,6 +80,19 @@ export function CreateDiskSideModalForm({
...
+  const { project } = useProjectSelector()
+  const projectImages = useApiQuery('imageList', { query: { project } })
+  const siloImages = useApiQuery('imageList', {})
+
+  // put project images first because if there are any, there probably aren't
+  // very many and they're probably relevant
+  const images = [...(projectImages.data?.items || []), ...(siloImages.data?.items || [])]
+  const areImagesLoading = projectImages.isPending || siloImages.isPending
+
+  const selectedImageId = control?._formValues?.diskSource?.imageId
```
david-crespo, line 93:
```suggestion
  const selectedImageId = form.watch('diskSource.imageId')
```

This ensures the component re-renders when the image changes. If you `console.log(selectedImage)` without the `watch`, you can see that it does not render when it should.

https://react-hook-form.com/docs/useform/watch

app/forms/disk-create.tsx
```diff
@@ -79,6 +80,19 @@ export function CreateDiskSideModalForm({
...
+  const selectedImageSize = images.find((image) => image.id === selectedImageId)?.size
+  const imageSize = selectedImageSize ? bytesToGiB(selectedImageSize) : undefined
```
david-crespo, line 95: I like putting GiB in the name when it's GiB.

app/forms/disk-create.tsx
```diff
@@ -79,6 +80,19 @@ export function CreateDiskSideModalForm({
...
 
   const form = useForm({ defaultValues })
+  const control = form.control
+  const { project } = useProjectSelector()
+  const projectImages = useApiQuery('imageList', { query: { project } })
+  const siloImages = useApiQuery('imageList', {})
+
+  // put project images first because if there are any, there probably aren't
+  // very many and they're probably relevant
+  const images = [...(projectImages.data?.items || []), ...(siloImages.data?.items || [])]
```
david-crespo, line 90: I know this is how it was before, but let's put this in a `useMemo(..., [projectImages.data, siloImages.data])` — constructing fresh arrays on every render will trigger extra renders further down the chain every time.

app/forms/disk-create.tsx
```diff
@@ -144,31 +178,36 @@ const DiskSourceField = ({ control }: { control: Control<DiskCreate> }) => {
...
           />
         )}
-        {value.type === 'image' && <ImageSelectField control={control} />}
+        {value.type === 'image' && (
+          <ImageSelectField
+            control={control}
+            images={images}
+            areImagesLoading={areImagesLoading}
+          />
+        )}
```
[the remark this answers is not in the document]
david-crespo: I think you could do this in a followup since it's gnarly!

app/forms/disk-create.tsx
charliepark: `form.watch` is so much more graceful for snagging that data

app/forms/disk-create.tsx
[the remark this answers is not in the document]
charliepark: Yeah, I think adding in a second PR will make it easier to focus in on the specific issue. Thank you for the proof of concept. Will iterate off of that in another branch.

david-crespo (approved):
I think it's good!

PR #1833, opened 2023-12-04
Update generated api files with latest omicron version

charliepark:
Updated `OMICRON_VERSION` and then ran `npm run gen-api`. Also updated oxide.ts, in https://github.com/oxidecomputer/oxide.ts/pull/226.

We'll do something similar once we have the "quotas" (name TBD) API updates, which should be coming shortly.

### DIFF
--- a/libs/api/__generated__/Api.ts
+++ b/libs/api/__generated__/Api.ts
@@ -268,7 +268,7 @@ export type BgpImportedRouteIpv4 = {
 /**
  * A BGP peer configuration for an interface. Includes the set of announcements that will be advertised to the peer identified by `addr`. The `bgp_config` parameter is a reference to global BGP parameters. The `interface_name` indicates what interface the peer should be contacted on.
  */
-export type BgpPeerConfig = {
+export type BgpPeer = {
   /** The address of the host to peer with. */
   addr: string
   /** The set of announcements advertised by the peer. */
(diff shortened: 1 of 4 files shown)


charliepark:
Something seems to have not quite landed quite right with the CI test

david-crespo:
Unimplemented handlers need to be added to the object in the handlers file. I guess in theory we could just make it accepted undefined, but I actually like having the list of unimplemented ones alongside the implemented ones.

david-crespo:
Oh also there is a file using a VPC router endpoint that is no longer in the spec. I think I’d just delete the file. It is not very interesting and I’m pretty sure it’s not in use at all.

david-crespo (approved):
congrats on going net negative

zephraph (approved):
Nice! Some good cleanup work here, appreciate it.

charliepark:
Seeing some proxy errors when running console locally, so will keep poking a few things
Screenshot 2023-12-04 at 3 50 32 PM

david-crespo:
I'm not sure I've seen that before. If the request gets to Vite at all, I think that means the request was passed through/not handled by MSW. We do not render the app until after MSW is spun up, so in theory there should not be a time when API calls are being made but the server is not up. But who knows!

charliepark:
Looked into the above errors some more and chatted with @david-crespo. We don't think the issue was affecting console, and is more of an issue with the local server. Probably good to have a record here of the current state of things, as we'll be making some changes to the system utilization page soon, but don't believe the proxy error in the screenshot above will be an issue.

Issue #1830, opened 2023-12-04
Active image select item should use accent colour


benjaminleonard:
Little active state glitch coming through there too – family name and version number should use the accent colour.

_Originally posted by @benjaminleonard in https://github.com/oxidecomputer/console/issues/1824#issuecomment-1838162568_

![image](https://github.com/oxidecomputer/console/assets/4020798/aced36e7-4727-4cf3-abb5-1fee8638f6d4)

PR #1836, opened 2023-12-06
Show image size on image picker; clean up empty metadata; small color fix

charliepark:
This PR does a few things:
1. Per https://github.com/oxidecomputer/console/issues/1824, it adds the size of the image to the option in the image picker. It rounds the image size to one decimal place. (We already had a `bytesToGiB` function that defaulted to 2 decimal points; I updated that function to receive a `digits` prop (like the `round` function's prop) and set it, in this case, to 1 decimal point.) Note: Late in testing I noticed … it appears that the API is sending back an already-rounded-to-the-next-largest-GiB value, so we should look into whether that behavior is expected/wanted, though we can do that in a separate PR. 
2. Per https://github.com/oxidecomputer/console/issues/1777, it cleans up the label when data is missing from the image metadata (e.g. os, version). There might be some changes to the image upload form to change how that metadata is handled, but this at least makes it less awkward if data is missing. As there'll be the image's size now, there'll probably always be _some_ data, but we won't have the awkward comma delineation we currently have when some data is missing.
3.  Per https://github.com/oxidecomputer/console/issues/1830, it fixes the color of the metadata in the selected listbox item.

When decimals are showing below, it's because I forced a larger size. But these screenshots show what would happen:
Screenshot 2023-12-05 at 7 27 38 PM
Screenshot 2023-12-05 at 8 59 54 PM

---

Up in item 1, I noted that the API is reporting image sizes as the next GiB up. Here you can see the API response calling the images, with this particular file reporting a size of 1073741824:
Screenshot 2023-12-05 at 9 16 32 PM

### DIFF
--- a/app/components/form/fields/ImageSelectField.tsx
+++ b/app/components/form/fields/ImageSelectField.tsx
@@ -9,7 +9,7 @@ import { useController, type Control } from 'react-hook-form'
 
 import type { Image } from '@oxide/api'
 import type { ListboxItem } from '@oxide/ui'
-import { GiB } from '@oxide/util'
+import { bytesToGiB, GiB } from '@oxide/util'
 
 import type { InstanceCreateInput } from 'app/forms/instance-create'
 
@@ -44,20 +44,48 @@ export function ImageSelectField({ images, control }: ImageSelectFieldProps) {
 const Slash = () => <span className="mx-0.5 text-quinary">/</span>
 
 export function toListboxItem(i: Image, includeProjectSiloIndicator = false): ListboxItem {
-  const projectSiloIndicator = includeProjectSiloIndicator ? (
+  const { name, os, projectId, size, version } = i
+  const formattedSize = `${bytesToGiB(size, 1)} GiB`
+
+  // filter out any undefined metadata and create a comma-separated list
+  // for the selected listbox item (shown in labelString)
+  const condensedImageMetadata = [os, version, formattedSize].filter((i) => !!i).join(', ')
+  const metadataForLabelString = condensedImageMetadata.length
+    ? ` (${condensedImageMetadata})`
+    : ''
+
+  // for metadata showing in the dropdown's options, include the project / silo indicator if requested
+  const versionInfo = includeProjectSiloIndicator ? (
     <>
-      <Slash /> {i.projectId ? 'Project image' : 'Silo image'}
+      {version} <Slash /> {projectId ? 'Project image' : 'Silo image'}
     </>
-  ) : null
+  ) : (
+    version
+  )
+  // filter out undefined metadata here, as well, and create a `<Slash />`-separated list
+  // for the listbox item (shown for each item in the dropdown)
+  const metadataForLabel = [os, versionInfo, formattedSize]
+    .filter((i) => !!i)
+    .map((i, index) => (
+      <span key={`${i}`}>
+        {index > 0 ? (
+          <>
+            {' '}
+            <Slash />{' '}
+          </>
+        ) : (
+          ''
+        )}
+        {i}
+      </span>
+    ))
   return {
     value: i.id,
-    labelString: `${i.name} (${i.os}, ${i.version})`,
+    labelString: `${name}${metadataForLabelString}`,
     label: (
       <>
-        <div>{i.name}</div>
-        <div className="text-secondary">
-          {i.os} <Slash /> {i.version} {projectSiloIndicator}
-        </div>
+        <div>{name}</div>
+        <div>{metadataForLabel}</div>
       </>
     ),
   }
--- a/libs/util/math.spec.ts
+++ b/libs/util/math.spec.ts
@@ -7,7 +7,20 @@
  */
 import { expect, it } from 'vitest'
 
-import { splitDecimal } from './math'
+import { GiB } from '.'
+import { round, splitDecimal } from './math'
+
+it('rounds properly', () => {
+  expect(round(123.456, 0)).toEqual(123)
+  expect(round(123.456, 1)).toEqual(123.5)
+  expect(round(123.456, 2)).toEqual(123.46)
+  expect(round(123.456, 3)).toEqual(123.456)
+  expect(round(123.456, 4)).toEqual(123.456) // trailing zeros are culled
+  expect(round(1.9, 0)).toEqual(2)
+  expect(round(1.9, 1)).toEqual(1.9)
+  expect(round(5 / 2, 2)).toEqual(2.5) // math expressions are resolved
+  expect(round(1879048192 / GiB, 2)).toEqual(1.75) // constants can be evaluated
+})
 
 it.each([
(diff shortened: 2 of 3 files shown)


libs/util/math.spec.ts
```diff
@@ -7,7 +7,20 @@
...
+  expect(round(123.456, 0)).toEqual(123)
+  expect(round(123.456, 1)).toEqual(123.5)
+  expect(round(123.456, 2)).toEqual(123.46)
+  expect(round(123.456, 3)).toEqual(123.456)
+  expect(round(123.456, 4)).toEqual(123.456) // trailing zeros are culled
+  expect(round(1.9, 0)).toEqual(2)
+  expect(round(1.9, 1)).toEqual(1.9)
+  expect(round(5 / 2, 2)).toEqual(2.5) // math expressions are resolved
+  expect(round(1879048192 / GiB, 2)).toEqual(1.75) // constants can be evaluated
+})
```
david-crespo, line 23: we stan a testing king

app/components/form/fields/ImageSelectField.tsx
```diff
@@ -44,20 +44,48 @@ export function ImageSelectField({ images, control }: ImageSelectFieldProps) {
...
+          <>
+            {' '}
+            <Slash />{' '}
+          </>
+        ) : (
+          ''
+        )}
+        {i}
+      </span>
+    ))
```
david-crespo, line 81: You can avoid this evil JSX formatting situation by increasing the margin around the slash and getting rid of the spaces around the other usage:

```diff
diff --git a/app/components/form/fields/ImageSelectField.tsx b/app/components/form/fields/ImageSelectField.tsx
index 2ba69fdf..256fd390 100644
--- a/app/components/form/fields/ImageSelectField.tsx
+++ b/app/components/form/fields/ImageSelectField.tsx
@@ -41,7 +41,7 @@ export function ImageSelectField({ images, control }: ImageSelectFieldProps) {
   )
 }
 
-const Slash = () => <span className="mx-0.5 text-quinary">/</span>
+const Slash = () => <span className="mx-1 text-quinary">/</span>
 
 export function toListboxItem(i: Image, includeProjectSiloIndicator = false): ListboxItem {
   const { name, os, projectId, size, version } = i
@@ -57,7 +57,9 @@ export function toListboxItem(i: Image, includeProjectSiloIndicator = false): Li
   // for metadata showing in the dropdown's options, include the project / silo indicator if requested
   const versionInfo = includeProjectSiloIndicator ? (
     <>
-      {version} <Slash /> {projectId ? 'Project image' : 'Silo image'}
+      {version}
+      <Slash />
+      {projectId ? 'Project image' : 'Silo image'}
     </>
   ) : (
     version
@@ -68,14 +70,7 @@ export function toListboxItem(i: Image, includeProjectSiloIndicator = false): Li
     .filter((i) => !!i)
     .map((i, index) => (
       <span key={`${i}`}>
-        {index > 0 ? (
-          <>
-            {' '}
-            <Slash />{' '}
-          </>
-        ) : (
-          ''
-        )}
+        {index > 0 ? <Slash /> : ''}
         {i}
       </span>
     ))
```

david-crespo: Actually... I think `Project image`/`Silo image` label is supposed to go on the end. That makes it even better:

```diff
diff --git a/app/components/form/fields/ImageSelectField.tsx b/app/components/form/fields/ImageSelectField.tsx
index 2ba69fdf..73cf9c11 100644
--- a/app/components/form/fields/ImageSelectField.tsx
+++ b/app/components/form/fields/ImageSelectField.tsx
@@ -41,7 +41,7 @@ export function ImageSelectField({ images, control }: ImageSelectFieldProps) {
   )
 }
 
-const Slash = () => <span className="mx-0.5 text-quinary">/</span>
+const Slash = () => <span className="mx-1 text-quinary">/</span>
 
 export function toListboxItem(i: Image, includeProjectSiloIndicator = false): ListboxItem {
   const { name, os, projectId, size, version } = i
@@ -55,27 +55,19 @@ export function toListboxItem(i: Image, includeProjectSiloIndicator = false): Li
     : ''
 
   // for metadata showing in the dropdown's options, include the project / silo indicator if requested
-  const versionInfo = includeProjectSiloIndicator ? (
-    <>
-      {version} <Slash /> {projectId ? 'Project image' : 'Silo image'}
-    </>
-  ) : (
-    version
-  )
+  const projectSiloIndicator = includeProjectSiloIndicator
+    ? projectId
+      ? 'Project image'
+      : 'Silo image'
+    : null
+
   // filter out undefined metadata here, as well, and create a `<Slash />`-separated list
   // for the listbox item (shown for each item in the dropdown)
-  const metadataForLabel = [os, versionInfo, formattedSize]
+  const metadataForLabel = [os, version, formattedSize, projectSiloIndicator]
     .filter((i) => !!i)
     .map((i, index) => (
       <span key={`${i}`}>
-        {index > 0 ? (
-          <>
-            {' '}
-            <Slash />{' '}
-          </>
-        ) : (
-          ''
-        )}
+        {index > 0 ? <Slash /> : ''}
         {i}
       </span>
     ))
```

app/components/form/fields/ImageSelectField.tsx
```diff
@@ -41,22 +41,54 @@ export function ImageSelectField({ images, control }: ImageSelectFieldProps) {
...
     value: i.id,
-    labelString: `${i.name} (${i.os}, ${i.version})`,
+    labelString: `${name}${metadataForLabelString}`,
     label: (
       <>
-        <div>{i.name}</div>
-        <div className="text-secondary">
-          {i.os} <Slash /> {i.version} {projectSiloIndicator}
+        <div>{name}</div>
+        <div className="text-tertiary selected:text-accent-secondary">
```
benjaminleonard, line 90: Small tweak to match component in Figma library

app/components/form/fields/ImageSelectField.tsx
```diff
@@ -41,22 +41,54 @@ export function ImageSelectField({ images, control }: ImageSelectFieldProps) {
   )
 }
 
-const Slash = () => <span className="mx-0.5 text-quinary">/</span>
+const Slash = () => (
+  <span className="mx-0.5 text-quinary selected:text-accent-disabled">/</span>
```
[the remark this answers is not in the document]
benjaminleonard: @charliepark Let me know if you want to run through the process of updating the colour tokens with me.

charliepark:
Just wanted to add a screenshot to log the color updates from Ben's commit:
Screenshot 2023-12-06 at 11 34 19 AM

david-crespo (approved):

david-crespo:
Random thought, not sure if it needs a fix here or in the API — if OS and version can be non-empty whitespace-only strings (and I think they can) then this will still show the weird empty spot. We should probably reject such a string API-side, so I'm inclined not to give it special handling here.

PR #1841, opened 2023-12-07
Turn off contextual alternates in the terminal

### LINKED ISSUE #1840: Typing "oxide" into an instance's serial console renders the echoed text as "0×ide"
I was experimenting with logging into a guest image with the username 'oxide' when I noticed something odd:

oxide-or-oxide

I see that the `<span>` containing the echoed console characters has the right contents, so this seems to be purely stylistic, but it threw me for a loop (the login attempt failed for unrelated reasons, but I had to triple-check that yes, I had, in fact, sent the guest the correct username!).

Is it intended that instance serial consoles render "oxide" this way? I'd venture that this behavior should be suppressed for them. (I looked around for where this substitution was happening and could find neither hide nor hair of it, so I'd be also curious about how it's implemented!)

benjaminleonard:
Fixes #1840 

Context – many years ago when we cut a new version of the typeface, we (I) thought it might be cute to include the contextual alternate so that "Oxide" or "Ox" would look more brand-like when typed in the console. It's not cute, it's stupid, and that's my cross to bear.

This fixes the issue through an update to the type styles in `@oxide/design-system`. I've also merged in the stylistic sets we're using for the ASCII diagrams on the website, but these don't interfere with regular usage.

### DIFF
--- a/app/components/Terminal.tsx
+++ b/app/components/Terminal.tsx
@@ -101,7 +101,10 @@ export const Terminal = ({ ws }: TerminalProps) => {
 
   return (
     <>
-      <div className="h-full w-[calc(100%-3rem)]" ref={terminalRef} />
+      <div
+        className="h-full w-[calc(100%-3rem)] [font-feature-settings:_'calt'_off]"
+        ref={terminalRef}
+      />
       <div className="absolute right-0 top-0 space-y-2 text-secondary">
         <ScrollButton onClick={() => term?.scrollToTop()}>
           <DirectionUpIcon />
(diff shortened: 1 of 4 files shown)


app/components/Terminal.tsx
```diff
@@ -101,7 +101,7 @@ export const Terminal = ({ ws }: TerminalProps) => {
 
   return (
     <>
-      <div className="h-full w-[calc(100%-3rem)]" ref={terminalRef} />
+      <div className="h-full w-[calc(100%-3rem)] text-mono-code" ref={terminalRef} />
```
benjaminleonard, line 104: Technically the terminal is styled inline with some values passed to `xterm.js`. By including `text-mono-code` here it's also adding a bunch of styles that are ignored/overriden by those inline styles, but the `font-feature-settings` come with it, so it's probably the easiest and best fix here.

david-crespo (approved):

PR #1844, opened 2023-12-08
Add quotas to Utilization page

charliepark:
Screenshot 2023-12-08 at 9 50 39 AM

This is an in-process branch, adding a new CapacityBar component for the Utilization page.

It is spoofing the data for now; once the API provides the relevant `quota` values, we can hook it up. This branch has a manual update to the generated API file, which, obviously, will go away once the real API update lands.

A few things that need doing …
- [x] @charliepark remove existing CapacityMetric components
- [x] @benjaminleonard make tweaks to padding, radii, etc.
- [ ] hook up API once it's ready to go (both for CapacityBar data and tabular data below)
- [ ] review math / logic on calculations

Open to any other input folks have!


charliepark:
Screenshot 2023-12-11 at 3 47 06 PM
The table has dummy data in it for now, but the ResourceMeter component is seen in this screenshot. I'm open to suggestions if `isError` and `isWarning` aren't the right variable names, or if the default thresholds should A) be modified or B) should even exist as defaults.

libs/ui/lib/resource-meter/ResourceMeter.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const ResourceMeter = ({
+  value,
+  warningThreshold = 50,
+  errorThreshold = 75,
+}: ResourceMeterProps) => {
+  const isError = errorThreshold && value > errorThreshold
+  const isWarning = !isError && warningThreshold && value > warningThreshold
+  const usagePercent = `${value}%`
+  const label = `${usagePercent} used`
+  const bg = isError ? 'bg-destructive' : isWarning ? 'bg-notice' : 'bg-success'
```
david-crespo, line 34: Because of the default value, `errorThreshold &&` is redundant unless you want to short-circuit the check if the threshold is zero. And I get the goal of using variable names as a kind of documentation, but the `!isError` bit means you've written this `if` twice: once to get the boolean indicators, and again to convert those indicators into classes.

```ts
const bg =
  value > errorThreshold
    ? 'bg-destructive' // error
    : value > warningThreshold
    ? 'bg-notice' // warning
    : 'bg-success'
```

This is maybe a little ugly, but I like that you don't have to wonder whether `isWarning` and `isError` are going to be used for anything else. We almost never use `// prettier-ignore`, but I'm tempted to do it this:

```ts
// prettier-ignore
const bg =
  value > errorThreshold ? 'bg-destructive'
  : value > warningThreshold ? 'bg-notice'
  : 'bg-success'
```

david-crespo: Wow, I swear I hadn't seen this. Prettier is doing a whole thing about this. They're considering formatting nested ternaries in the way I suggested.

https://prettier.io/blog/2023/11/13/curious-ternaries

benjaminleonard:
Lovely work! Implementation was very close, I've just made some small nitpick tweaks.

See commit description for more detail if you're interested.

libs/ui/lib/resource-meter/ResourceMeter.tsx
benjaminleonard: Is this caught by Tailwind JIT? In practice not a big issue since these styles are used so much everywhere else there's little risk of them being omitted, but just a note.

benjaminleonard: I'd also consider `noticeThreshold` and `warningThreshold`  instead – since the only real threshold for error is when you completely run out of capacity.

app/pages/system/UtilizationPage.tsx
```diff
@@ -100,9 +100,9 @@ export function SystemUtilizationPage() {
         <CapacityBar
           icon={<Ssd16Icon />}
           title="Disk"
-          unit="GiB"
+          unit="TiB"
           metricName="virtual_disk_space_provisioned"
-          valueTransform={bytesToGiB}
+          valueTransform={bytesToTiB}
```
benjaminleonard, line 105: This explains the 73,000% utilisation aha.

app/pages/system/UtilizationPage.tsx
```diff
@@ -215,32 +243,103 @@ function UsageTab({ silos }: { silos: SiloResultsPage }) {
...
+}) => {
+  const usagePercent = (used / total) * 100
+  return (
+    <div className="flex w-full items-center justify-between">
+      <div>
+        {used} {unit && <span className="text-tertiary">{unit}</span>}
+      </div>
+      {/* We only show the ResourceMeter if the percent crosses the warning threshold (66%) */}
+      {usagePercent > 66 && (
+        <div className="absolute right-3">
```
benjaminleonard, line 339: Unsure about the least 'dangerous' way to do this.
We have extra padding on the trailing side of the cell but the resource meter needs to be aligned to this side and ignore that padding.

This is the simple solution, downside being that there's a risk the meter overlaps with the content of the cell.

libs/ui/lib/resource-meter/ResourceMeter.tsx
david-crespo: Tailwind looks for strings anywhere in the file.

> The way Tailwind scans your source code for classes is intentionally very simple — we don’t actually parse or execute any of your code in the language it’s written in, we just use regular expressions to extract every string that could possibly be a class name.
>
> We don’t just limit our search to class="..." attributes because you could be using classes anywhere

https://tailwindcss.com/docs/content-configuration#class-detection-in-depth

app/pages/system/UtilizationPage.tsx
charliepark: I think either the `absolute right-3` approach could work, or keeping it with its current setup (parent div has `justify-between` to space it out) and adding something like `transform: translateX(20px);` to shift it to the right.

To your point about the meter overlapping with the content of the cell, wouldn't the padding on the cell's trailing side prevent that collision from happening? My in-browser experiments seem to work equally well with either of the two above approaches.

app/components/CapacityBar.tsx
```diff
@@ -0,0 +1,113 @@
...
+    amount,
+  }: {
+    name: 'Provisioned' | 'Quota' | 'Capacity'
+    amount: number
+  }) => (
+    <div className="p-3 text-mono-sm">
+      <div className="text-quaternary">{name}</div>
+      <div className="text-secondary">{amount.toLocaleString()}</div>
+    </div>
+  )
```
david-crespo, line 71: I'd move this component definition outside of `CapacityBar`

libs/api-mocks/msw/handlers.ts
```diff
@@ -630,8 +630,8 @@ export const handlers = makeHandlers({
     return {
       nextPage,
       items: silos.map((silo) => ({
-        allocated: { cpus: 0, storage: 0, memory: 0 },
-        provisioned: { cpus: 0, storage: 0, memory: 0 },
+        allocated: { cpus: 52, storage: 26, memory: 130 },
+        provisioned: { cpus: 20, storage: 18, memory: 30 },
```
david-crespo, line 634: Like @zephraph said, storage and memory are going to need to be in bytes I think. `130 * GiB`. Then you use `bytesToGiB` in the code.

app/components/UtilizationDatum.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+
+export const UtilizationDatum = ({
+  name,
+  amount,
+}: {
+  name: 'Provisioned' | 'Quota' | 'Capacity'
```
zephraph, line 13:
```suggestion
  name: 'Provisioned' | 'Allocated' | 'Capacity'
```
Probably best that we stay consistent with the API

app/pages/system/UtilizationPage.tsx
```diff
@@ -55,67 +52,70 @@ SystemUtilizationPage.loader = async () => {
...
+          provisioned={totalProvisioned.cpus}
+          quota={totalAllocated.cpus}
+        />
+        <CapacityBar
+          icon={<Ssd16Icon />}
+          title="Disk"
+          unit="TiB"
+          provisioned={bytesToTiB(totalProvisioned.storage)}
+          quota={bytesToTiB(totalAllocated.storage)}
         />
```
zephraph, line 89: It's worth swapping this around to be after memory. CPU, Memory are kind of like PB&J, they just go together.

david-crespo:
This is looking great. Two things in the design that I don't love at first glance:

### Missing units

These have the unit next to the title of the card, but in general, unless it's a table column where all the numbers are lined up and it's easy to keep track of what's what, I think it's better to have units on numbers where it's not obvious. When I see these numbers without units, I have to look around and think about what the unit is.

image

### Hanging slash

The visual balance is nice with the two lines in these cells, but I think the hanging slash is a little distracting. I'd like to play around with other visual treatments, including the simple option of putting them on the same line and only wrapping if necessary. Could also get cute and do like a diagonal fraction thing.

image

david-crespo:
image

david-crespo:
Another design concern: when I'm looking at the silo rows, I want to see something like a percentage — the absolute amounts are useful, but I want to quickly gauge how close a silo is to running out. I think the little red `ResourceMeter` is supposed to give a sense of that, but for me is too small and too low-contrast to tell me that it's about 80%.

image

image

david-crespo:
Added units on storage and memory but not CPUs.

image

david-crespo:
# Final screenshots

As of right now.

Screenshot 2023-12-15 at 11 17 16 AM
Screenshot 2023-12-15 at 11 17 26 AM
Screenshot 2023-12-15 at 11 17 39 AM

benjaminleonard:
We should check that adding the units like that doesn't cause any issues at smaller sizes (especially with the three column version).

On the `ResourceMeter` we might consider an early exploration, though the size still makes getting an accurate read difficult. Or getting the percentage in there as a second row or in parentheses.

image

PR #1845, opened 2023-12-09
Move CopyToClipboard to its own component

charliepark:
This is a small refactor, pulling the CopyToClipboard functionality out of Truncate and into its own component so we can re-use it independently, as needed.

There's no change to the current interaction, but here's a recording of it in situ:
https://github.com/oxidecomputer/console/assets/22547/ea7f7a9a-c0cc-44b5-9d67-1b1b29308dfb
[Eh; thought that would just embed the .mov. Sorry for the extra step.]


charliepark:
I want to look at some accessibility affordances we might add to the button / test it out with a screenreader. Will update tomorrow when ready for review.

libs/ui/lib/copyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,44 @@
...
+      setHasCopied(true)
+    })
+  }
+
+  return (
+    <button
+      className="text-tertiary hover:text-accent-secondary"
+      onClick={handleCopy}
+      type="button"
+      aria-label={hasCopied ? 'Copied' : ariaLabel}
```
charliepark, line 35: Tested this with VoiceOver on Mac, and it reads the appropriate text in the appropriate state, and live-updates when it changes.

zephraph (approved):

PR #1853, opened 2023-12-12
Add toast when snapshotting a disk

### LINKED ISSUE #1815: Disk snapshot action should provide some UI feedback
Snapshotting a large disk can take seconds.  I often clicked "snapshot" on the same disk more than once (unnecessarily) after thinking that I might not have clicked the right spot. Some immediate popup feedback that acknowledges my action, like the kind I get for instance stop/start below, would have been helpful:
image

charliepark:
Fixes #1815.

Adds a toast when snapshotting a disk so the user knows the process has kicked off.

Screenshot 2023-12-12 at 11 17 50 AM

### DIFF
--- a/app/pages/project/disks/DisksPage.tsx
+++ b/app/pages/project/disks/DisksPage.tsx
@@ -93,6 +93,7 @@ export function DisksPage() {
     {
       label: 'Snapshot',
       onActivate() {
+        addToast({ content: `Creating snapshot of disk '${disk.name}'` })
         createSnapshot.mutate({
           query: projectSelector,
           body: {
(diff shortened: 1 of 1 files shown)


david-crespo (approved):
Noice

david-crespo:
Actually… “Success” is a kind of a silly thing to say there. It’s just the default heading on a success-styled toast. I think you can either have nothing, or something less enthusiastic. I’d lean nothing unless it looks weird.

charliepark:
Yeah, agreed that `Success` was strange; have updated it to just include the text as the title, without secondary content.
Screenshot 2023-12-12 at 11 45 02 AM

david-crespo (approved):
Double noice

Issue #1851, opened 2023-12-11
[nics] Error returned when optional field not set


karencfv:
When adding a custom network interface I got an error when I didn't set the optional IP address field

![Screenshot from 2023-12-12 11-16-37](https://github.com/oxidecomputer/console/assets/7670244/c094a5ec-65ab-4f32-93ec-2957243e04ba)
![Screenshot from 2023-12-12 11-16-59](https://github.com/oxidecomputer/console/assets/7670244/fa079a24-16c8-43d8-864d-30610d3294b1)

david-crespo:
Dupe of #1438, which was half-fixed by #1811 — it will no longer error as long as you don't type something in the field and delete it, which probably covers 99% of use. That fix is out on dogfood but not on colo.

charliepark:
Should be fully fixed by #1854

PR #1854, opened 2023-12-12
Handle empty IP Address field in Network Interface create flow

### LINKED ISSUE #1438: Add NIC to instance: validation problems with IP address field

charliepark:
We were looking at some modifications to the NumberField in https://github.com/oxidecomputer/console/pull/1812, and were going to fix #1438 with this changes, but we can decouple that work. This PR focuses on handling the issue in #1438 where a typed-in-and-then-deleted field was getting passed through as an empty string, when we actually want for that field to be `undefined`.

We'll revisit the NumberField migration separately.


charliepark:
This screenshot was in #1851 :
image

I'm not seeing that same error when I try entering in a non-IP-address syntax. Did we lose something in the recent fixes to that form?

charliepark:
Screenshot 2023-12-12 at 3 30 09 PM

david-crespo:
> This screenshot was in #1851 : image
> 
> I'm not seeing that same error when I try entering in a non-IP-address syntax. Did we lose something in the recent fixes to that form?

Real or mock API?

charliepark:
That makes sense, then!

app/forms/network-interface-create.tsx
```diff
@@ -80,7 +80,16 @@ export default function CreateNetworkInterfaceForm({
...
-      <TextField name="ip" label="IP Address" control={form.control} />
+      <TextField
+        name="ip"
+        label="IP Address"
+        control={form.control}
+        onBlur={() => {
+          if (form.getValues('ip')?.trim() === '') {
+            form.setValue('ip', undefined)
+          }
+        }}
```
david-crespo, line 91: Say you have this field focused and you click submit. That means you have blur and submit happening at roughly the same time. Is there a race condition there? Is it possible submit could run before this gets a chance to fix the value? If so, it might be necessary to intercept `onChange` somehow instead, maybe with a `transform` prop?

See for example: https://github.com/orgs/react-hook-form/discussions/8090#discussioncomment-2430118

After writing the above, I now see that's exactly what Justin is doing in #1812, so maybe I'm just remembering that.

https://github.com/oxidecomputer/console/pull/1812/files#diff-6333dd86e85b981135bbc9c2392fbd9f9d058b9ac1ea86159dd8275e9e50b404R94

This also says it's better to hook into onChange: https://react-hook-form.com/advanced-usage#TransformandParse

charliepark:
There are a few other changes in #1812 from what I have here that I'll want to take another look at, just to make sure I'm not missing something.

app/test/e2e/network-interface-create.e2e.ts
```diff
@@ -0,0 +1,79 @@
...
+  // make sure the IP address field has a non-conforming bit of text in it
+  await page.getByLabel('IP Address').fill('x')
+
+  // try to submit it
+  const sidebar = page.getByRole('dialog', { name: 'Add network interface' })
+  await sidebar.getByRole('button', { name: 'Add network interface' }).click()
+
+  // it should error out
+  // todo: improve error message from API
+  await expect(sidebar.getByText('Unknown server error')).toBeVisible()
```
david-crespo, line 67: this is because we don't handle MSW errors properly in the UI. One thing we could do, it just occurred to me, is make them look more like real API errors. As in, where right now we just return the Zod error directly as JSON, we could stick it somewhere in an object that looks like `{ "error_code": "InvalidRequest", "message": "..." }`. The Zod error would probably look like garbage stringified, but maybe we put a short version of it in `message` and then log the full thing to the console.

david-crespo: https://github.com/oxidecomputer/oxide.ts/issues/228

app/components/form/fields/TextField.tsx
```diff
@@ -56,6 +56,8 @@ export interface TextFieldProps<
   units?: string
   validate?: Validate<FieldPathValue<TFieldValues, TName>, TFieldValues>
   control: Control<TFieldValues>
+  /** Alters the value of the input during the field's onChange event. */
+  transform?: (value: string) => FieldPathValue<TFieldValues, TName>
```
david-crespo, line 60: I changed the return type from `string | undefined` to this, which effectively says it returns whatever the type of the field in question is. The explicit `string | undefined` was fine for now, but we'd run into a type error later when we tried to use this on a field with a different type.

app/test/e2e/network-interface-create.e2e.ts
```diff
@@ -71,9 +71,8 @@ test('can create a NIC with a blank IP address', async ({ page }) => {
...
   // test that the form can be submitted and a new network interface is created
   await sidebar.getByRole('button', { name: 'Add network interface' }).click()
-  const lastRow = page.getByRole('table').locator('tr:last-child')
+  await expect(sidebar).toBeHidden()
 
-  await expect(lastRow.getByText('nic-2')).toBeVisible()
-  // it should use the default value
-  await expect(lastRow.getByText('123.45.68.8')).toBeVisible()
+  // ip address is auto-assigned
+  await expectRowVisible(page.getByRole('table'), { name: 'nic-2', ip: '123.45.68.8' })
```
david-crespo, line 77: handy row assert util

david-crespo (approved):

david-crespo (approved):
nice work

PR #1857, opened 2023-12-13
Update omicron version and generated files to allow access to silo quotas

charliepark:
We now have silo quotas in Omicron; this PR makes them available to Console.

Targets Omicron sha 7c3cd6abe9d957a51465209497064fb133854932.


zephraph:
https://github.com/oxidecomputer/console/pull/1857/files#file-app-forms-silo-create-tsx-L30 will need defaults for the quota (all 0s should be fine). 

For the handlers I think you can just add `NotImplemented` helpers for the new quota operations.

app/forms/silo-create.tsx
```diff
@@ -36,6 +36,11 @@ const defaultValues: SiloCreateFormValues = {
   tlsCertificates: [],
   siloAdminGetsFleetAdmin: false,
   siloViewerGetsFleetViewer: false,
+  quotas: {
+    cpus: 0,
+    memory: 0,
+    storage: 0,
+  },
```
zephraph, line 43: So this is a good start _but_ we want to add the form elements where the operator can actually set these values. That can be done in a separate PR if you wish though.

zephraph (approved):
This looks great, thanks!

As mentioned, we'll need to add the extra form fields to silo create to allow the operators to create quotas. That can be done in a separate PR if desired though.

charliepark:
Solid. Will get this in and handle the form next.

PR #1858, opened 2023-12-14
Add quota inputs to silo create form

charliepark:
Should have gotten in on my last PR, but I thought there'd be more to this form. There are a few more things we should do to polish it (some tooltip descriptions; smarter defaults), but this at least sets it up so users can set quotas when creating a silo.

Screenshot 2023-12-14 at 11 32 36 AM

### DIFF
--- a/app/forms/silo-create.tsx
+++ b/app/forms/silo-create.tsx
@@ -44,6 +44,10 @@ const defaultValues: SiloCreateFormValues = {
   },
 }
 
+function validateQuota(value: number) {
+  if (value < 0) return 'Must be at least 0'
+}
+
 export function CreateSiloSideModalForm() {
   const navigate = useNavigate()
   const queryClient = useApiQueryClient()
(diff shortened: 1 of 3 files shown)


david-crespo (approved):
This solves the immediate problem! I think I'd like to see the silo create E2E test put something in these fields. Do they have the default min value of 1?

Is there a design for showing this on silo detail? It should be extremely easy, there's already a whole separate silo detail page so you don't have to put it in the table.

app/forms/silo-create.tsx
```diff
@@ -97,6 +98,32 @@ export function CreateSiloSideModalForm() {
...
+        control={form.control}
+        label="Memory quota"
+        name="quotas.memory"
+        required
+        type="number"
+        units="GiB"
+      />
+      <NumberField
+        control={form.control}
+        label="Disk quota"
```
david-crespo, line 120: I think for consistency with the API, we should call this `Storage quota`.

zephraph:
Thank you so much @charliepark for tackling this!

PR #1860, opened 2023-12-15
support different language

zephraph:
This updates the CapacityBar component and friends to allow for `provisioned`, `allocated`, and `capacity`. The case for rendering all three together isn't fully working yet, but inside the silo we correctly pass capacity and at the system level we correctly pass allocated (with the quotas name preserved)


app/components/CapacityBar.tsx
```diff
@@ -6,30 +6,41 @@
...
-  const percentOfAllocatedUsed = (provisioned / allocated) * 100
+  invariant(
+    typeof allocated === 'number' || typeof capacity === 'number',
+    'CapacityBar must be provided with at least one of allocated or capacity'
+  )
+  const hasAllocated = Number.isInteger(allocated)
+  const hasCapacity = Number.isInteger(capacity)
 
-  const [wholeNumber, decimal] = splitDecimal(percentOfAllocatedUsed)
+  const percentOfAllocatedUsed = (provisioned / (allocated || 0)) * 100
```
charliepark, line 37: Dividing by 0 seems like it might do something unexpected/unwanted? (next line as well)

zephraph: Weird thing w/ JS is that dividing by zero results in `Infinity`. I've leveraged that below (with `isFinite`) to make some logic distinctions. That's probably too clever by half and at least deserves an explanation via a comment if it stays.

PR #1862, opened 2023-12-15
Fix blank flash while loading the next page

### LINKED ISSUE #1861: Page goes blank while loading next page
I could be wrong but I don't remember it being like this before. I bet this is a consequence of upgrading to React Query v5, [the switch from `keepPreviousData` to `placeholderData`](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function) not being done in QueryTable.
 
https://github.com/oxidecomputer/console/assets/3612203/4abae12b-5d8f-446f-9fa3-5951c8dd59c0

david-crespo:
Closes #1861 

We could use a loading indicator while the next page loads, though. It just sits there until it changes, no feedback on click.

### DIFF
--- a/app/pages/project/instances/InstancesPage.tsx
+++ b/app/pages/project/instances/InstancesPage.tsx
@@ -80,11 +80,7 @@ export function InstancesPage() {
     )
   )
 
-  const { Table, Column } = useQueryTable(
-    'instanceList',
-    { query: projectSelector },
-    { placeholderData: (x) => x }
-  )
+  const { Table, Column } = useQueryTable('instanceList', { query: projectSelector })
 
   if (!instances) return null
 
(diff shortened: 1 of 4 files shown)


david-crespo:
This works but I noticed it does not scroll you to the top when the next page loads — it leaves you at the same spot. The blank flash version doesn't have this problem because you automatically get scrolled to the top when the page is empty. So that would have to be fixed here.

benjaminleonard:
Adding Eugene to assist with the design of this. The top loading bar doesn't show up because this isn't a route change right? 
We could loop this into #1995 – that is the same spinner that appears when the data is refreshed is used for page changes.

P.S. Making a note that we should discuss pagination (and other filtering) included in the URL or not when we get sorting and filtering.

david-crespo:
Closing, will pick this back up at some point. Probably part of #2029.

PR #1863, opened 2023-12-15
Add silo charts to Utilization Page

charliepark:
I floated this idea while chatting with @david-crespo and @zephraph, and wanted to create a quick proof of concept.

Essentially, this creates a page of silo-level charts, either to enhance or replace the existing table of silo data. The PR also puts it as the default view, which might or might not be what we want. I think it might be a little easier to understand it at a glance than the table, and has a bit of an "ooh shiny" appeal to it as well.

Screenshot 2023-12-15 at 1 31 47 PM

You'll note that it has a red bar for the element that's at a high threshold. Similarly, if we had bars over the warning threshold, we could make them yellow, similar to how we have the "ResourceMeter" component on the table data. (In time, we'll likely want to make those thresholds configurable by operators?)

Here's how it'd look if some of the data were over the "warning" threshold:
image
(this was an earlier screenshot, so some of the tab text and other small details in this screenshot are out-of-date with the PR)

I'm holding lightly to this, so if it goes nowhere, cool cool, but wanted to share it here for discussion.

### DIFF
--- a/app/components/CapacityBar.tsx
+++ b/app/components/CapacityBar.tsx
@@ -26,10 +26,17 @@ export const CapacityBar = ({
   includeUnit?: boolean
 }) => {
   const percentOfAllocatedUsed = (provisioned / allocated) * 100
-
   const [wholeNumber, decimal] = splitDecimal(percentOfAllocatedUsed)
-
   const formattedPercentUsed = `${percentOfAllocatedUsed}%`
+  const warningThreshold = 66
+  const errorThreshold = 75
+
+  const barColor =
+    percentOfAllocatedUsed > errorThreshold
+      ? 'bg-destructive-secondary border-destructive-secondary'
+      : percentOfAllocatedUsed > warningThreshold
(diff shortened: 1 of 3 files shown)


benjaminleonard:
Always love to test ideas like this!

At first glance the colour switch is something I think we should avoid here. Since initially colour is used to label the bars – green being provisioned – switching that could be a little confusing. If we want to apply a warning we could probably drop a little warning icon next to the utilization percentage.

I think if we were to do something like this it'd be good to alter the layout to be more useful comparatively. We start to see it a little bit because each resource is vertically aligned with the ones below it, and I think making it a little bit more tabular through our composition (without actually putting it in a table), might feel a bit more considered.

But I think it could be a useful addition, we could do a little Figma session together and work some ideas through directly in the designs.

charliepark:
I'm going to close this draft for now to keep the `/pulls` list clean, but we can always pick it back up, or just reference it when brainstorming other options.

PR #1865, opened 2023-12-17
Update how our util functions round numbers

### LINKED ISSUE #1864: Utilization Summary: Available quantity of resources should be kept to two decimals
The large number of decimals seem to show up only in the available storage column but please ensure all numbers in the table are handled in the same way.

image

charliepark:
Fixes #1864

### DIFF
--- a/app/pages/system/UtilizationPage.tsx
+++ b/app/pages/system/UtilizationPage.tsx
@@ -24,7 +24,7 @@ import {
   Table,
   Tabs,
 } from '@oxide/ui'
-import { bytesToGiB, bytesToTiB } from '@oxide/util'
+import { bytesToGiB, bytesToTiB, round } from '@oxide/util'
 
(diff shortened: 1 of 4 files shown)


app/pages/system/UtilizationPage.tsx
```diff
@@ -254,7 +254,8 @@ const AvailableCell = ({
   return (
     <div className="flex w-full items-center justify-between">
       <div>
-        {allocated - provisioned} {unit && <span className="text-tertiary">{unit}</span>}
+        {round(allocated - provisioned, 2)}
+        {unit && <span className="text-tertiary"> {unit}</span>}
```
david-crespo, line 258: ah so we did forget to round! the mistake was assuming the difference of two rounded numbers would remain rounded. JAVASCRIPT

david-crespo (approved):

Issue #1818, opened 2023-11-22
Select and add SSH keys within instance create form


benjaminleonard:
Console implementation of this [3056](https://github.com/oxidecomputer/omicron/issues/3056) when it is added to Omicron.

> The current implementation automatically [retrieves](https://github.com/oxidecomputer/omicron/blob/6f8cc328740b0f535df01c91efc38108a34edad4/nexus/src/app/instance.rs#L713-L748) and [inserts](https://github.com/oxidecomputer/omicron/blob/6f8cc328740b0f535df01c91efc38108a34edad4/nexus/src/cidata.rs#L17-L36) a user's SSH keys into the instances they create. However, it would be beneficial for users to have control over which keys are added, as well as the ability to pass an SSH key not associated with their account.

> I propose adding an option to the [instance_create](https://docs.oxide.computer/api/instance_create) API that accepts an array of attached SSH keys identified by name or ID, or a custom SSH key provided as a string (which will not be linked to the user's account).

> For an enhanced user experience, we could also include an option that automatically incorporates a user's SSH keys. This would be particularly useful for the CLI, where retrieving SSH keys and adding them as an option is more cumbersome.

> It might be reasonable to defer implementation until after FCS. In the meantime, I could work on adding a console message to inform users that all their SSH keys will be injected into the instance (where cloud-init is supported).

david-crespo:
I'm worried about the checkbox version because it means this section can get arbitrarily tall. In practice maybe not an issue, but it would be good if the box could scroll.

![image](https://github.com/oxidecomputer/console/assets/3612203/6db110a5-db79-43c9-800d-40d0c867833c)

david-crespo:
Love that the checkbox design is very compact but it could get arbitrarily tall unless we put it in a box that can scroll.

![image](https://github.com/oxidecomputer/console/assets/3612203/6db110a5-db79-43c9-800d-40d0c867833c)

benjaminleonard:
Considered a few solutions but perhaps the simplest would be a "More ↓" if there's more than x number of keys.

david-crespo:
Yeah, that works. 99.9% of the time there should not be more than a few keys.

charliepark:
Will assume a "check all / uncheck all" button would be overkill for now, assuming there won't be more than a few keys in most cases, but it could be a nice affordance if we find that's not the case.

david-crespo:
Closed by #1867

PR #1867, opened 2023-12-20
Instance custom SSH key select

benjaminleonard:
Starting to block out the UI for this (hopefully with some support from @charliepark on this and `omicron`).

CleanShot 2023-12-20 at 14 52 08

The text area is within the checkbox which is worth looking into incase of accessibility ramifications. But it does make the alignment easier.

This PR adds the `CheckboxGroup` component, modelled off the `RadioGroup`. 

---

Here's the types I am proposing. Names might not be quite right. We might also accept either `InstancePublicKeys[]` or an alternative value that includes all of a user's SSH keys – or decide that it can be added to the console and/or CLI but a user can grab those SSH keys themselves (as we are doing here) if they're creating an instance with the API.

```ts
export type InstancePublicKeys = 
  { key: NameOrId; type: 'user_key'} |
  { key: string; type: 'string' }

/**
 * Create-time parameters for an `Instance`
 */
export type InstanceCreate = {
  description: string
  /** The disks to be created or attached for this instance. */
  disks?: InstanceDiskAttachment[]
  /** The external IP addresses provided to this instance.

By default, all instances have outbound connectivity, but no inbound connectivity. These external addresses can be used to provide a fixed, known IP address for making inbound connections to the instance. */
  externalIps?: ExternalIpCreate[]
  hostname: string
  memory: ByteCount
  name: Name
  ncpus: InstanceCpuCount
  publicKeys: InstancePublicKeys[]
  /** The network interfaces to be created for this instance. */
  networkInterfaces?: InstanceNetworkInterfaceAttachment
  /** Should this instance be started upon creation; true by default. */
  start?: boolean
  /** User data for instance initialization systems (such as cloud-init). Must be a Base64-encoded string, as specified in RFC 4648 § 4 (+ and / characters with padding). Maximum 32 KiB unencoded data. */
  userData?: string
}
```

---

- [ ] Add SSH key validation (likely just required)
- [ ] Accessibility checks
- [ ] Omicron API work
- [ ] Documentation updates (especially if the default behaviour is no longer injecting all of a users SSH keys)

---

Related issues:
#1818 
https://github.com/oxidecomputer/omicron/issues/3056


app/components/form/fields/SshKeysField.tsx
```diff
@@ -0,0 +1,122 @@
...
+          />
+        </div>
+      )}
+      <CheckboxGroup name="ssh-keys" column className="mt-4">
+        {[
+          ...keys.map((key) => (
+            <Checkbox
+              key={key.id}
+              value={key.id}
+              onChange={(e) => {
```
benjaminleonard, line 75: In theory we could add this to the `CheckboxGroup` if that feels less wasteful (but would need to ignore the "New SSH Key" checkbox.

libs/ui/lib/checkbox/Checkbox.tsx
```diff
@@ -20,7 +20,7 @@ const inputStyle = `
   appearance-none border border-default bg-default h-4 w-4 rounded-sm absolute left-0 outline-none
   disabled:cursor-not-allowed
   hover:border-hover hover:cursor-pointer
-  checked:bg-accent-secondary checked:border-accent-secondary checked:hover:border-accent
+  checked:bg-accent-secondary checked:border-accent-secondary checked:hover:border-accent [&:checked+svg]:block
```
benjaminleonard, line 23: Not really affected by what we're doing here but felt like a reasonable QoL change – this means that the checkbox still shows up if the `checked` prop is not being controlled.

david-crespo:
Not sold on taking up all that space to allow pasting in a one-off key. Is this aimed at preventing someone from having to leave the flow because they didn’t know to add a key beforehand? What about a button that pops up the SSH key add modal? Maybe a little confusing because unlike the disk modals, which are building to the response, that one would actually add a key.

benjaminleonard:
The text input only appears if the "New SSH Key" checkbox is checked. So 99% of the time it's not visible.

david-crespo:
Ok, that makes sense. Will think about the API

david-crespo:
Ok, that makes sense. On the API shape, I think we can avoid having two types of things in the list because there is already a `user_data` field on the instance create request body, so we could use that for the one-off key, and keep the list of keys only for existing ones with NameOrId.

david-crespo:
Sorry, I’m mashing buttons on my phone.

benjaminleonard:
> Ok, that makes sense. On the API shape, I think we can avoid having two types of things in the list because there is already a `user_data` field on the instance create request body, so we could use that for the one-off key, and keep the list of keys only for existing ones with NameOrId.

That would be simpler. Does that mean we can have multiple `public_keys` in the cloud init config?

benjaminleonard:
Might also be worthwhile adding the `user_data` free form or file input field to this PR too.

david-crespo:
Right, good point. And yes, I think we could have multiple one-off keys in there. It’s free-form.

zephraph:
Instead of having `New SSH Key` as some hybrid select/button logic how about we just allow opening the side modal form (like for regular SSH key creation) via a button and give the user the option to save or not save the SSH key. 

When you're first going through the flow (we'll have plenty of first time instance creators) it's easy to get here an not have an SSH key. This leads to a major point of friction: if you want to persist the key you either have to leave and repopulate the field (as is currently the case) or remember to persist it later. Both seem unfortunate.

benjaminleonard:
> Instead of having `New SSH Key` as some hybrid select/button logic how about we just allow opening the side modal form (like for regular SSH key creation) via a button and give the user the option to save or not save the SSH key.

I think that could work – I'll hold off on the UI side until we settle on the omicron implementation in case the shape of that API informs this.

benjaminleonard:
SSH keys can be added from within the form with a notice to say they will stick around and are scoped to the user not the instance – I experimented with one-time keys but that'd need some more complex work merging it into `user_data`.
CleanShot 2024-01-25 at 14 54 27

Can now select/deselect all with a checkbox (all keys are checked by default):
CleanShot 2024-01-25 at 14 31 46

Ideally the new keys are pre-selected I suppose, though it might not be worth the work.

benjaminleonard:
Just need to fix the `onChange` error on the `<Checkbox>` and I think it's ready to go (when the omicron pre-req is merged).

app/components/form/fields/SshKeysField.tsx
```diff
@@ -0,0 +1,138 @@
...
+    // todo: find some way to ensure that this is not out of date with the omicron limit
+    if (sshKeys && sshKeys.length > MAX_KEYS_PER_INSTANCE) {
+      control.setError('sshKeys', {
+        type: 'manual',
+        message: `An instance supports a maximum of ${MAX_KEYS_PER_INSTANCE} SSH keys`,
+      })
+    } else {
+      clearErrors()
+    }
+  }, [sshKeys, control, clearErrors])
```
david-crespo, line 56: I think there's a more direct way to do this validation. Will work on it.

app/components/form/fields/SshKeysField.tsx
```diff
@@ -0,0 +1,134 @@
...
+    control,
+    name: 'sshKeys',
+    rules: {
+      validate(keys) {
+        if (keys.length > MAX_KEYS_PER_INSTANCE) {
+          return `An instance supports a maximum of ${MAX_KEYS_PER_INSTANCE} SSH keys`
+        }
+      },
+    },
+  })
```
david-crespo, line 69: It turns out this does work. I had forgotten that we changed all form validation to happen on submit, so this will not validate the number of keys until the user clicks submit. It's not ideal as a UX, but it is consistent with everything else. After that initial validation on submit, I think it validates on change, so that is a nice experience.


https://github.com/oxidecomputer/console/assets/3612203/d6f66745-9417-4610-be87-4f23eae0e9c9

benjaminleonard: Now that this limit has increased to 100, I think it'll be a relatively uncommon occurrence.

benjaminleonard:
Updated API based based on Omicron 5780ff6d1baaa9199e2a64a6b91665d636bf2e3e.
Ready for review.

david-crespo (approved):
we got it

app/components/form/fields/SshKeysField.tsx
zephraph: We'll need a nicer UI for longer lists here eventually I think. Looks good for now though!

zephraph (approved):

PR #1868, opened 2023-12-20
Enable alternate icon on Message component

charliepark:
This came up in conversation with @benjaminleonard earlier and was quick to throw together.

We already have four kinds of Message — 'success' | 'error' | 'notice' | 'info'. Each has a default icon that shows up with it, as seen here:
Screenshot 2023-12-20 at 11 44 00 AM

In a recent mock from @paryhin, we see the possibility of a different icon, to correspond to the message inside the box:
Screenshot 2023-12-20 at 11 49 27 AM

This PR adds a new optional prop to the Message component, allowing for an `alternateIcon`.
Screenshot 2023-12-20 at 11 35 32 AM
Screenshot 2023-12-20 at 11 33 37 AM
Screenshot 2023-12-20 at 11 34 35 AM
Screenshot 2023-12-20 at 11 34 58 AM

The color of the icon is controlled by the `variant` prop, so the only thing that needs to be done here is to pass in one of the existing icon components.

One thing to note is that the Message will happily receive a 16px or 24px icon and will appropriately size it to the Message box, which is good from an ease-of-use standpoint, but might yield issues from an "if we're expecting a 12px icon, we should probably use a 12px icon" perspective. I've added a comment to that end.

Feel free to push back on any of it. If we nix the idea in the interest of consistency of message types (or any other reason), no worries.

### DIFF
--- a/libs/ui/lib/message/Message.tsx
+++ b/libs/ui/lib/message/Message.tsx
@@ -72,7 +72,7 @@ export const Message = ({
   className,
   variant = 'success',
   cta,
-  icon = defaultIcon[variant],
+  icon,
 }: MessageProps) => {
   return (
     <div
@@ -83,7 +83,7 @@ export const Message = ({
         className
       )}
     >
-      <div className="mt-[2px] flex svg:h-3 svg:w-3">{icon}</div>
+      <div className="mt-[2px] flex svg:h-3 svg:w-3">{icon || defaultIcon[variant]}</div>
       <div className="flex-1 pl-2.5">
         {title && <div className="text-sans-semi-md">{title}</div>}
         <div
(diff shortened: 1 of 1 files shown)


libs/ui/lib/message/Message.tsx
```diff
@@ -70,6 +72,7 @@ export const Message = ({
   className,
   variant = 'success',
   cta,
+  alternateIcon,
```
zephraph, line 75:
```suggestion
  icon: alternateIcon,
```

Small nit here, but from the perspective of the consumer of this component they have the option to provide an icon or a default will be used... so I think having this just be `icon` would be sufficient. Non-blocking, fine if you just want to move forward as is.

zephraph (approved):

PR #1878, opened 2024-01-10
Increase pagination size from 10 to 25

### LINKED ISSUE #1794: Want option to display more resources (instances, disks) per page
From a customer who came off a demo of the console to internal users:
Is there a way to increase the number of resources (instances, disks) etc listed as a table on various pages? Or could that be dynamic? It would be nice to take better advantage of available screen real estate (especially vertical monitors) and minimize clicks.

benjaminleonard:
Open to other suggestions for the number here (maybe 24 if this feels like too big of a jump) – don't want to go too high because then we need to worry about scroll restoration and it defeats the point of paginating to begin with.

Somewhat fixes #1794

### DIFF
--- a/libs/api-mocks/snapshot.ts
+++ b/libs/api-mocks/snapshot.ts
@@ -5,11 +5,16 @@
  *
  * Copyright Oxide Computer Company
  */
+
 import type { Snapshot } from '@oxide/api'
 
 import { disks } from './disk'
 import type { Json } from './json-type'
 import { project } from './project'
+
+const generatedSnapshots: Json<Snapshot>[] = Array.from({ length: 25 }, (_, i) =>
+  generateSnapshot(i)
+)
 
 export const snapshots: Json<Snapshot>[] = [
   {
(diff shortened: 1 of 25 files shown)


david-crespo:
32 is weird, why not a multiple of 10? This number also needs to be changed in all the prefetches. Look for `limit: 10`.

benjaminleonard:
We all have numbers we think are prettier than others...

Looking at a few other examples – another service allows the user to choose between 10, 25 and 50. Shooting for the middle for now but we might eventually allow the user to select their own.

zephraph (approved):
25 seems fine to me

libs/api-mocks/snapshot.ts
```diff
@@ -67,4 +72,28 @@ export const snapshots: Json<Snapshot>[] = [
...
+}
+
+// Can't use `crypto.randomUUID()` with playwright
+function generateUUID() {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+    const r = (Math.random() * 16) | 0,
+      v = c === 'x' ? r : (r & 0x3) | 0x8
+    return v.toString(16)
+  })
+}
```
zephraph, line 99: The `uuid` library should work instead of doing this, right?

PR #1879, opened 2024-01-11
Add a tooltip with absolute datetime to relative time element

### LINKED ISSUE #641: Human-formatted "ago" dates should have full timestamp in tooltip

charliepark:
Fixes #641 

This PR pulls the formatted "X {seconds|minutes|hours|days|etc.} ago" relative time into a new component that also wraps it in a tooltip that renders the absolute value of that timestamp.

For example:
Screenshot 2024-01-11 at 12 05 57 PM

The description is optional; a relative time without that label still renders the absolute time appropriately:
Screenshot 2024-01-11 at 12 06 44 PM

### DIFF
--- a/app/components/TimeAgo.tsx
+++ b/app/components/TimeAgo.tsx
@@ -1,0 +1,33 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+
+import { format } from 'date-fns'
+
+import { Tooltip } from '@oxide/ui'
+
+import { timeAgoAbbr } from 'app/util/date'
+
+export const TimeAgo = ({
+  datetime,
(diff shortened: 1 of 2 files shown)


david-crespo:
If `FAILED for 8d` sounds silly, you could imagine something like `FAILED since 8d ago`. But I think `FAILED for 8d` is tolerable.

charliepark:
Fixed the tooltip centering issue:
Screenshot 2024-01-11 at 3 02 57 PM

`[failed] for 8d` sounds okay to me, and will hopefully not be a common state for people. `[starting] for 2s` sounds a bit weird to me, but I don't have strong feelings on it. Happy to handle any updates in this PR once Ben shares his thoughts.

benjaminleonard:
The `[State] for 9s` feels uncomfortable to me in a different way. Perhaps because it places extra emphasis on the duration of the state when really the user is most interested in the state itself, or that it's written as a sentence in a way that the other columns are not.

I'd like to investigate some other options, and look into how this is handled on other platforms. But this doesn't need to be a blocker on this PR.

david-crespo (approved):
love it

PR #1882, opened 2024-01-15
Add VPC and instance external IP to instance page

### LINKED ISSUE #1775: Copy button on instance external IP

### LINKED ISSUE #1849: External IP address on instance list and view page
The first thing I want to do once I create an instance is connect to it. It'd be nice to have the external IP address as a first class citizen on both list and view pages. It took me a while to figure out where it was.

Screenshot from 2023-12-12 11-10-05
Screenshot from 2023-12-12 11-10-16

benjaminleonard:
Fixes #1849 

image

Some assumptions here: we're just interested in showing the primary VPC here, but we might want to show all of the external IPs.

Added VPC here also because whilst the current navigation to get to the associated VPC feels systematically correct, we have had people struggle to find their firewall rules and this is part of a larger effort to make that easier to find.

Also making it clearer they are clickable with the underline (following up with some similar changes elsewhere to make other links clearer).

Also fixes #1775 through #1885


app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -52,10 +52,11 @@ const VpcNameFromId = ({ value }: { value: string }) => {
   // possible because you can't delete a VPC that has child resources, but let's
   // be safe
   if (isError) return <Badge color="neutral">Deleted</Badge>
-  if (!vpc) return <Spinner /> // loading
+  if (!vpc)
+    return <div className="h-4 w-12 rounded bg-tertiary motion-safe:animate-pulse" /> // loading
```
benjaminleonard, line 56: Using a skeleton state here rather than the full spinner which can be a bit strong.

benjaminleonard:
Perhaps also beyond this PR, but would be good to have a mechanism to truncate arrays of strings. See:

image

david-crespo:
> Perhaps also beyond this PR, but would be good to have a mechanism to truncate arrays of strings. See:
> image

Are you picturing a truncate with "show more" button? In this spot I think it would be fine if it wrapped and stretched the box vertically if the user asked for it. Though as the thing is written I think it would mess up the grid.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
david-crespo: I didn't know we had that. That's great. You should update `SubnetNameFromId` too, which is also using the spinner. Also maybe make it wider (maybe not the pulsing thing itself, but its container) so the pop-in is less severe when the name shows up? Though that depends on making a guess about the average length of name.

app/pages/project/instances/instance/InstancePage.tsx
```diff
@@ -54,6 +62,14 @@ export function InstancePage() {
     query: { project: instanceSelector.project },
   })
 
+  const { data: nicList } = usePrefetchedApiQuery('instanceNetworkInterfaceList', {
+    query: {
+      project: instanceSelector.project,
+      instance: instanceSelector.instance,
+      limit: 10,
```
david-crespo, line 69: a stupid thing, but I think there is a max of 8 results here anyway. you can leave off the `limit` and rely on the API default of 100

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -77,14 +79,36 @@ const SubnetNameFromId = ({ value }: { value: string }) => {
...
   const { project, instance } = useInstanceSelector()
-  const { data } = useApiQuery('instanceExternalIpList', {
+  const { data, isLoading } = useApiQuery('instanceExternalIpList', {
     path: { instance },
     query: { project },
   })
-  const ips = data?.items.map((eip) => eip.ip).join(', ')
-  return <span className="text-secondary">{primary ? ips : <>&mdash;</>}</span>
+  if (isLoading)
+    return <div className="h-4 w-12 rounded bg-tertiary motion-safe:animate-pulse" /> // loading
```
david-crespo, line 89:
```ts
const Skeleton = div.classed`h-4 w-12 rounded bg-tertiary motion-safe:animate-pulse`
```

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -77,14 +79,36 @@ const SubnetNameFromId = ({ value }: { value: string }) => {
...
+      >
+        {eip.ip}
+      </a>
+      {index < array.length - 1 && <span className="text-quinary"> / </span>}
+    </Fragment>
+  ))
+
+  return (
+    <div className="flex items-center gap-1 text-secondary">
+      {primary ? ips : <>&mdash;</>}
```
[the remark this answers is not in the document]
david-crespo: fixed it I think

image

libs/ui/lib/properties-table/PropertiesTable.tsx
```diff
@@ -45,7 +45,7 @@ PropertiesTable.Row = ({ label, children }: PropertiesTableRowProps) => (
     <span className="flex items-center">
       <Badge>{label}</Badge>
     </span>
-    <div className="flex items-center overflow-hidden whitespace-nowrap pr-4 text-sans-md text-secondary">
+    <div className="flex h-11 items-center overflow-hidden whitespace-nowrap pr-4 text-sans-md text-secondary">
```
benjaminleonard, line 48: We were getting a flash of movement when the skeleton switched to the text since they're not the exact same height. This fixes that and any future issues where the content of the row is not the same height as `text-sans-md`.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -40,7 +42,13 @@ import { pb } from 'app/util/path-builder'
 
 import { fancifyStates } from './common'
 
-const VpcNameFromId = ({ value }: { value: string }) => {
+export const Skeleton = classed.div`h-4 w-12 rounded bg-tertiary motion-safe:animate-pulse`
+
+export const EmptyCellContent = () => (
```
benjaminleonard, line 47: Might be a better home for this?

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -77,14 +85,43 @@ const SubnetNameFromId = ({ value }: { value: string }) => {
   return <span className="text-secondary">{subnet.name}</span>
 }
 
-function ExternalIpsFromInstanceName({ value: primary }: { value: boolean }) {
+function IpLink({ ip }: { ip: string }) {
+  return (
+    <a
+      className="underline text-sans-semi-md text-secondary hover:text-default"
+      href={ip}
```
benjaminleonard, line 92: This needs a protocol – we can assume this is `http` right?

david-crespo: Not sure. Might be better to do https and let them fix it after nav if it’s wrong.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -209,7 +203,9 @@ export function NetworkingTab() {
           // we use primary to decide whether to show the IP in that row
           accessor="primary"
           id="external_ip"
-          cell={ExternalIpsFromInstanceName}
+          cell={({ value: primary }) =>
+            primary ? <ExternalIps {...instanceSelector} /> : <EmptyCell />
+          }
```
david-crespo, line 208: This primary logic was inside the component before, but it's only relevant here, not in the table at the top. But you know... what if we just got rid of this column? The relation to the network interface is kind of tenuous to begin with. VPC name is different because each network interface can point to a different VPC. But external IPs really are instance level. So let's take them out of this table?

david-crespo: I think it's an improvement! One other thing we could do is put a tooltip on IP to explain that it's an internal IP?

image

app/pages/project/instances/instance/InstancePage.tsx
```diff
@@ -81,6 +116,8 @@ export function InstancePage() {
 
   const memory = filesize(instance.memory, { output: 'object', base: 2 })
 
+  const primaryVpcId = nics.items.length > 0 ? nics.items[0].vpcId : undefined
+
```
david-crespo, line 120: It just occurred to me — I don't think we can rely on the first NIC being the primary one. We have to find the primary directly.

PR #1883, opened 2024-01-15
Use dynamic z-index for tooltips too

benjaminleonard:
Was testing a tooltip and noticed it wasn't appearing within a side modal – carried over the dynamic z-index from the listbox.

### DIFF
--- a/libs/ui/lib/tooltip/Tooltip.tsx
+++ b/libs/ui/lib/tooltip/Tooltip.tsx
@@ -32,6 +32,9 @@ import {
 } from 'react'
 import { mergeRefs } from 'react-merge-refs'
 
+import { useIsInModal } from '../modal/Modal'
+import { useIsInSideModal } from '../side-modal/SideModal'
+
 import './tooltip.css'
 
 export interface TooltipProps {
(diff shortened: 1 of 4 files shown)


libs/ui/lib/tooltip/Tooltip.tsx
```diff
@@ -83,14 +86,22 @@ export const Tooltip = forwardRef(
...
     })
 
+    const isInModal = useIsInModal()
+    const isInSideModal = useIsInSideModal()
+    const zIndex = isInModal
+      ? 'z-modalDropdown'
+      : isInSideModal
+        ? 'z-sideModalDropdown'
+        : 'z-contentDropdown'
+
```
david-crespo, line 96: Since the logic is the same and feels unlikely to diverge, might as well share it:

```ts
function usePopoverZIndex() {
  const isInModal = useIsInModal()
  const isInSideModal = useIsInSideModal()
  return isInModal
    ? 'z-modalDropdown'
    : isInSideModal
      ? 'z-sideModalDropdown'
      : 'z-contentDropdown'
}
```

PR #1885, opened 2024-01-15
Adds copy button on instance external IP

### LINKED ISSUE #1775: Copy button on instance external IP

benjaminleonard:
Fixes #1775 

With updated copy to clipboard component:
CleanShot 2024-01-15 at 16 11 28


david-crespo (approved):
love it

PR #1886, opened 2024-01-15
Improve clickability of link cells

benjaminleonard:
Draft because I want to get people's thoughts on it. I think it might be easier to discover and a nicer experience to make the entire link cell clickable. Presently just the text is clickable.

CleanShot 2024-01-15 at 17 06 57

### DIFF
--- a/libs/table/cells/LinkCell.tsx
+++ b/libs/table/cells/LinkCell.tsx
@@ -18,6 +18,7 @@ export const linkCell =
         to={makeHref(value)}
       >
         {value}
+        {/* Pushes out the link area to the entire cell for improved clickability™ */}
         <div className="absolute inset-0" />
       </Link>
     )
--- a/libs/table/cells/LinkCell.tsx
+++ b/libs/table/cells/LinkCell.tsx
@@ -13,8 +13,14 @@ export const linkCell =
   (makeHref: (value: string) => string) =>
   ({ value }: Cell<string>) => {
     return (
-      <Link className="text-sans-semi-md text-default hover:underline" to={makeHref(value)}>
-        {value}
-      </Link>
+      <>
+        <Link
+          className="link-cell absolute inset-0 flex h-full w-full"
+          to={makeHref(value)}
+        />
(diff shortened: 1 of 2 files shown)


benjaminleonard:
The `Link` is positioned absolutely to work around the padding on its element. Wondering if it might however cause issues with screen readers? Perhaps expecting text to be inside the link element.

benjaminleonard:
Third option seems to work the best with minimal jank. Need to do some cross-browser testing.

david-crespo:
What if it's just underlined all the time? Then you probably wouldn't need to change the target.

benjaminleonard:
Worth including as an option, I think it gets a bit visually intense but lets give it a go.

I do think that can be separate, since it's also about the comfort of the size of the clickable area.

zephraph:
You can definitely tell the improvement on mobile (not that our layout is optimized for that generally). It does feel pretty good. I agree that the third option seems the least fragile, but I think you should definitely add an inline comment saying what its purpose is.

zephraph (approved):

libs/table/cells/LinkCell.tsx
```diff
@@ -13,8 +13,14 @@ export const linkCell =
   (makeHref: (value: string) => string) =>
   ({ value }: Cell<string>) => {
     return (
-      <Link className="text-sans-semi-md text-default hover:underline" to={makeHref(value)}>
+      <Link
+        className="flex h-full w-full items-center text-sans-semi-md text-default hover:underline"
+        aria-label={value}
```
david-crespo, line 18: is this to prevent the div from messing with it? in general I would expect links to get their accessible name from the text, which is already there

david-crespo (approved):
Looks good to me. I kinda like seeing a hover background color on the cell but it would be redundant with the cursor and link underline.

libs/table/cells/LinkCell.tsx
benjaminleonard: Will fix that – was a fix no longer needed

PR #1887, opened 2024-01-15
Firewall rule improvements

### LINKED ISSUE #1571: Firewall rule issues
There are several issues with the popup for adding/editing firewall rules:

1. Unable to add multiple host filters of the same host type when editing an existing firewall rule: I tried to add 2 different IP addresses under the host filter section but it only allows the first one. There is no javascript error as far as I can tell from the browser console. (Interestingly, I am able to do so if I add a new fw rule.)

2. Target name is always reset to "VPC name" after adding a target: This issue happens to both Target and Host filters. The name field is set dynamically set based on the target/host type, e.g. VPC name, Instance name, IP name. After adding an IP address, the name field is reset to "VPC name" while the type stays as "IP". If I enter a second IP address in the name box now, it's added as VPC.
image
To workaround this, I have to toggle the "Type" dropdown to something else and then back to IP (to poke the onChange action). It's a minor annoyance.

### LINKED ISSUE #1915: Show firewall rule priority and sort by highest priority first
Also show inbound/outbound if possible.

benjaminleonard:
Closes #1571
Closes #1915

- Makes name clickable (takes you to the edit form)
- Add direction column:
image

- Make action title case
- Use number field for priority:
image

- Use mini table:
image

- Clear is disabled when associated fields are not dirty


app/pages/project/networking/VpcPage/tabs/VpcFirewallRulesTab.tsx
```diff
@@ -32,10 +33,17 @@ import { confirmDelete } from 'app/stores/confirm-delete'
 
 const colHelper = createColumnHelper<VpcFirewallRule>()
 
+const directionMap = {
+  inbound: 'Incoming',
+  outbound: 'Outgoing',
+}
```
david-crespo, line 39: We should use the API's language where possible. In this case the disparity feels unnecessary.

benjaminleonard: Ah I agree, thought this came from
you and was matching the form. In that case I should probably tweak the form too?

libs/util/str.ts
```diff
@@ -42,3 +42,10 @@ export const commaSeries = (items: string[], conj: string) => {
   }
   return [...items.slice(0, -1), `${conj} ${items.at(-1)}`].join(', ')
 }
+
+export const titleCase = (text: string): string => {
+  return text.replace(
+    /\w\S*/g,
+    (text) => text.charAt(0).toUpperCase() + text.substring(1).toLowerCase()
+  )
+}
```
david-crespo, line 51:
```
$ pbpaste | codeblock ts | ai -a 'write tests for this'


    import { describe, it, expect } from 'vitest';
    import { titleCase } from './titleCaseFunction'; // Replace with the
  correct path to titleCase function

    describe('titleCase', () => {
      it('converts single words to title case', () => {
        expect(titleCase('hello')).toBe('Hello');
      });

      it('converts multiple words to title case', () => {
        expect(titleCase('hello world')).toBe('Hello World');
      });

      it('handles mixed case input correctly', () => {
        expect(titleCase('hElLo WoRlD')).toBe('Hello World');
      });

      it('works correctly with strings containing punctuation', () => {
        expect(titleCase('hello, world!')).toBe('Hello, World!');
      });

      // lol this title doesn't match the assert
      it('retains existing capitalization of non-initial letters', () => {
        expect(titleCase('hElLo wOrLd')).toBe('Hello World');
      });

      it('works correctly with empty strings', () => {
        expect(titleCase('')).toBe('');
      });

      it('handles strings with only one character', () => {
        expect(titleCase('a')).toBe('A');
      });

      it('doesn’t modify non-letter characters', () => {
        expect(titleCase('123 abc')).toBe('123 Abc');
      });
    });
```

app/pages/project/networking/VpcPage/tabs/VpcFirewallRulesTab.tsx
david-crespo: Yep, fix it

david-crespo:
I think we need to undo the checkbox changes like in #1867

app/pages/project/networking/VpcPage/tabs/VpcFirewallRulesTab.tsx
```diff
@@ -74,6 +81,18 @@ export const VpcFirewallRulesTab = () => {
...
     return [
+      colHelper.accessor('name', {
+        header: 'Name',
+        cell: (info) => (
+          <>
+            <button
+              className="peer absolute inset-0"
+              onClick={() => setEditing(info.row.original)}
+            />
+            <div className="peer-hover:underline">{info.getValue()}</div>
```
david-crespo, line 92: Bad news — this trick isn't working for me in Firefox. The button has zero width.

image

benjaminleonard: Can probably replace it with the same trick we're using on the clickability PR

app/pages/project/networking/VpcPage/VpcPage.tsx
```diff
@@ -19,7 +19,15 @@ import { VpcSubnetsTab } from './tabs/VpcSubnetsTab'
...
-  await apiQueryClient.prefetchQuery('vpcView', { path: { vpc }, query: { project } })
+  await Promise.all([
+    apiQueryClient.prefetchQuery('vpcView', { path: { vpc }, query: { project } }),
+    apiQueryClient.prefetchQuery('vpcFirewallRulesView', {
+      query: { project, vpc },
+    }),
+    apiQueryClient.prefetchQuery('vpcSubnetList', {
+      query: { project, vpc, limit: 25 },
+    }),
+  ])
```
david-crespo, line 30: this makes the tabs load instantly, been meaning to fix this

libs/util/str.spec.ts
```diff
@@ -46,3 +46,38 @@ it('commaSeries', () => {
...
+  })
+
+  it('works correctly with strings containing punctuation', () => {
+    expect(titleCase('hello, world!')).toBe('Hello, World!')
+  })
+
+  // lol this title doesn't match the assert
+  it('retains existing capitalization of non-initial letters', () => {
+    expect(titleCase('hElLo wOrLd')).toBe('Hello World')
+  })
```
david-crespo, line 70: delete this or fix the test description. should be "does not retain"

libs/ui/lib/checkbox/Checkbox.tsx
```diff
@@ -43,15 +43,15 @@ export const Checkbox = ({
...
+  <label className="items-top inline-flex">
     <span className="relative h-4 w-4">
       <input
         className={cn(inputStyle, className)}
         type="checkbox"
         ref={(el) => el && (el.indeterminate = !!indeterminate)}
         {...inputProps}
       />
-      {inputProps.checked && !indeterminate && <Check />}
+      {!indeterminate && <Check />}
```
david-crespo, line 54: Not sure any of the changes in this file are needed anymore after d14332cfdce7852a32e1c0089264618a78699550. This line in particular seems dubious.

app/pages/project/networking/VpcPage/tabs/VpcFirewallRulesTab.tsx
benjaminleonard: I can create a `buttonCell` component, but really I think the best solution here would to change it to route tabs, reuse `linkCell`, and allow the edit modal to be linkable.

david-crespo:
On second thought, the route tabs thing is invasive enough that I want to do it in a second PR.

app/pages/project/networking/VpcPage/tabs/VpcFirewallRulesTab.tsx
david-crespo: Will redo this in a followup.

app/forms/firewall-rules-create.tsx
```diff
@@ -259,7 +259,11 @@ export const CommonFields = ({ error, control }: CommonFieldsProps) => {
                 <MiniTable.Cell>
                   <button
                     onClick={() =>
-                      targets.onChange(targets.value.filter((i) => i.value !== t.value))
+                      targets.onChange(
+                        targets.value.filter(
+                          (i) => !(i.value === t.value && i.type === t.type)
+                        )
+                      )
```
david-crespo, line 266: this is a bug you introduced. wait for it...

app/forms/firewall-rules-create.tsx
```diff
@@ -313,11 +317,12 @@ export const CommonFields = ({ error, control }: CommonFieldsProps) => {
           <Button
             size="sm"
             onClick={hostForm.handleSubmit(({ type, value }) => {
+              // ignore click if it's a duplicate
               // TODO: show error instead of ignoring click
               if (
                 type &&
                 value &&
-                !hosts.value.some((t) => t.value === value || t.type === type)
+                !hosts.value.some((t) => t.value === value && t.type === type)
```
david-crespo, line 325: this is a very similar bug that has been there since Oct 2022, preventing anyone from adding two host filters with the same value or the same type (or both, which is what we actually want)

benjaminleonard: I learned from the best

PR #1890, opened 2024-01-16
Associated firewall rules

benjaminleonard:
Ignore the code in this PR, it's more of a discussion point than anything else, since if we did decide to do this it should probably happen at the nexus level. But I took a stab at what it might take to grab all the firewall rules that apply to the current instance.

The logic _seems_ pretty simple – though I'm only looking at the first VPC.

image

The things to consider would be: 
**Is this useful?**
I think so, much easier than looking through each firewall rule and trying to figure out if the target applies to this instance.

**Can you edit items inline here** 
It might give the user the impression that these rules are scoped to this instance. Likewise it might be useful to allow a user to add a firewall rule here but we'd need to pre fill the target as the current instance.

**How do we effectively link from back to the VPC / make it clear that these rules belong to the VPC not the instance?**

This is another case for moving pagination directly under the table itself – since it doesn't support multiple tables on one page. Though it might be better to move firewall rules into its own tab.

---

Does this feel useful to explore? If not, or better to defer we can close this.


david-crespo:
I like the idea. My concern is that we'd be recreating API (or even networking subsystem) logic in an incomplete and possibly misleading way. For example, unless we could be very confident about our logic (not sure this is possible), we would not want to give the impression that the list is exhaustive and these rules are the _only_ way the instance could be receiving traffic — that would be a nightmare for someone trying to debug instance networking. We'd have to include a disclaimer that the list is for convenience.

benjaminleonard:
> I like the idea. My concern is that we'd be recreating API (or even networking subsystem) logic in an incomplete and possibly misleading way. For example, unless we could be very confident about our logic (not sure this is possible), we would not want to give the impression that the list is exhaustive and these rules are the _only_ way the instance could be receiving traffic — that would be a nightmare for someone trying to debug instance networking. We'd have to include a disclaimer that the list is for convenience.

That's fair, and perhaps the case even if implemented in `omicron`. As an example, this would need to change when we have VPC peering, and you could imagine that might get missed.

zephraph:
Overall I _do_ like bringing like concerns closer together. 

> My concern is that we'd be recreating API (or even networking subsystem) logic in an incomplete and possibly misleading way

I think this is an excellent point. Further, our differentiator (tight hardware/software co-design) _should enable this exact kind of interaction_. Now, that doesn't mean doing this kind of thing would be easy, but it is compelling. 

@benjaminleonard I would definitely want to pull in someone with more networking experience to be able to have an inform conversation about what the possibilities are here. Overall though, I think if you _could_ get a list of all firewall rules associated to an instance then that would be a tremendously valuable experience.

---

One thing I'd recommend is having a `parent` column or something that points back to the owning VPC. I think that would clear up the ownership questions. 

---

I think being able to add a firewall rule here would be nice too. It's obviously of secondary importance _and_ we should be 10000% sure that the user understands where they're adding it and _if it impacts anything else_. So, while it would be nice, we should only do it if it doesn't generate more confusion for users.

FelixMcFelix:
I think this is very useful in principle for helping to visualise and debug a user's network configuration! Looking at it rendered:
* I think it would be useful to highlight which part(s) of a rule are responsible for the match, if possible. Incoming/outgoing might also be worth displaying, as it's probably more necessary for contextualising these rules when viewed on an individual instance.
* I don't know if there might also be some value in being able to see which rules *don't* affect an arbitrary instance, from e.g. the VPC page? I think that's harder to get right though and I'm just spitballing.

> My concern is that we'd be recreating API (or even networking subsystem) logic in an incomplete and possibly misleading way

I had another look over features outlined in RFD21 that we don't have yet (VPC subnet routing, VPC Peering) -- I think you're right and there should 100% be a disclaimer for the reason that, e.g., custom VPC routers would be able to filter or redirect traffic such that it'd be difficult to resolve the exact set of relevant rules. I initially thought that as long as we didn't perform any excess pruning we'd have false positive FW rules at worst, but I can imagine at least one scenario where you'd miss rules *unless* you resolved custom router logic. I could be misinterpreting though.

> One thing I'd recommend is having a `parent` column or something that points back to the owning VPC. I think that would clear up the ownership questions.

Definitely necessary if/when VPC peering is implemented.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -256,6 +269,145 @@ export function NetworkingTab() {
...
+        for (const ip of ips) {
+          if (ipRangeCheck(ip, target.value)) {
+            return true
+          }
+        }
+        break
+    }
+  }
+  return false
+}
```
david-crespo, line 413: Didn't see this function till now. Very fun. Makes me think maybe this can work.

david-crespo:
Closing for now since this was just a PoC. A great one, though!

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
sion42x: We should probably use .exhaustive() instead of a switch. If we add future variants we don't want it to silently fail. Exhaustive makes sure it fails at compile time so we can update it.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -256,6 +269,145 @@ export function NetworkingTab() {
...
+        }
+        break
+      case 'instance':
+        if (instanceName === target.value) {
+          return true
+        }
+        break
+      case 'ip':
+        if (ips.includes(target.value)) {
+          return true
```
sion42x, line 400: Should parse/compare similar to ipRangeCheck. A string inclusion will work fine for IPv4 but have problems for IPv6.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -256,6 +269,145 @@ export function NetworkingTab() {
...
+        }
+        break
+      case 'subnet':
+        if (subnets.includes(target.value)) {
+          return true
+        }
+        break
+      case 'instance':
+        if (instanceName === target.value) {
```
sion42x, line 394: Assumes a name, but the PR/line 347 indicate it can be an ID.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
benjaminleonard: Haven't touched these since last year. Though probably some evidence that if we did decide to do this, it should sit within the API rather than the console where it can drift (more easily).

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -256,6 +269,145 @@ export function NetworkingTab() {
...
+  const { data: subnet } = useApiQuery(
+    'vpcSubnetView',
+    { path: { subnet: nicList?.items[0].subnetId as string } },
+    { throwOnError: false, enabled: nicList && nicList?.items.length > 0 }
+  )
+
+  if (vpc && firewallRules && firewallRules.rules && subnet) {
+    const vpcs = [vpc.name]
+    const subnets = [subnet.name]
+    const ips = nicList && nicList.items.length > 0 ? [nicList?.items[0].ip] : []
```
sion42x, line 344: Why collapse the nicList to only 1? We pull up to 10 and `isRuleApplicableToInstance` can handle the inputs on a loop. Also... can an instance have two NICs across VPCs? I'll have to look at that.

Issue #1773, opened 2023-10-05
Validate disk size against image size on disk create form


david-crespo:
On instance create, we validate the size of the boot disk against the size of the chosen image, and when you change the image, we update the size to be big enough for it if necessary.

https://github.com/oxidecomputer/console/blob/0cc1e03a24b3f5da275d15b969978a385d6b3b27/app/forms/instance-create.tsx#L331-L341

On the standalone disk create form, we don't do that, and we should.

https://github.com/oxidecomputer/console/blob/0cc1e03a24b3f5da275d15b969978a385d6b3b27/app/forms/disk-create.tsx#L96

david-crespo:
This could be worth picking up @charliepark if you become an expert on this while working on #1823.

PR #1891, opened 2024-01-17
Update disk size from image (and snapshots)

benjaminleonard:
Started looking at #1773 

We were already doing the validation, just not the disk size update which happens inside of the image listbox.

Whilst I was adding that I noticed that we weren't doing the same behaviour for snapshots, which I'm guessing we didn't notice because the values are so small. In practice I'm not sure how large these snapshots can get, but it's a small tweak to match the functionality it both.

Also a question of `GiB` versus a dynamic formatted size. Is it likely that image sizes get large enough to make `TiB` a good option? I suppose a image in `MiB` is realistic. I could switch that over to render with `filesize` too.

Basically ready to go but I need to do some more checks to make sure it's working as it should.


app/forms/disk-create.tsx
```diff
@@ -225,22 +249,33 @@ const SnapshotSelectField = ({ control }: { control: Control<DiskCreate> }) => {
...
       items={snapshots.map((i) => {
+        const formattedSize = fileSize(i.size, { base: 2, output: 'object' })
         return {
           value: i.id,
           labelString: `${i.name}`,
           label: (
             <>
               <div>{i.name}</div>
-              <div className="text-secondary">
+              <div className="text-tertiary selected:text-accent-secondary">
```
benjaminleonard, line 259: Sneaking in a fix for the selected state not rendering correctly.

benjaminleonard:
Adding this to v7 milestone. @charliepark if you do get a sec to review this that'd be great!

charliepark:
It seems like we're able to get into the state where there's a "negative" size for the disk we're creating. Just noting it here for now and will try to look into it in a bit, but wanted to mention it, as I thought we'd fixed this elsewhere.
Screenshot 2024-03-15 at 5 09 41 PM

benjaminleonard:
Just to confirm this isn't happening automatically? It's just the field doesn't prevent going below zero?

This is quite an old PR – I'll merge main into it and see if that fixes.

charliepark:
Ah, yup. That must have been it. Looks good!

charliepark (approved):

PR #1893, opened 2024-01-18
Freeze the Create Instance form when submitting, so the user can't make post-submit edits

### LINKED ISSUE #1613: Editing instance settings while it's being created
It's possible to edit instance settings while it's in the process of being created.

If you do that, then when the instance is created, you get a "your changes won't be saved" popup, which is very confusing.

https://github.com/oxidecomputer/console/assets/745333/5697cfbf-72bc-4bea-ac23-accbd18689c8

charliepark:
Fixes #1613

In https://github.com/oxidecomputer/console/issues/1613, we see that it's currently possible for a user to edit the Create Instance form after they hit the submit button, and — as the process can sometimes take a bit of time — that this can create confusion in the user about whether last-second edits will be accepted, and, maybe more problematic, a confusing modal appears asking the user if they want to navigate away from the page.

This PR disables the form fields when the user hits the submit button, so edits can't take place. It also disables the tabs that would otherwise allow the user to switch between instance sources (silo vs. project) and switching between the pre-configured sizes for CPU and storage.

One thing I'm not sure about how to elegantly handle is that we have it setting the `useState` value to `isSubmitting` when the form is submitted, but there isn't a way to re-set that to `false` if something blows up in the process. I suspect there's some sort of event that happens when an error is thrown, but am still looking into that.

The only other thing to call out, I think, is that I had to remove a CSS class, which normally affects a selected-but-disabled form field's text color. Note the removal of `peer-disabled:text-secondary` in the code below.

With that class present (as it is currently), when the form fields are disabled, the selected option looks like this:
Screenshot 2024-01-17 at 2 05 06 PM

By removing the class, the _disabled_ state for the selected option …
Screenshot 2024-01-17 at 2 04 40 PM
… looks more like the _normal_ state for the selected option:
Screenshot 2024-01-17 at 2 04 27 PM
(with the disabled state (top) having slightly thinner outlines)

I believe this is doing what we want, but know there are other situations where removing that class might not be what we want? "Selected but disabled" is a pretty rare situation, though, so I think this class possibly fine to drop.


app/components/form/fields/DiskSizeField.tsx
```diff
@@ -57,6 +59,7 @@ export function DiskSizeField<
         // Run any additional validators passed in from the callsite
         return validate?.(diskSizeGiB)
       }}
+      disabled={disabled}
```
david-crespo, line 62: it's already spread as part of `{...props}` if you delete this line and 37

david-crespo: Actually, it looks like `TextFieldProps` already has `disabled?: boolean` on it (through a chain of inheritances I don't love). So you could get rid of all three lines in this file.

https://github.com/oxidecomputer/console/blob/644a45b8e4ab673ad51754e372277abc3ddfd036/app/components/form/fields/TextField.tsx#L30-L33

https://github.com/oxidecomputer/console/blob/644a45b8e4ab673ad51754e372277abc3ddfd036/app/components/form/fields/TextField.tsx#L24

https://github.com/oxidecomputer/console/blob/644a45b8e4ab673ad51754e372277abc3ddfd036/libs/ui/lib/text-input/TextInput.tsx#L35-L38

Amusingly it doesn't even need to be there, it seems, because it is already on the native `input` props. The code typechecks fine with it removed.

image

app/components/form/fields/DisksTableField.tsx
```diff
@@ -81,16 +87,31 @@ export function DisksTableField({ control }: { control: Control<InstanceCreateIn
...
         <div className="space-x-3">
-          <Button size="sm" onClick={() => setShowDiskCreate(true)}>
+          <Button
+            size="sm"
+            onClick={() => {
+              if (!disabled) {
+                setShowDiskCreate(true)
+              }
+            }}
+          >
```
david-crespo, line 97: `Button` has a `disabled` prop that seems to work ok and gets you disabled styling too.

david-crespo:
Look great overall, thanks for doing this. I also wondered about how to do a `setSubmitting(false)` at the right time and I wish I had an easy answer for how to do it. First thing that comes to mind is wrapping the `mutate` in a try/catch and doing it in the catch? Kinda ew, and you might have to change it to `await createInstance.mutateAsync(...)`. Another way would be to use a `useEffect` that looks for when `createInstance.error` goes from nothing to something.

app/forms/instance-create.tsx
```diff
@@ -318,7 +336,11 @@ export function CreateInstanceForm() {
               />
```
david-crespo, line 336: You should disable the tab triggers for silo/project images like you did on the instance size ones.

app/components/form/fields/DiskSizeField.tsx
charliepark: good catch!

app/components/form/fields/RadioField.tsx
```diff
@@ -168,6 +169,7 @@ export function RadioFieldDyn<
             aria-describedby={description ? `${id}-label-tip` : undefined}
             onChange={onChange}
             name={name}
+            disabled={disabled}
```
david-crespo, line 172: You can probably get rid of these too. But not the one in `NumberField`, I guess.

app/components/form/fields/DisksTableField.tsx
```diff
@@ -99,7 +110,7 @@ export function DisksTableField({ control }: { control: Control<InstanceCreateIn
           onDismiss={() => setShowDiskCreate(false)}
         />
       )}
-      {showDiskAttach && (
+      {showDiskAttach && !disabled && (
```
david-crespo, line 113: This means that if you have the modal open and click submit, it will close, but if there's an error and we land back on the form (however we do that) it will pop back open?

charliepark: I believe that is what should happen, as it isn't affecting the showDiskAttach (or Create) values. Will try testing that scenario to verify.

charliepark: Turns out you can't even submit the form with the modals open, as they lose focus and disappear. So I've removed the `!disabled` from those two spots, as they're moot.

app/forms/instance-create.tsx
```diff
@@ -138,6 +140,12 @@ export function CreateInstanceForm() {
   const image = allImages.find((i) => i.id === imageInput)
   const imageSize = image?.size ? Math.ceil(image.size / GiB) : undefined
 
+  useEffect(() => {
+    if (createInstance.error) {
+      setIsSubmitting(false)
+    }
+  }, [createInstance])
+
```
david-crespo, line 148: Clean! Might be better to put `createInstance.error` directly in the deps instead of `createInstance`. I don’t know the behavior of RQ off the top of my head with respect to the top level object.

charliepark: Checked it out (turns out if you just use an existing instance name you can trigger the error state) and `createInstance.error` works well there.

david-crespo (approved):
Very nice! Not as complicated as I expected.

Not sure about the `peer-disabled:text-secondary` thing. @benjaminleonard?

david-crespo:
Ooh, there is an `isSubmitting` on RHF `formState`. Worth checking if you can use that instead of the explicit state variable before merge. If it doesn't have the desired behavior (e.g., fails to go back to `false` as desired on submit error).

https://react-hook-form.com/docs/useform/formstate

david-crespo:
Sorry, never mind on that. It doesn't do what we want. I set the request to take 4 seconds. If you look at the log before and after the 400 comes back, you can see the 4 seconds go by while `formState.isSubmitting` is `false` rather than the desired `true`.

image

benjaminleonard:
I think we're okay on the `peer-disabled:text-secondary` – I'll take a look and can follow up if there's an issue.

PR #1894, opened 2024-01-18
Prettier -> Biome

david-crespo:
Runs on the entire console in half a second instead of 2 seconds (warm cache) to 8 seconds (no cache). Runs on a single file _extremely_ fast.

```
$ cat app/main.tsx | time biome format --stdin-file-path=file.tsx
...
biome format --stdin-file-path=file.tsx  0.00s user 0.00s system 82% cpu 0.009 total
```

Prettier's result is... grotesque:

```
$ cat app/main.tsx | time prettier --parser typescript
prettier --parser typescript  1.32s user 0.12s system 117% cpu 1.226 total
```

Leaving as draft because I doubt it will be worth the time required to get this going on all our apps, which to me is the only way it makes sense.

Things missing:

- Tailwind class sorting (https://github.com/biomejs/biome/discussions/164)
- Import sort and group (exists but it's part of the linter, not the formatter)

There are other possible issues:

- Not sure what you have to do to make VS Code use this instead of Prettier. They have a [guide](https://biomejs.dev/guides/integrate-in-editor/), but it only covers setting Biome as your default formatter. What I'd really want is for it to use Biome when the app has a Biome config and Prettier when it has a Prettier config. Seems kind of obvious.
- Using a different formatter here from the other webapps is silly, so we’d probably want to aim to use this on all of them


zephraph:
So the tailwind import sorting is being actively worked on: https://github.com/biomejs/biome/pull/1362. The import sorting is real, but perhaps there's something we can do to bridge the gap there. 

Regardless, I like this change. I'd really like to do it. It would be worth creating a draft in at least one other project (the RFD site?) to see if there's any significant delta.

zephraph:
As for import sorting... it does look like they have auto formatting options (even if it's logically grouped with the linter): https://biomejs.dev/analyzer/#imports-sorting

david-crespo:
I don’t think we can justify doing this work now. Let’s wait to see how the performance work in Prettier 4 pans out.

PR #1897, opened 2024-01-22
Improve disabled radio state

benjaminleonard:
Tweaked the disabled radio state to match the other inputs, change cursor, hover no longer changes border colour.

CleanShot 2024-01-22 at 12 26 44

That class is getting pretty complex, but unsure on how we might clean it up. I ended up grouping by state and type.


david-crespo (approved):
Looks great. The class is getting ridiculous. Don't tell the Tailwind haters.

PR #1898, opened 2024-01-22
Add `user_data` field to instance create

### LINKED ISSUE #800: [Instance create] `user_data` textarea/file input
[Discussed in chat](https://matrix.to/#/!YNYPOVxjAUeXksTcRj:oxide.computer/$U3fCENcgUYFVwnXh3zyy39m6YF2D1kMbC612uFnNFVE?via=oxide.computer&via=unix.house&via=timecube.club). This is not fully fleshed out, but it's [in the API](https://github.com/oxidecomputer/omicron/pull/911). `user_data` is an optional string field on the instance create POST. The easy version is a text area you can paste into. The fancy version would also allow a file upload. This is sort of an advanced feature on instance create, so we would probably want to hide it in a collapsed section.

benjaminleonard:
Fixes #800 as per: https://github.com/oxidecomputer/console/issues/800#issuecomment-1895534344

CleanShot 2024-01-22 at 15 00 34

Works fine with the mock API, we should test against dogfood.
We're just reading their file as a string and converting into a base64 string using the same mechanism as cert upload. Do we want to constrain the files a user can upload?


app/components/form/FullPageForm.tsx
```diff
@@ -81,7 +81,7 @@ export function FullPageForm<TFieldValues extends FieldValues>({
       <form
         className="ox-form pb-20"
         id={id}
-        onSubmit={(e) => {
+        onSubmit={async (e) => {
```
david-crespo, line 84: My understanding is if there are no `await`s in the function, this shouldn't do anything. Does it make a difference for you?

benjaminleonard: Hmm, will it handle the `await readBlobAsBase64` without that?

david-crespo: Pretty sure it is ok. @charliepark @zephraph let me know if you think any of this is not correct.

This is a kind of evil thing about `async/await` in JS, though it's sort of a nice thing: all JS code is already asynchronous, basically. You can call `setTimeout` anywhere, execute a function that returns a promise anywhere. All `async` on a function means is that you might call `await` directly inside that function, directly in the sense of literally right there in the definition of the function. It's primarily a syntactic thing.

The instance-create `onSubmit`, where the `await readBlobAsBase64` takes place, _is_ marked `async`, necessarily so because you are using `await` in there. The tricky thing to understand is that the `await` aspect of that is only about sequencing events _within_ that function. What the `await` is primarily doing there is ensuring that `createInstance.mutate()` does not happen until after `readBlobAsBase64` is done. But that is about sequencing inside the function. Because you are not calling `await` inside of `FullPageForm` `onSubmit` (plus `handleSubmit` is the last thing in the function anyway), there is no sequencing question. This function ends and whatever async ops are put on the queue are just going to do whatever they do — we don't don't wait for them to finish, and whatever calls this function also doesn't wait.

Now... `form.handleSubmit` does return a promise, so you actually could `await` on line 94, and then you would have to keep the `async`. But I'm not sure whether it accomplishes anything. In order for it to matter, there would have to be something waiting for this `onSubmit` to finish, but since this is being passed to native `<form>`, I doubt it. RHF is not doing anything like that. It's all regular props and function calls.

david-crespo:
This looks good to me functionally. The tooltip/help text could use workshopping. 

image

For one thing, if we're going to have help text anyway, I don't think we need the tooltip. The whole section is already collapsed under `Advanced`. Might as well put all the text in the help text. 

More importantly, though, we should explain better what this content is supposed to be (the API doc this is based on could use the same improvement). As discussed in chat, it would make things a lot easier if we could be specific an say "this is about cloud-init" rather than "instance initialization systems (such as cloud-init)". Just so I'm not complaining without making a constructive suggestion, here is some text that is IMO at least a marginal improvement on what's there:

> Data or scripts to be passed to cloud-init as [user data](https://cloudinit.readthedocs.io/en/latest/explanation/format.html) ([examples](https://cloudinit.readthedocs.io/en/latest/reference/examples.html#yaml-examples)) if the selected boot image supports it.

In lieu of our own doc explaining this at length, this at least gives them _something_. I'm on the fence about whether to mention that the content should be plain text.

benjaminleonard:
Thanks for the feedback! Added the links object and test that checks that the external links are accessible – I could pull that out into its own PR if that's helpful.

Its become a bit of a mishmash because I've changed `helpText` -> `description` and `description` to `tooltipText` as discussed. That can also be picked into its own PR to merge into this one.

app/util/links.ts
```diff
@@ -0,0 +1,18 @@
...
+  cloudInitFormat: 'https://cloudinit.readthedocs.io/en/latest/explanation/format.html',
+  cloudInitExamples:
+    'https://cloudinit.readthedocs.io/en/latest/reference/examples.html#yaml-examples',
+}
+
+const links = {
+  external,
+}
+
+export default links
```
david-crespo, line 18: I should put in a lint rule or something about this — I think the best practice is to avoid default exports. Bundlers don't like them, something like that. They don't really do anything useful anyway. For now you can also collapse this down, we don't need to distinguish external. I'd just do `export const links = { cloudInitFormat: ... }`.

david-crespo:
Decided the fallback to GET is overkill since the 2 links we have both work with HEAD and it's possible other ones will for the foreseeable future.

PR #1899, opened 2024-01-22
Make links in tables more clearly clickable

benjaminleonard:
Follow up from #1886 and feedback from the team.

- Whole row hover state (helps distinguish between rows)
- Link cells are now the same style universally
- Fixed disk cell that was styled differently

CleanShot 2024-01-22 at 15 56 19

I'd would have liked to have the first cell be `text-default` but unsure what to do with the hover state in that instance since we don't have a colour brighter than that.

### DIFF
--- a/app/pages/project/disks/DisksPage.tsx
+++ b/app/pages/project/disks/DisksPage.tsx
@@ -16,7 +16,7 @@ import {
   useApiQueryClient,
   type Disk,
 } from '@oxide/api'
-import { DateCell, SizeCell, useQueryTable, type MenuAction } from '@oxide/table'
+import { DateCell, linkCell, SizeCell, useQueryTable, type MenuAction } from '@oxide/table'
 import {
   buttonStyle,
   EmptyMessage,
@@ -43,14 +43,12 @@ function AttachedInstance({
   const { data: instance } = useApiQuery('instanceView', {
     path: { instance: instanceId },
   })
-  return instance ? (
-    <Link
-      className="text-sans-semi-md text-accent hover:underline"
-      to={pb.instancePage({ ...projectSelector, instance: instance.name })}
-    >
-      {instance.name}
-    </Link>
-  ) : null
+
+  const instanceLinkCell = linkCell((instanceName) =>
+    pb.instancePage({ ...projectSelector, instance: instanceName })
+  )
+
+  return instance ? instanceLinkCell({ value: instance.name }) : null
 }
 
 const EmptyState = () => (
--- a/libs/table/cells/LinkCell.tsx
+++ b/libs/table/cells/LinkCell.tsx
@@ -14,12 +14,12 @@ export const linkCell =
   ({ value }: Cell<string>) => {
     return (
       <Link
-        className="flex h-full w-full items-center underline text-sans-semi-md text-secondary hover:text-default"
+        className="link-with-underline group flex h-full w-full items-center text-sans-semi-md"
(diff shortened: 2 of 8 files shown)


zephraph:
I'm noticing in the gif that the more action button gets a slightly different color than the other row cells on hover. Could we make that consistent? I like the idea of the row being highlighted on hover, but it's _also_ nice if when a cell is clickable it's hover color is differentiable. The underlines are nice too as a persistent reminder.

app/pages/project/disks/DisksPage.tsx
```diff
@@ -43,14 +43,12 @@ function AttachedInstance({
...
-    >
-      {instance.name}
-    </Link>
-  ) : null
+
+  const instanceLinkCell = linkCell((instanceName) =>
+    pb.instancePage({ ...projectSelector, instance: instanceName })
+  )
+
+  return instance ? instanceLinkCell({ value: instance.name }) : null
```
zephraph, line 51: Nice, happy to see the reuse of linkcell here

libs/table/cells/LinkCell.tsx
```diff
@@ -14,7 +14,7 @@ export const linkCell =
   ({ value }: Cell<string>) => {
     return (
       <Link
-        className="flex h-full w-full items-center text-sans-semi-md text-default hover:underline"
+        className="flex h-full w-full items-center underline text-sans-semi-md text-secondary hover:text-default"
```
zephraph, line 17: Is `hover:bg-raise` to distracting here?

david-crespo: I was kinda thinking the same

benjaminleonard: Happy to remove

benjaminleonard: "bg-hover" is too strong, and my concern was that "bg-raise" was inconsistent.

My rationale for more actions and the link cell having a different treatment is partly the distinction between a button and a link but that's perhaps a weak argument.

I'll collect our options and drop them in here and we can decide which is working the best.

zephraph: Yeah, my preference would be to treat clickable things consistently, regardless of what kind of clickable thing they are. I feel like `bg-raise` does a good job of calling extra attention which (when combined with a pointer) is a doubly helpful indication.

benjaminleonard:
Few options:

1. Both links and more actions button have filled contained
CleanShot 2024-01-29 at 12 53 04

2. `bg-raise` on hover across entire cell
CleanShot 2024-01-29 at 12 54 29

3. Same but icon has it's own fill for emphasis
CleanShot 2024-01-29 at 12 57 02

@zephraph @david-crespo @paryhin

david-crespo:
None are bad but I think 2 is most natural. Like Eugene said,the permanent underline on links is most accessible, even though it feels a little noisy.

benjaminleonard:
> None are bad but I think 2 is most natural. Like Eugene said, the permanent underline on links is most accessible, even though it feels a little noisy.

Agreed. On the underline. `text-decoration-color` could be useful to reduce some of that noise:

CleanShot 2024-01-29 at 14 51 45

david-crespo (approved):
Tried it in the preview and it's honestly fantastic. Didn't realize what a difference it would make.

benjaminleonard:
Perfect, just spotted a little stroke overlap I need to fix then will merge.

PR #1908, opened 2024-01-24
Fix file disappearing in userData field

### LINKED ISSUE #1907: `userData` uploaded file disappears when accordion is closed
Guessing it's still in the form state just not in the `FileField`.

Wondering if it's a controlled input state thing or we force radix to keep the inputs mounted.

benjaminleonard:
Fixes #1907 

For reasons outlines in that issue, CSS animation had become troublesome.

Here we keep the accordion mounted so never lose the file state.

Will defer to you all on whether this feels like a better fix than controlling the file input.


benjaminleonard:
This has climbed in complexity a little, which might be a case for use controlling the file input. Perhaps the other question here is whether we might have some other input state that becomes more awkward after demounting and remounting.

benjaminleonard:
Will defer to y'all on the best way to tackle this – this PR or a controlled input.

If we want a quick uncontroversial fix I can just disable the animation and add `forceMount` and we can revisit in the next release.

app/forms/instance-create.tsx
```diff
@@ -468,6 +432,56 @@ export function CreateInstanceForm() {
...
 }
 
+const AdvancedAccordion = ({
+  control,
+  isSubmitting,
+}: {
+  control: Control<InstanceCreateInput>
+  isSubmitting: boolean
+}) => {
+  const [value, setValue] = useState<string[]>([])
```
david-crespo, line 442: does it still need this advancedness to solve the original problem?

benjaminleonard: Advanced is the name of the accordion. This was just a general refactor to make it easier and neater to add to this section in the future.

david-crespo: I'm talking about the `useState`, which wasn't there before.

benjaminleonard: My bad, you're right I can do this all with css

benjaminleonard: ...or can I. We need `isOpen` to do the auto scrolling when you open an accordion I believe.

david-crespo:
I feel good about this fix as-is.

I do think there's a good chance controlling the file input would be better, as you say — it feels like the right structure from the form's POV. Without this fix, the form state value does stay set — the problem is about initializing the state of the file input when in pops into existence as the accordion expands.

A counterpoint is that we still need to track `isOpen` in order to trigger `scrollIntoView` when the thing expands. I'll give the controlled input thing a try, and if it feels bad I'll just merge this since it's good to go.

PR #1909, opened 2024-01-24
bump omicron version

charliepark:
Bumping the omicron version to pick up some of the Silo IP Pool updates.


charliepark:
handlers is missing an endpoint that we'll need; looking into that

PR #1911, opened 2024-01-25
Fix missing key on intersperse map

benjaminleonard:
Fixes error on `InstancesPage` with missing key on map


david-crespo:
I don’t think this works because the commas all have the same key

benjaminleonard:
How about this one?

david-crespo:
That will definitely shut React up, at least. Let me give it one more try before we merge this.

david-crespo:
Covered in #1912

PR #1910, opened 2024-01-25
Manage IP pools

david-crespo:
- [x] List all IP pools
- [x] List IP pools for silo on silo detail page
- [x] Make/clear default on IP pool for silo
- [x] Move fleet role mapping into tab on silo detail
- [x] Add properties table to silo detail
- [x] Create and delete IP pool from list view
- [x] List IP ranges for a pool
- [x] Edit IP pool from list view
- [x] Epic e2e tests for everything
- [x] Link to IP pools docs _somewhere_
- [x] Confirm modal on row actions
- [x] Link pool from silo pools list and link silo from pool silos list

Not planned/followup work: add/remove IP ranges in pool

https://github.com/oxidecomputer/console/assets/3612203/bfa08d46-6fc1-435f-95c4-3de6613aa14f

PR #1916, opened 2024-01-27
Dynamically render units on metrics tab

### LINKED ISSUE #1866: Instance disk metrics should not be displayed in bytes
I'm seeing raw counts of things like number of/bytes for reads/writes. I'm assuming this is not the design intent given those raw numbers aren't really that useful and it also is not how we display `/utilization`.

image

Is there supposed to be some sort of dynamic unit selection here?

Also, when I resize my browser those Y-axis labels continue to run outside the box (sorry I don't know the technical word here). I guess I'd have expected the chart to scale more dynamically, perhaps some % of new width would be allocated to the axis labels and the rest to the graph itself.

charliepark:
Fixes #1866.

We're pulling in data for the instance page's metrics tab, showing read / write counts and size, and the number of flushes. Currently, we aren't adjusting the numbers, and are just showing the raw count / bytes.

This PR updates how we render the data. Because the data is cumulative, we can use the largest data value to determine the unit of measurement for the set of data. We then divide the given datapoint by the appropriate divisor and pass that data along to the chart.

I'm open to suggestions for better ways to label the overall chart.

While it's fine on datapoints using Byte counts …
Screenshot 2024-01-26 at 3 36 21 PM

… it's a little meh on Count charts …
Screenshot 2024-01-26 at 4 27 08 PM
(The thing you're supposed to be looking at is at the very top of the chart, currently showing "count x K".)

@benjaminleonard if you have any thoughts, I'm all ears.

### DIFF
--- a/app/pages/project/instances/instance/tabs/MetricsTab.tsx
+++ b/app/pages/project/instances/instance/tabs/MetricsTab.tsx
@@ -85,10 +85,12 @@ function DiskMetric({
     }
   }
 
+  const divisor = divisorBase ^ cycleCount
+
   const data = (metrics?.items || []).map(({ datum, timestamp }) => ({
     timestamp: timestamp.getTime(),
     // all of these metrics are cumulative ints
-    value: (datum.datum as Cumulativeint64).value / (divisorBase ^ cycleCount),
+    value: (datum.datum as Cumulativeint64).value / divisor,
   }))
 
   return (
--- a/app/pages/project/instances/instance/tabs/MetricsTab.tsx
+++ b/app/pages/project/instances/instance/tabs/MetricsTab.tsx
@@ -24,7 +24,7 @@ const TimeSeriesChart = React.lazy(() => import('app/components/TimeSeriesChart'
 
 type DiskMetricParams = {
   title: string
-  unit?: string
+  unit: 'Bytes' | 'Count'
   startTime: Date
   endTime: Date
   metric: DiskMetricName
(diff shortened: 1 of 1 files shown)


charliepark:
In the latest commit I pulled in @benjaminleonard's suggestions for that Y-Axis labeling vs. popup number.
Screenshot 2024-01-29 at 1 04 53 PM
Screenshot 2024-01-29 at 1 06 00 PM

benjaminleonard:
Wondering if there's some logic we can write, something like a find comfortable axis function – that given a range and a number of desired steps quantises to multiples. In this case we'd say snap to 0.5k and it'd give us 7 steps up to 3.5k.

We have a similar issue with the `KiB` / `MiB` / `GiB` where they aren't universally divisible in a neat way.

Not to blow up the complexity of this PR. You could merge this and we could tackle in a follow up.

david-crespo:
Yeah, I was thinking that is the main outstanding issue here — would be much nicer if those were round numbers. I will think about it a bit.

benjaminleonard:
Actual implementation requires some experimentation to get something more robust, but in the realm of this might work:

```ts
function findAxis(range: number, targetSteps: number, snap: number): number[] {
    let stepSize = Math.ceil((range / targetSteps) / snap) * snap;
    let steps = Math.ceil(range / stepSize);

    let axis: number[] = [];
    for(let i = 0; i <= steps; i++) {
        axis.push(i * stepSize);
    }

    return axis;
}

console.log(findAxis(3.3, 7, 0.5));
// [0, 0.5, 1, 1.5, 2, 2.5, 3, 3.5]
console.log(findAxis(10, 6, 0.5));
// [0, 2, 4, 6, 8, 10]
```

app/pages/project/instances/instance/tabs/MetricsTab.tsx
```diff
@@ -54,27 +54,81 @@ function DiskMetric({
...
+    if (isBytesChart) {
+      const byteUnits = ['Bytes', 'KiB', 'MiB', 'GiB', 'TiB']
+      unitForSet = byteUnits[cycleCount]
+      label = `(${unitForSet})`
+    } else {
+      label = '(COUNT)'
+    }
+  }
+
+  const divisor = divisorBase ^ cycleCount
```
david-crespo, line 88: I think this is supposed to be `**`. `^` is bitwise XOR. I'm seeing values of 1001 and 1025 for `divisor`.

image

david-crespo:
So.. we are already doing work to get clean round numbers for the ticks. We pick a `maxY` that's divisible by the number of ticks, and then we get ticks evenly spaced out between 0 and `maxY`. 

https://github.com/oxidecomputer/console/blob/764f731013c659a14f857dac731fdef945076ac8/app/components/TimeSeriesChart.tsx#L133-L147

https://github.com/oxidecomputer/console/blob/764f731013c659a14f857dac731fdef945076ac8/app/components/TimeSeriesChart.tsx#L42-L45

So I think all we need to do to avoid the decimals is pass `TICK_COUNT * <relevant power of 10>` to `roundUpToDivBy`. The hard bit is figuring out the power of 10. Or it would be if you hadn't already done it — that's what divisor is. If I pass in `divisor` to `TimeSeriesChart` and do this `roundUpToDivBy(dataMax * 1.2, TICK_COUNT * divisor)` (and turn the `toFixed(2)` into `toFixed(0)` in `yAxisTickFormatter`) I am able to get the following:

image

That confirms I'm thinking about this roughly the right way. However, the bytes numbers are pre-divided while the counts are not, and this messes up the bytes charts, which don't want their axis multiplied by 1024:

image

So I'll need to figure out how to reconcile this.

david-crespo:
Not sure how this happens (we should look into it) but it seems we cannot assume these numbers are monotonically increasing and therefore the last value is the largest.

https://oxide.sys.rack2.eng.oxide.computer/projects/trey-trey-trey/instances/apache/metrics

image

david-crespo:
I think that's this: https://github.com/oxidecomputer/omicron/issues/1624

I knew it sounded familiar.

david-crespo:
While we could improve this, it's clearly an improvement on what's there and works well with big numbers (tested by changing the fake data generator). I want to wrap up #1910 for the release, so I am good with merging this and fixing it up later.

Screenshot 2024-01-29 at 10 55 32 PM

Screenshot 2024-01-29 at 10 55 38 PM

app/pages/project/instances/instance/tabs/MetricsTab.tsx
```diff
@@ -54,27 +64,77 @@ function DiskMetric({
...
-    timestamp: timestamp.getTime(),
-    // all of these metrics are cumulative ints
-    value: (datum.datum as Cumulativeint64).value,
-  }))
+  const isBytesChart = unit === 'Bytes'
+
+  const largestValue = useMemo(() => {
+    if (!metrics || metrics.items.length === 0) return 0
+    return Math.max(...metrics.items.map((m) => (m.datum.datum as Cumulativeint64).value))
+  }, [metrics])
```
david-crespo, line 72: I fixed the last value not necessarily being the largest by looking at all values and finding the max. I always feel weird about spreading 3000 args into `Math.max` but I tested it a while ago (we do this elsewhere) and it is apparently fine.

charliepark:
There are a few more enhancements we've talked about in chat, but we'll merge this PR and then decide on prioritization and path forward on general metrics improvements separately.

PR #1917, opened 2024-01-28
Tweak CSS to remove awkward line wrapping; widen select box

### LINKED ISSUE #1905: Awkward string wrapping inside dropdowns
Screenshot 2024-01-23 at 9 37 26 AM

Visible [here](https://oxide.sys.r3.oxide-preview.com/projects/charlie/instances/augustus-clone/metrics)

charliepark:
Fixes #1905 

Currently, we have some awkward line wrapping when the disk or silo name is too long, as seen here:
image

This PR adds a few CSS rules to prevent line wrapping for the ListBox select fields.
Screenshot 2024-01-28 at 10 02 49 AM

The full text of the option is still shown within the dropdown itself
Screenshot 2024-01-28 at 10 02 58 AM

This PR also expands the width of the dropdown from `w-48` to `w-80`, but that change isn't strictly necessary, if there's a reason we want to keep the `w-48` width in place. Or we could compromise with `w-64`. I thought it might be an issue where `w-48` works better for a mobile screen with the other elements of the metrics view, but even at `w-48` there are some collisions and overflow issues, so I don't think it's that. Regardless, if we want to leave the width as-is, fine.

### DIFF
--- a/app/pages/SiloUtilizationPage.tsx
+++ b/app/pages/SiloUtilizationPage.tsx
@@ -82,7 +82,7 @@ export function SiloUtilizationPage() {
       <div className="mb-3 mt-8 flex justify-between gap-3">
         <Listbox
           selected={filterId}
-          className="w-48"
+          className="w-80"
           aria-labelledby="filter-id-label"
           name="filter-id"
           items={projectItems}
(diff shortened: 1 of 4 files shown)


david-crespo (approved):
Looks clearly better to me, but I defer to @benjaminleonard on the details.

benjaminleonard:
`w-64` looks like a better compromise to me.

image


Added `overflow-ellipsis` seems neater than cutting it off completely.

benjaminleonard (approved):

PR #1918, opened 2024-01-29
Make `VpcNameFromId` more linkCell-like

benjaminleonard:
This was the only one I found like that. Eventually we might want a component that's a bit more flexible? The challenge here is that we're using this component inside both tables and the properties table and they appear a little differently.

### DIFF
--- a/app/pages/project/instances/instance/InstancePage.tsx
+++ b/app/pages/project/instances/instance/InstancePage.tsx
@@ -8,10 +8,11 @@
 import { format } from 'date-fns'
 import filesize from 'filesize'
 import { useMemo } from 'react'
-import { useNavigate, type LoaderFunctionArgs } from 'react-router-dom'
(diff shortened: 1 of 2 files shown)


app/pages/project/instances/instance/InstancePage.tsx
```diff
@@ -147,9 +156,16 @@ export function InstancePage() {
...
+            {vpc ? (
+              <Link
+                className="link-with-underline group text-sans-semi-md"
+                to={pb.vpc({ project: instanceSelector.project, vpc: vpc.name })}
+              >
+                {vpc.name}
+              </Link>
+            ) : (
+              <EmptyCell />
+            )}
```
david-crespo, line 168: I considered doing this in #1882 but was too lazy. Glad to have the push to do it the ugly right way.

PR #1926, opened 2024-01-31
Use NumberFields in more places; enhance NumberField component

charliepark:
We currently have a few places in the UI where we have TextInputs with `type="number"`. This renders like this:
Screenshot 2024-01-30 at 8 08 18 PM

We have a NumberField component that we're currently using in a few places, but could use in the other places where we currently have `type="number"` … this PR converts those over:
Screenshot 2024-01-30 at 8 01 30 PM

This PR also adds a small bit of logic to translate the `min` and `max` value that we pass in as props to NumberField, converting those to the `minValue` and `maxValue` props that the NumberInput component is expecting. The benefit from that is that we now automatically disable the up/down arrow button in the NumberField component when the input's value is at the boundary number. If the user tries typing in a number beyond the boundary, it gets pulled back to the min/max automatically.

### DIFF
--- a/app/components/form/fields/NumberField.tsx
+++ b/app/components/form/fields/NumberField.tsx
@@ -69,6 +69,8 @@ export const NumberFieldInner = <
   required,
   id: idProp,
   disabled,
+  max,
+  min,
 }: TextFieldProps<TFieldValues, TName>) => {
   const generatedId = useId()
   const id = idProp || generatedId
@@ -89,6 +91,8 @@ export const NumberFieldInner = <
               })}
               aria-describedby={tooltipText ? `${id}-label-tip` : undefined}
               isDisabled={disabled}
+              maxValue={max ? Number(max) : undefined}
+              minValue={min !== undefined ? Number(min) : undefined}
               {...field}
               formatOptions={{
                 useGrouping: false,
--- a/app/components/form/fields/NumberField.tsx
+++ b/app/components/form/fields/NumberField.tsx
@@ -70,7 +70,7 @@ export const NumberFieldInner = <
   id: idProp,
   disabled,
   max,
-  min,
+  min = 0,
 }: TextFieldProps<TFieldValues, TName>) => {
   const generatedId = useId()
   const id = idProp || generatedId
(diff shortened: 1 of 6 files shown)


app/components/form/fields/NumberField.tsx
```diff
@@ -89,6 +91,8 @@ export const NumberFieldInner = <
               })}
               aria-describedby={tooltipText ? `${id}-label-tip` : undefined}
               isDisabled={disabled}
+              maxValue={max ? Number(max) : undefined}
+              minValue={min !== undefined ? Number(min) : undefined}
```
charliepark, line 95: the `min !== undefined` construct is necessary because if `min` is `0`, `min ? __ : __` evaluates to false and `minValue` is set as `undefined`

charliepark:
Looks like I have a few tests to update.

benjaminleonard:
Does this overlap with #1812?

charliepark:
Weird; I could swear I'd added a note to link it to #1812 last night. There is some overlap. There were a few pieces of #1812 that I picked up in #1854, but skipped over the NumberField migration. This picks some of that up.

Justin had some adjustment to types and input manipulation that I didn't incorporate. I'll see if there are other pieces of that that make sense to pull over.

app/components/form/fields/TextField.tsx
```diff
@@ -30,10 +30,8 @@ import { ErrorMessage } from './ErrorMessage'
 export interface TextFieldProps<
   TFieldValues extends FieldValues,
   TName extends FieldPath<TFieldValues>,
-> extends UITextFieldProps {
+> extends Omit<UITextFieldProps, 'type'> {
   name: TName
-  /** HTML type attribute, defaults to text */
-  type?: string
```
charliepark, line 36: This was necessary when we needed to differentiate between the `text` and `number` uses of TextField. Now that numbers will be going through NumberField, this prop isn't necessary.

charliepark:
Ah, just noticed the broken CI check. :eyes:

charliepark:
Just to close the loop from before, yes, this takes care of the same tasks from #1812, and we can close that PR once this is merged.

app/components/form/fields/TextField.tsx
zephraph: Good catch!

zephraph (approved):
Fantastic! Thank you so much for taking care of this.

PR #1927, opened 2024-01-31
Abstract confirm-delete into more general confirm-action

david-crespo:
Needed to confirm IP pool link/unlink/make default in #1910. We have a really nice way of confirming deletes, but what if we want to confirm an action that's not a delete? Turns out there were only a couple of bits specific to deleting, and I was able to wrap them up in a helper that keeps the API the same.

Existing delete confirmations look and work exactly as they do now:

image

### DIFF
--- a/app/components/ConfirmActionModal.tsx
+++ b/app/components/ConfirmActionModal.tsx
@@ -8,16 +8,13 @@
 import { useState } from 'react'
 
 import { type ApiError } from '@oxide/api'
-import { Message, Modal } from '@oxide/ui'
-import { classed } from '@oxide/util'
+import { Modal } from '@oxide/ui'
 
-import { clearConfirmDelete, useConfirmDelete } from 'app/stores/confirm-delete'
+import { clearConfirmAction, useConfirmAction } from 'app/stores/confirm-action'
 import { addToast } from 'app/stores/toast'
 
-export const HL = classed.span`text-sans-semi-md text-default`
-
-export function ConfirmDeleteModal() {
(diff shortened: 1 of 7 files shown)


app/components/ConfirmActionModal.tsx
```diff
@@ -8,53 +8,45 @@
...
+  if (!actionConfig) return null
 
-  const displayLabel = typeof label === 'string' ? <HL>{label}</HL> : label
+  const { doAction, modalContent, errorTitle, modalTitle } = actionConfig
 
   return (
-    <Modal isOpen onDismiss={clearConfirmDelete} title="Confirm delete">
-      <Modal.Section>
-        <p>Are you sure you want to delete {displayLabel}?</p>
-        {warning && <Message variant="error" content={warning} />}
```
david-crespo, line 38: turns out this `warning` bit was not used by any existing call site

app/pages/project/instances/actions.tsx
```diff
@@ -111,12 +111,6 @@ export const useMakeInstanceActions = (
                   options.onDelete?.()
                   addToast({ title: `Deleting instance '${instance.name}'` })
                 },
-                onError: (error) =>
-                  addToast({
-                    variant: 'error',
-                    title: `Error deleting instance '${instance.name}'`,
-                    content: error.message,
-                  }),
```
david-crespo, line 119: This was a bug on `main`. `confirmDelete` shows an error toast, so this one is redundant.

PR #1928, opened 2024-01-31
Enter key in firewall rule subforms submits subform

### LINKED ISSUE #1919: Pressing enter in firewall rule "subforms" should not submit full form
I instinctively press enter when trying to add a target or host, expecting it to be equivalent to clicking the add button, but instead it submits the whole form, which is very frustrating. This is especially bad because it can be hard to tell what you did and didn't change.

image

david-crespo:
Closes #1919

It's actually kind of nice to consolidate the submits at the top so you can see the logic side by side.


zephraph:
I agree, I do like the submits organized at the top!

charliepark:
Chiming in a minute too late; works well locally!

PR #1929, opened 2024-01-31
Bump API for hostname validation change

david-crespo:
Technically not required as the only change is from `string` to `type Hostname = string`, but we might as well. Will update commit hash to one on main once https://github.com/oxidecomputer/omicron/pull/4938 is merged.


david-crespo:
Did this in 2aa720d297dd758e83a356a60af6396cacf45b7b

PR #1930, opened 2024-02-01
Add/remove IP ranges

### LINKED ISSUE #1937: Confirm action modal should have variants
I moved the implementation from confirm delete and it's all `danger` style.

image

david-crespo:
Followup to #1603 
Closes #1937 

The most interesting part is the IP range validation. I manually confirmed that Rust's `std::net` considers the same IPs valid and invalid as my regexes:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=babc49cd34bf19669137e22b9202d2eb

https://github.com/oxidecomputer/console/assets/3612203/df94e8fa-5201-4a58-b231-7ddf6463ae46


app/forms/ip-pool-range-add.tsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+  } else if ((first.isv4 && last.isv6) || (first.isv6 && last.isv4)) {
+    errors = { first: diffVersionError, last: diffVersionError }
+  }
+
+  // TODO: if we were really cool we could check first <= last but that sounds
+  // like a pain
+
+  return errors ? { values: {}, errors } : { values, errors: {} }
+}
```
david-crespo, line 63: It turns out writing a custom resolver to validate multiple fields together is very easy. We could do this more often. The only problem is that I think it can't be used with other kinds of validation. For example, I think the `required` prop below is being ignored because of this resolver. In this case, that's fine because empty string is not valid IP, so we catch it here.

PR #1933, opened 2024-02-01
Bump React Hook Form to latest

david-crespo:
There was an existing race condition in our instance create nav blocking code that was revealed by RHF 7.48.0. @charliepark and I are not sure what changed that caused it to be a problem, but it's a good opportunity to fix it regardless. See the RHF diff at [7.47.0...7.48.0](https://github.com/react-hook-form/react-hook-form/compare/v7.47.0...v7.48.0).

We're going to wait until after the release to merge this because we want to give it extra time to bake.


david-crespo:
I was curious how it handles server errors, as in, would `isSubmitSuccessful` stay false, or does it always turn true when the submit ends? Looks like it does the right thing, so it is looking at whether our async function resolves or rejects.

image

Issue #1932, opened 2024-02-01
Firewall filter alignment


augustuswm:
When applying multiple filters to a firewall (i.e. host + port) the alignment of the indicators looks a little off. The second line gets a slight indent and the vertical spacing is really tight.

![image](https://github.com/oxidecomputer/console/assets/1377730/7af35aae-89a7-4beb-b250-6ce2a7eb2d32)

PR #1934, opened 2024-02-02
Update layout and spacing of badges in firewall rules table

charliepark:
In #1932 we have a report of the badges having funky alignments:
image

There's a bit more we can do to improve these tables, but this PR at least addresses the layout. It pulls everything to the left-hand side of the column. It also slightly tightens the horizontal spacing between the badges (to 0.5), to create a bit more of a visual connection between them.
Screenshot 2024-02-01 at 5 02 18 PM

This PR also makes some changes to the mock api call, to have a bit more data showing on the firewall rules table by default.

I'll note that Justin and I were looking at this and think there could be improvements that (visually) more tightly link the filter type and the value, perhaps a split badge or something? Types with multiple filters (e.g. ports) would need consideration. All that's outside the scope of this PR, though, which is just intended to fix some of the spacing issues.

### DIFF
--- a/libs/api-mocks/vpc.ts
+++ b/libs/api-mocks/vpc.ts
@@ -55,12 +55,20 @@ export const defaultFirewallRules: Json<VpcFirewallRule[]> = [
     name: 'allow-internal-inbound',
     status: 'enabled',
     direction: 'inbound',
-    targets: [{ type: 'vpc', value: 'default' }],
+    targets: [
+      { type: 'vpc', value: 'default' },
+      { type: 'subnet', value: 'default' },
+    ],
     action: 'allow',
     description:
       'allow inbound traffic to all instances within the VPC if originated within the VPC',
     filters: {
-      hosts: [{ type: 'vpc', value: 'default' }],
+      hosts: [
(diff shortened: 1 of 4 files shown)


david-crespo:
Definitely an improvement. I wonder if we might want to let the row get taller. @benjaminleonard is professionally obligated to not let a border get this close to cell contents.

image

charliepark:
Can definitely look into row embiggening, though I seem to recall that playing with flexible heights led to some goofiness due to some non-normative table rendering we're doing.

benjaminleonard:
Instead of flexible row heights, this is probably the right opportunity to implement something like the overflowing tags idea we toyed with on the instances table. This will also be an important addition to the updated Access & IAM.

image

I wonder if the `useOverflow` hook might be used to tell us if a badge cell is overflowing, and at that point we can limit the rows and show how many extra rules have been truncated. A tooltip will then show the full list, or clicking on the rule itself.

benjaminleonard:
On linking the type and value, we had this on the designs (mix of dim and solid badge). Let's give it a go, see if it helps!
image

charliepark:
going to restart this effort in a new branch

PR #1935, opened 2024-02-02
Display API message on not found errors

david-crespo:
We have had multiple people raise the issue that when they tried to create an instance without a default IP pool in their silo, it was almost impossible to figure out what the problem was.

https://github.com/oxidecomputer/omicron/issues/4864
https://github.com/oxidecomputer/omicron/issues/4954

I improved the API error message in https://github.com/oxidecomputer/omicron/pull/4880, but this morning I figured out that in both cases, they were using the web console, and it did not matter what the API error was because we only show `OBJECT NOT FOUND` on instance create when we get a 404 back.

https://github.com/oxidecomputer/console/blob/2aa720d297dd758e83a356a60af6396cacf45b7b/libs/api/errors.ts#L58-L59

So this PR changes that behavior so that the client-side error object we pass around contains the error message from the API. As I mentioned [here](https://docs.google.com/document/d/1MfbRgxfYGRk2uQhGzZu3LpdAV5xO3fuKeqazBR14V_w/edit), it might be nice to change the error code from 404 to something else, but I really think this PR gets us 80% of the way there.

My only concern here is that there might be some spots where "Object not found" was nicer than the API error and made sense in context, but now we will be displaying the API error. But hopefully that can't be _too_ bad, since the API messages are all designed to be human readable and helpful.

### Before

image

### After

image

### DIFF
--- a/libs/api/errors.ts
+++ b/libs/api/errors.ts
@@ -56,7 +56,7 @@ export function processServerError(method: string, resp: ErrorResult): ApiError 
   if (code === 'Forbidden') {
     message = 'Action not authorized'
   } else if (code === 'ObjectNotFound') {
-    message = resp.data.message ? capitalize(resp.data.message) : 'Object not found'
+    message = capitalize(resp.data.message)
   } else if (code === 'ObjectAlreadyExists') {
     const resource = getResourceName(method, resp.data.message)
     if (resource) {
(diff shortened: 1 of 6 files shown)


libs/api/errors.ts
```diff
@@ -56,7 +56,7 @@ export function processServerError(method: string, resp: ErrorResult): ApiError
   if (code === 'Forbidden') {
     message = 'Action not authorized'
   } else if (code === 'ObjectNotFound') {
-    message = 'Object not found'
+    message = capitalize(resp.data.message)
```
david-crespo, line 59: This is the entire change. Everything else is tests.

PR #1949, opened 2024-02-09
Move ip validation

charliepark:
We'll want to access the IP validation in more places, including the Floating IP form. This PR moves the validator @david-crespo added recently to the string utils file for easier access.

### DIFF
--- a/libs/util/str.spec.ts
+++ b/libs/util/str.spec.ts
@@ -15,6 +15,7 @@ import {
   IPV6_REGEX,
   kebabCase,
   titleCase,
+  validateIp,
 } from './str'
 
 describe('capitalize', () => {
@@ -90,6 +91,7 @@ describe('titleCase', () => {
 
 test.each(['123.4.56.7', '1.2.3.4'])('ipv4Regex passes: %s', (s) => {
   expect(IPV4_REGEX.test(s)).toBe(true)
+  expect(validateIp(s)).toStrictEqual({ isv4: true, isv6: false, valid: true })
 })
 
 test.each([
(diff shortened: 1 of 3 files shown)


libs/util/str.spec.ts
```diff
@@ -154,3 +157,25 @@ test.each([
...
+  '::1:2:3:4:5:6:7:8',
+  '1:2:3:4:5:6:7:8::',
+  '1:2:3:4:5:6:7:88888',
+  '2001:db8:3:4:5::192.0.2.33', // std::new::Ipv6Net allows this one
+  'fe08::7:8%',
+  'fe08::7:8i',
+  'fe08::7:8interface',
+])('validateIp catches invalid IP: %s', (s) => {
+  expect(validateIp(s)).toStrictEqual({ isv4: false, isv6: false, valid: false })
+})
```
david-crespo, line 181: Since the tests for the regexes are so thorough, I think it would be ok to just do one or two for each of the possible outcomes here

```ts
test('validateIp', () => {
  expect(validateIp('').toEqual({ isv4: false, isv6: false, valid: false })
  expect(validateIp('1.1.1').toEqual({ isv4: false, isv6: false, valid: false })
  expect(validateIp('1.1.1.1').toEqual({ isv4: true, isv6: false, valid: true })
  // etc
})
```

david-crespo: Oh, I didn't realize you had added a line to each of the other tests. Sorry!

In that case, do what you did, love it, and don't even export the regexes if we're not using them directly anywhere. Just test this function.

PR #1950, opened 2024-02-12
Spring clean

benjaminleonard:
There are some other unused components, but I left the ones which we're likely to use in the future.


david-crespo (approved):
Beautiful

PR #1955, opened 2024-02-14
Fix `undefined` in capacity bar

### LINKED ISSUE #1954: `undefined` in capacity bar decimal
This is a v5 bug report. I thought this would be fixed with #1865, but v5 console is 02c6ce747fd5dd05e9d454ecb1bf70392c9d954e, precisely the merge commit of #1865. So most likely this is still possible.

image

david-crespo:
Closes #1954 

image


libs/util/math.ts
```diff
@@ -6,13 +6,9 @@
...
 
-export function splitDecimal(value: number) {
-  const wholeNumber = Math.trunc(value)
-  const decimal = value % 1 !== 0 ? round(value % 1, 2) : null
-  return [
-    wholeNumber.toLocaleString(),
-    decimal ? '.' + decimal.toLocaleString().split('.')[1] : '',
-  ]
+export function splitDecimal(value: number): [string, string] {
+  const [whole, decimal] = round(value, 2).toLocaleString().split('.')
```
david-crespo, line 10: this won't work if you're in a locale that doesn't use `.` as decimal separator

PR #1957, opened 2024-02-14
Create Floating IPs and attach them to instances

charliepark:
This PR adds the ability in the web console to create / attach / detach / delete Floating IPs. As the API doesn't currently have an update endpoint for Floating IPs, that's outside the scope of this PR.

Kapture 2024-02-21 at 16 54 08


app/forms/floating-ip-edit.tsx
```diff
@@ -0,0 +1,127 @@
...
+      id="edit-floating-ip-form"
+      form={form}
+      title="Floating IP"
+      onSubmit={(values) => {
+        updateAttachment.mutate({
+          path: { floatingIp: floatingIpName },
+          query: { project },
+          body: {
+            kind: 'instance',
+            parent: values.instanceId,
```
david-crespo, line 88: If you can be confident that at runtime you can't get here without this being defined (for example by enforcing that instanceis a required field) then the solution would be to use non-null assertion to shut TS up. Whenever I use one I like to include a comment about why. Example:

https://github.com/oxidecomputer/console/blob/b454cefd9b480bd95220b887b4c58d384c0dba94/app/components/form/fields/TlsCertsField.tsx#L69-L78

Another approach would be to use `invariant` to tell TS it's non-null, which I suppose has the advantage of blowing up at runtime if it's violated. I think we used to do that more and stopped.

The most strictly correct way of doing this, which we are not ready to get into, would be to use a `resolver` on the form, which would run the input `FormValues` type through a parse/transform step, in which case I _think_ `onSubmit` only gets called if it succeeds, and it gets the transformed type, so it would no longer worry `instanceId` can be undefined.

app/forms/floating-ip-edit.tsx
```diff
@@ -0,0 +1,127 @@
...
+    query: { project },
+  })
+  const { data: floatingIps } = usePrefetchedApiQuery('floatingIpList', {
+    query: { project },
+  })
+  console.log(instances, floatingIps)
+
+  const dismissLink = pb.floatingIps({ project })
+  const navigate = useNavigate()
+  const form = useForm({ defaultValues: { instanceId: floatingIp.instanceId } })
```
david-crespo, line 63: Changing this to `instanceId: floatingIp.instanceId || null` will fix the type error on the `selected` prop below by making this `string | null` instead of `string | undefined`.

In other forms we tend to have an explicit type for the form values. I think that makes it easier to think about the form. Often we can use the `*Create` type from the API directly or modify it slightly, but here we can't, so I would just make it from scratch:

```ts
type FormValues = {
  instanceId: string | null
}
```

And then on this line you can do something like

```ts
const form = useForm({ 
  defaultValues: { instanceId: floatingIp.instanceId || null } satisfies FormValues
})
```

libs/api-mocks/msw/handlers.ts
```diff
@@ -223,6 +224,61 @@ export const handlers = makeHandlers({
...
     return 204
   },
+  floatingIpCreate({ body, query }) {
+    const project = lookup.project(query)
+    errIfExists(db.vpcs, { name: body.name })
+
+    const newFloatingIp: Json<Api.FloatingIp> = {
+      id: uuid(),
+      project_id: project.id,
+      ip: body.address || '', // 👀 needs a legit ip
```
david-crespo, line 234: This is necessarily going to be fakey without a robust mock IP situation, which I don't think is necessary yet. I think you should just stick in a dummy IP like `12.34.56.7`. We'll get duplicates if you create more than one, but I think that's ok for now.

app/forms/floating-ip-create.tsx
```diff
@@ -0,0 +1,158 @@
...
+      <DescriptionField name="description" control={form.control} />
+      {allPools && (
+        <ListboxField
+          name="pool"
+          items={allPools?.items.map((p) => toListboxItem(p))}
+          label="Pool"
+          required
+          control={form.control}
+        />
+      )}
```
david-crespo, line 148: This field not be required. Also by using `usePrefetchedApiQuery` above you can get rid of the `allPools` check because you know it's defined.

libs/api-mocks/floating-ip.ts
```diff
@@ -0,0 +1,38 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+
+import type { FloatingIp } from '@oxide/api'
+
+import { instance, project } from '.'
+import type { Json } from './json-type'
```
david-crespo, line 12: This will fix the confusing test failure.

```suggestion
import { instance } from './instance'
import type { Json } from './json-type'
import { project } from './project'
```

Vite does not like the import from `'.'`. I suspect it is due to the circular import: `./index.ts` imports this file, and we're importing from it. There are very few spots we import `from '.'`, and they should probably be changed. I'll do that separately and see if there's a lint rule we can turn on.

```
$ rg "from '.'"
libs/api/roles.ts
18:import { usePrefetchedApiQuery } from '.'

libs/util/math.spec.ts
10:import { GiB } from '.'

libs/util/math.ts
9:import { splitOnceBy } from '.'

libs/table/cells/BooleanCell.tsx
14:import type { Cell } from '.'

libs/table/cells/EnabledCell.tsx
11:import type { Cell } from '.'

libs/table/cells/TypeValueListCell.tsx
8:import { TypeValueCell, type Cell, type TypeValue } from '.'

libs/table/cells/FirewallFilterCell.tsx
11:import { TypeValueCell, type Cell } from '.'
```

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+        </Link>
+      </TableActions>
+      <Table emptyState={<EmptyState />} makeActions={makeActions}>
+        <Column accessor="name" />
+        <Column accessor="description" />
+        <Column accessor="ip" />
+        <Column
+          accessor="instanceId"
+          header="Attached to instance"
+          cell={({ value: instanceId }) => getInstanceName(instanceId)}
```
david-crespo, line 139: This should be a linky cell that goes to the instance.

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+        },
+      },
+      {
+        label: 'Detach',
+        disabled: isAttachedToAnInstance
+          ? false
+          : 'This floating IP is not attached to an instance',
+        onActivate() {
+          setFloatingIpToModify(floatingIp)
+          setDetachModalOpen(true)
```
david-crespo, line 98: This whole confirm process can be done a bit more automatically. See IP pool unlink:

https://github.com/oxidecomputer/console/blob/d4ec192725e7410c2f5b4d653fc223eb04f0569b/app/pages/system/networking/IpPoolPage.tsx#L118-L136

app/routes.tsx
```diff
@@ -350,6 +352,15 @@ export const routes = createRoutesFromElements(
           />
         </Route>
 
+        <Route loader={FloatingIpsPage.loader} element={<FloatingIpsPage />}>
+          <Route path="floating-ips" handle={{ crumb: 'Floating IPs' }} element={null} />
+          <Route
+            path="floating-ips-new"
+            element={<CreateFloatingIpSideModalForm />}
```
david-crespo, line 359: I think this is missing this, which is why the pools listbox pops in when it should be prefetched.

```
loader={CreateFloatingIpSideModalForm.loader}
```

app/forms/floating-ip-create.tsx
```diff
@@ -0,0 +1,157 @@
...
+  onSubmit,
+  onSuccess,
+}: CreateSideModalFormProps) {
+  // Fetch 1000 to we can be sure to get them all. There should only be a few
+  // anyway. Not prefetched because the prefetched one only gets 25 to match the
+  // query table. This req is better to do async because they can't click make
+  // default that fast anyway.
+  const { data: allPools } = useApiQuery('projectIpPoolList', {
+    query: { limit: 1000 },
+  })
```
david-crespo, line 69: Use `usePrefetchedApiQuery` here and ditch the comment except the first sentence.

app/forms/floating-ip-create.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const defaultPool = useMemo(
+    () => allPools?.items.find((p) => p.isDefault)?.name,
+    [allPools]
+  )
+
+  const defaultValues: FloatingIpCreate = {
+    name: '',
+    description: '',
+    // defaultPool doesn't seem to be getting set in the form when page is loaded directly
```
david-crespo, line 79: Fixing the prefetch on the pools will probably fix this. The form only pulls from these `defaultValues` on initial render, and if the pools are not fetched yet, then `defaultPool` is always going to be undefined on that first render.

app/forms/floating-ip-create.tsx
```diff
@@ -0,0 +1,157 @@
...
+   * the RQ `onSuccess` defined for the mutation.
+   */
+  onSubmit?: (floatingIpCreate: FloatingIpCreate) => void
+  /**
+   * Passing navigate is a bit of a hack to be able to do a nav from the routes
+   * file. The callers that don't need the arg can ignore it.
+   */
+  onDismiss?: (navigate: NavigateFunction) => void
+  onSuccess?: (floatingIp: FloatingIp) => void
+}
```
david-crespo, line 57: I'd get rid of these props. Some other forms have them because they get loaded in multiple contexts, like how you can create a disk (sort of virtually) from the instance create form in addition to the disks list. Forms that get called in a single context don't need these props because they can hard code their behavior. Some examples:

https://github.com/oxidecomputer/console/blob/d4ec192725e7410c2f5b4d653fc223eb04f0569b/app/forms/ip-pool-create.tsx#L22

https://github.com/oxidecomputer/console/blob/d4ec192725e7410c2f5b4d653fc223eb04f0569b/app/forms/silo-create.tsx#L51

https://github.com/oxidecomputer/console/blob/d4ec192725e7410c2f5b4d653fc223eb04f0569b/app/forms/ip-pool-create.tsx#L22

david-crespo:
You have to `rm -rf app/pages/projects/networking`. You moved its contents up a level but must have forgotten to commit the delete half, so you have all those files added here rather than moved.

app/util/path-builder.ts
```diff
@@ -66,6 +66,8 @@ export const pb = {
   vpcs: (params: Project) => `${pb.project(params)}/vpcs`,
   vpc: (params: Vpc) => `${pb.vpcs(params)}/${params.vpc}`,
   vpcEdit: (params: Vpc) => `${pb.vpc(params)}/edit`,
+  floatingIps: (params: Project) => `${pb.project(params)}/floating-ips`,
+  floatingIpsNew: (params: Project) => `${pb.project(params)}/floating-ips-new`,
```
david-crespo, line 70: For whatever reason (I am probably to blame) the convention for this name is `floatingIpNew`, even though we always put the `s` in the path. We should probably rename all of these to match the paths (separate PR obviously).

libs/api-mocks/msw/handlers.ts
```diff
@@ -223,6 +223,53 @@ export const handlers = makeHandlers({
...
+    const ips = db.floatingIps.filter((i) => i.project_id === project.id)
+    return paginated(query, ips)
+  },
+  floatingIpView({ query, path }) {
+    const project = lookup.project(query)
+    const ip = db.floatingIps.filter(
+      (i) => i.project_id === project.id && i.name === path.floatingIp
+    )[0]
+    return ip
+  },
```
david-crespo, line 251:
```suggestion
  floatingIp: ({ path, query }) => lookup.floatingIp({ ...path, ...query }),
```

libs/api-mocks/msw/handlers.ts
```diff
@@ -223,6 +223,53 @@ export const handlers = makeHandlers({
...
+
+    return 204
+  },
+  floatingIpAttach({ path, query, body }) {
+    const floatingIp = lookup.floatingIp({ ...path, ...query })
+    const instance = lookup.instance({ ...path, ...query, instance: body.parent })
+    floatingIp.instance_id = instance.id
+
+    return floatingIp
+  },
```
david-crespo, line 264: This should probably 400 if the IP already has an `instance_id`. It looks like that is the API behavior:

https://github.com/oxidecomputer/omicron/blob/1f48138840644e2573800b8f52dd00af35acc179/nexus/tests/integration_tests/external_ips.rs#L687-L704

I think the easy way to get that 400 is literally

```ts
if (floatingIp.instance_id) {
  throw "floating IP cannot be attached to one instance while still attached to another"
}
```

That will definitely return an error, and I'm like 95% sure it will automatically get a 400 status.

libs/api-mocks/msw/handlers.ts
```diff
@@ -223,6 +223,53 @@ export const handlers = makeHandlers({
...
+  floatingIpDetach({ path, query }) {
+    const floatingIp = lookup.floatingIp({ ...path, ...query })
+    db.floatingIps = db.floatingIps.map((ip) =>
+      ip.id !== floatingIp.id ? ip : { ...ip, instance_id: undefined }
+    )
+
+    return floatingIp
+  },
```
david-crespo, line 272: I thought maybe this should error if the IP isn't attached to an instance, but it looks like the API does not:

https://github.com/oxidecomputer/omicron/blob/1f48138840644e2573800b8f52dd00af35acc179/nexus/tests/integration_tests/external_ips.rs#L632-L636

libs/api-mocks/msw/handlers.ts
```diff
@@ -223,6 +223,53 @@ export const handlers = makeHandlers({
...
     return 204
   },
+  floatingIpCreate({ body, query }) {
+    const project = lookup.project(query)
+    errIfExists(db.floatingIps, { name: body.name })
+
+    const newFloatingIp: Json<Api.FloatingIp> = {
+      id: uuid(),
+      project_id: project.id,
+      ip: `${[...Array(4)].map(() => Math.floor(Math.random() * 256)).join('.')}`,
```
david-crespo, line 233: Nice. Because of the `join` it's already a string, so you don't need the interpolation.

```suggestion
      ip: [...Array(4)].map(() => Math.floor(Math.random() * 256)).join('.'),
```

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+FloatingIpsPage.loader = async ({ params }: LoaderFunctionArgs) => {
+  const { project } = getProjectSelector(params)
+  await Promise.all([
+    apiQueryClient.prefetchQuery('floatingIpList', {
+      query: { project, limit: 25 },
+    }),
+    apiQueryClient.prefetchQuery('instanceList', {
+      query: { project },
+    }),
```
david-crespo, line 56: I love the list trick, and I want to make followup PRs doing this in other places where we're currently fetching names one by one using the resource ID. I'm not sure the default page size of 100 is enough, though. I think we either want to go big here (like 1000) _or_ get elaborate and fall back to fetching the name by instance ID if it's somehow not in the 100 we fetched. I don't think there will regularly be 1000 instances in projects any time soon, so I am leaning toward keeping it simple and just fetching a lot.

app/forms/floating-ip-create.tsx
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+    // For the default pool, add a label to the dropdown
+    return {
+      value: p.name,
+      labelString: p.name,
+      label: (
+        <>
+          {p.name}{' '}
+          <span className="text-quaternary selected:text-accent-secondary">(default)</span>
```
david-crespo, line 86: Try using a `<Badge>default</Badge>` for this.

app/forms/floating-ip-create.tsx
```diff
@@ -0,0 +1,124 @@
...
+    onSuccess() {
+      queryClient.invalidateQueries('floatingIpList')
+      addToast({ content: 'Your Floating IP has been created' })
+      navigate(pb.floatingIps(projectSelector))
+    },
+  })
+
+  const form = useForm({ defaultValues })
+
+  const toListboxItem = (p: SiloIpPool) => {
```
david-crespo, line 72: If you add `defaultPool` as an argument, this function could be static, defined at top level outside of the component.

app/forms/floating-ip-create.tsx
```diff
@@ -0,0 +1,182 @@
...
+      <Accordion.Content
+        ref={contentRef}
+        forceMount
+        className={cn('ox-accordion-content overflow-hidden py-4', { hidden: !isOpen })}
+      >
+        {children}
+      </Accordion.Content>
+    </Accordion.Item>
+  )
+}
```
david-crespo, line 182: This could be simplified a bit if we decide we don't care about the `scrollIntoView` business. It's a lot more important on instance create because those accordions are at the bottom of a long form. In this case it's less like that the user will have to scroll. On the other hand, it is pretty nifty. My inclination now after seeing the code is to shared `AccordionItem` only. Trying to share a full Accordion component with the `openItems` stuff wired up would force us into a rather horrible props API.

https://github.com/oxidecomputer/console/assets/3612203/c51ddc0e-752b-47eb-a0e7-772d5cd5d04f

For completeness, here would be the diff for the simplified version. It works minus the scrolling — this is more the default way of using the Radix accordion component. But like I said I think it's better to keep the fanciness and share `AccordionItem`. You could move it into `app/components`.

```diff
diff --git a/app/forms/floating-ip-create.tsx b/app/forms/floating-ip-create.tsx
index d259820b..ff565b9e 100644
--- a/app/forms/floating-ip-create.tsx
+++ b/app/forms/floating-ip-create.tsx
@@ -87,8 +87,6 @@ export function CreateFloatingIpSideModalForm() {
   const form = useForm({ defaultValues })
   const isPoolSelected = !!form.watch('pool')
 
-  const [openItems, setOpenItems] = useState<string[]>([])
-
   return (
     <SideModalForm
       id="create-floating-ip-form"
@@ -108,17 +106,8 @@ export function CreateFloatingIpSideModalForm() {
       <NameField name="name" control={form.control} />
       <DescriptionField name="description" control={form.control} />
 
-      <Accordion.Root
-        type="multiple"
-        className="mt-12 max-w-lg"
-        value={openItems}
-        onValueChange={setOpenItems}
-      >
-        <AccordionItem
-          value="advanced"
-          label="Advanced"
-          isOpen={openItems.includes('advanced')}
-        >
+      <Accordion.Root type="single" className="mt-12 max-w-lg" collapsible>
+        <AccordionItem value="advanced" label="Advanced">
           <Message
             variant="info"
             content="If you don’t specify a pool, the default will be used."
@@ -148,20 +137,11 @@ export function CreateFloatingIpSideModalForm() {
 
 type AccordionItemProps = {
   value: string
-  isOpen: boolean
   label: string
   children: React.ReactNode
 }
 
-function AccordionItem({ value, label, children, isOpen }: AccordionItemProps) {
-  const contentRef = useRef<HTMLDivElement>(null)
-
-  useEffect(() => {
-    if (isOpen && contentRef.current) {
-      contentRef.current.scrollIntoView({ behavior: 'smooth' })
-    }
-  }, [isOpen])
-
+function AccordionItem({ value, label, children }: AccordionItemProps) {
   return (
     <Accordion.Item value={value}>
       <Accordion.Header className="max-w-lg">
@@ -170,11 +150,7 @@ function AccordionItem({ value, label, children, isOpen }: AccordionItemProps) {
           <DirectionRightIcon className="transition-all text-secondary" />
         </Accordion.Trigger>
       </Accordion.Header>
-      <Accordion.Content
-        ref={contentRef}
-        forceMount
-        className={cn('ox-accordion-content overflow-hidden py-4', { hidden: !isOpen })}
-      >
+      <Accordion.Content className="ox-accordion-content overflow-hidden py-4">
         {children}
       </Accordion.Content>
     </Accordion.Item>
```

app/test/e2e/floating-ip-create.e2e.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Copyright Oxide Computer Company
+ */
+
+import { clickRowAction, expect, expectNotVisible, expectVisible, test } from './utils'
+
+const floatingIpsPage = '/projects/mock-project/floating-ips'
+
+test('can create a Floating IP', async ({ page }) => {
+  await page.goto(floatingIpsPage)
+  await page.locator('text="New Floating IP"').click()
```
david-crespo, line 15: I try to use `getByRole` for things like this as it lets us exercise the [accessible name](https://developer.mozilla.org/en-US/docs/Glossary/Accessible_name) and ARIA role of the element.

```suggestion
  await page.getByRole('link', { name: 'New Floating IP' }).click()
```

app/test/e2e/floating-ip-create.e2e.ts
```diff
@@ -0,0 +1,71 @@
...
+  await expectVisible(page, ['role=textbox[name="Address"]'])
+
+  await page.getByRole('button', { name: 'Create Floating IP' }).click()
+
+  await expect(page).toHaveURL(floatingIpsPage)
+
+  await expectVisible(page, [
+    `text=${floatingIpName}`,
+    'text=A description for this Floating IP',
+  ])
```
david-crespo, line 45: You can use this util here and in a few spots in the other test.

```suggestion
  await expectRowVisible(page.getByRole('table'), {
    name: floatingIpName,
    description: 'A description for this Floating IP',
  })
```

app/test/e2e/floating-ip-create.e2e.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  // Click the down arrow and select top option
+  await page.keyboard.press('ArrowDown')
+  await page.keyboard.press('Enter')
+  await page.getByRole('button', { name: 'Attach' }).click()
+
+  // The dialog should be gone
+  await expectNotVisible(page, ['role=heading[name*="Attach Floating IP"]'])
+  await expectVisible(page, ['text=db1'])
+})
```
david-crespo, line 71: I'd add one more test that puts a bad IP address in the IP field and looks for the validation error. Compare:

https://github.com/oxidecomputer/console/blob/5d989a703359b6fa6723f59a87ba22224e62da34/app/test/e2e/ip-pools.e2e.ts#L126-L178

app/forms/floating-ip-create.tsx
david-crespo: We talked about this on the phone because my above stream of consciousness was not clear — the short version would have been: on second thought, let's share `AccordionItem` with instance create.

david-crespo:
This is looking great. A few things remaining for me. Some could be done in followups so we can get this merged and work more incrementally:

* https://github.com/oxidecomputer/console/issues/1966
* https://github.com/oxidecomputer/console/issues/1965 (more of a separate thing, not really fixing anything here, but worth mentioning)
* I personally prefer having only `detach` or `attach` in a given set of actions at a time. I think it's clear enough what they means (and we explain, or should explain, the action in the modal that pops up) and it saves us from having to write out this disabled copy. I think only having one or the other helps the user see them (correctly) as a sort of toggle on and toggle off rather than two parallel actions.

  image


* Info box in the attach modal explaining what this action does 

  image

* @benjaminleonard and @paryhin taking a look at what we did with the advanced accordion on floating IP create. Copy could also use a second look

  image

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -0,0 +1,242 @@
...
+              <p>
+                Are you sure you want to detach the floating IP <HL>{floatingIp.name}</HL>{' '}
+                from the instance{' '}
+                <HL>
+                  {
+                    // instanceId is guaranteed to be non-null here
+                    getInstanceName(floatingIp.instanceId!)
+                  }
+                </HL>
+                ?
```
david-crespo, line 127: I would add another line here like `The instance will no longer be reachable at <HL>{floatingIp.ip}</HL>`." And it's a nitpick, but I think both "the"s are unnecessary. `Are you sure you want to detach floating IP cola-float from instance db1` reads ok to me.

image

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -0,0 +1,249 @@
...
+        />
+      </Table>
+      <Outlet />
+      {attachModalOpen && floatingIpToModify && (
+        <AttachFloatingIpModal
+          floatingIp={floatingIpToModify.name}
+          address={floatingIpToModify.ip}
+          instances={instances.items}
+          project={project}
+          onDismiss={() => setAttachModalOpen(false)}
```
david-crespo, line 175: The whole `attachModalOpen` state variable is unnecessary as you can use whether `floatingIpToModify` is present to mean the same thing. So here you could just do `setFloatingIpToModify(null)` instead, and get rid of all the other references to `attachModalOpen`. Here's another spot where we do this:

https://github.com/oxidecomputer/console/blob/5d989a703359b6fa6723f59a87ba22224e62da34/app/pages/project/networking/VpcPage/tabs/VpcSubnetsTab.tsx#L25

david-crespo (approved):
One more small thing but it's ready to go!

PR #1960, opened 2024-02-15
Prevent `import` `from '.'`

david-crespo:
@charliepark ran into a very confusing circular import issue in #1957 due to importing from `'.'`, which made me realize we should just disallow that entirely. The rule worked exactly as I hoped. Here's the output before I fixed the imports:

Screenshot 2024-02-14 at 9 03 00 PM

### DIFF
--- a/libs/api/index.ts
+++ b/libs/api/index.ts
@@ -5,66 +5,11 @@
  *
  * Copyright Oxide Computer Company
  */
-import { QueryClient } from '@tanstack/react-query'
 
 // for convenience so we can do `import type { ApiTypes } from '@oxide/api'`
 import type * as ApiTypes from './__generated__/Api'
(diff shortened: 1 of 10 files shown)


libs/api/index.ts
```diff
@@ -5,66 +5,11 @@
...
-    },
-  },
-})
-
-// to be used in loaders, which are outside the component tree and therefore
-// don't have access to context
-export const apiQueryClient = wrapQueryClient(api.methods, queryClient)
-
-// to be used to retrieve the typed query client in components
-export const useApiQueryClient = getUseApiQueryClient(api.methods)
```
david-crespo, line 66: had to move this all to its own file to do `import { usePrefetchedApiQuery } from './client'` in `roles.ts`

PR #1968, opened 2024-02-23
Return floating IPs from mock instance external IPs endpoint

### LINKED ISSUE #1966: Floating IPs should be returned by mock `instanceExternalIpList`
Floating IPs don't come back in the mock response for `instanceExternalIpList` and are therefore missing on instance detail.

image

This is because floating IPs are stored in `db.floatingIps`, while ephemeral IPs are stored in `db.externalIps`, and the mock endpoint only pulls from the latter. There are a few ways we could fix this — we could either get rid of `db.floatingIps` and store floating IPs in `db.externalIps` instead, or we could have the endpoint combine results from both tables. I think I prefer the former because `ExternalIp` is designed to represent either a floating IP or an ephemeral IP, and this is more or less what it's like in the real DB, which has a single table called [`external_ip`](https://github.com/oxidecomputer/omicron/blob/dea0ea5a1e26464dc9780ae908f028d210f5f637/schema/crdb/dbinit.sql#L1704-L1708), though there is a view on that table called [`floating_ip`](https://github.com/oxidecomputer/omicron/blob/dea0ea5a1e26464dc9780ae908f028d210f5f637/schema/crdb/dbinit.sql#L1852-L1870).

david-crespo:
Closes #1966

I considered storing floating and ephemeral IPs in a single `externalIps` table, but it was a bit annoying because the response for the floating IPs list is slightly different from the `ExternalIp` type — it doesn't have a `kind` field because there is only one kind. For some reason it felt better to add the kind in the `instanceExternalIpList` handler than to remove the kind for the `floatingIp` endpoint and friends. We can revisit this in the future if more complex mock functionality makes the two tables feel overcomplicated. Right now nothing ever writes to the ephemeral IPs table, though it probably should — when you create an instance it should add an ephemeral IP.

### DIFF
--- a/app/test/e2e/floating-ip-create.e2e.ts
+++ b/app/test/e2e/floating-ip-create.e2e.ts
@@ -68,19 +68,34 @@ test('can create a Floating IP', async ({ page }) => {
 })
 
 test('can detach and attach a Floating IP', async ({ page }) => {
+  // check floating IP is visible on instance detail
+  await page.goto('/projects/mock-project/instances/db1')
+  await expect(page.getByText('192.168.64.64')).toBeVisible()
+
+  // now go detach it
   await page.goto(floatingIpsPage)
 
   await expectRowVisible(page.getByRole('table'), {
+    name: 'cola-float',
+    ip: '192.168.64.64',
     'Attached to instance': 'db1',
   })
   await clickRowAction(page, 'cola-float', 'Detach')
   await page.getByRole('button', { name: 'Confirm' }).click()
 
-  await expectNotVisible(page, ['role=heading[name*="Detach Floating IP"]'])
+  await expect(page.getByRole('dialog')).toBeHidden()
   // Since we detached it, we don't expect to see db1 any longer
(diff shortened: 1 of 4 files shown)


app/test/e2e/floating-ip-create.e2e.ts
```diff
@@ -68,28 +68,45 @@ test('can create a Floating IP', async ({ page }) => {
...
   // Since we detached it, we don't expect to see db1 any longer
-  await expectNotVisible(page, ['text=db1'])
+  await expect(page.getByRole('cell', { name: 'db1' })).toBeHidden()
+
+  // click back over to instance page (can't use goto because it refreshes) and
+  // confirm the IP is no longer there either
+  await page.getByRole('link', { name: 'Instances' }).click()
+  await page.getByRole('link', { name: 'db1' }).click()
+  await expect(page.getByRole('heading', { name: 'db1' })).toBeVisible()
+  await expect(page.getByText('192.168.64.64')).toBeHidden()
```
david-crespo, line 95: This is all a bit silly but why not

PR #1970, opened 2024-02-23
Shuffle @oxide/pagination files 

david-crespo:
Initially I was trying to use this as a simple case for changing the aliases, but it's so simple we don't even need multiple files. So I kept the alias pointing at the index file and just put everything in that file.


david-crespo:
I ended up reverting this in 3ec255f9fba44236736440e5df5088406e1d9977 to make the diff on #1971 trivial.

Issue #1964, opened 2024-02-21
Avoid barrel files 


david-crespo:
Barrel files are bad for Vite dev server performance (and probably bad for code splitting too if we were doing more of it — and we intend to). We should get rid of them, starting with the biggest ones. It should be a noisy but very mechanical change. I think this also means getting rid of `@oxide/ui` and friends in favor of normal direct imports from, e.g., `~/libs/ui/...` or whatever.

https://vitejs.dev/guide/performance.html#avoid-barrel-files

> ### Avoid Barrel Files
> 
> Barrel files are files that re-export the APIs of other files in the same directory. For example:
> js
> 
> ```js
> // src/utils/index.js
> export * from './color.js'
> export * from './dom.js'
> export * from './slash.js'
> ```
> 
> When you only import an individual API, e.g. import { slash } from './utils', all the files in that barrel file need to be fetched and transformed as they may contain the slash API and may also contain side-effects that run on initialization. This means you're loading more files than required on the initial page load, resulting in a slower page load.
> 
> If possible, you should avoid barrel files and import the individual APIs directly, e.g. import { slash } from './utils/slash.js'. You can read [issue #8237](https://github.com/vitejs/vite/issues/8237) for more information.

---

We have numerous examples, some worse than others.

https://github.com/oxidecomputer/console/blob/5d989a703359b6fa6723f59a87ba22224e62da34/app/pages/project/index.tsx#L9-L14

I think this means that from Vite's point of view, any file that imports any UI component depends on _every_ UI component.

https://github.com/oxidecomputer/console/blob/5d989a703359b6fa6723f59a87ba22224e62da34/libs/ui/index.ts#L15-L55

david-crespo:
There are a couple more (`app/hooks` is the one that matters) but we're close enough and are seeing (I think) substantially faster HMR as a result of only having to reload the one changed file instead of nearly all of them. Initial page load doesn't seem faster.

PR #1971, opened 2024-02-23
Move `libs/table` and `libs/pagination` into `app/`

david-crespo:
Changes here:

* Move `libs/table` to `app/table` and update `@oxide/table` alias for that
* Move contents of `libs/pagination` to `app/components` and `app/hooks` (there was a component and a hook), get rid of `@oxide/pagination` alias (it was only used twice)

---

We're reworking aliases as part of #1964 and I'm thinking the `app` vs. `libs` distinction doesn't really make sense and we should get rid of it. I picked a couple of smaller pieces here to show what that looks like. It's very simple because of the `@oxide/table` alias: we don't have to change any imports, all we have to do is change what the alias points to. With pagination there were only two imports so I converted those to the `~/pagination` style already.

After moving all the libs into `app`, we can scrap the aliases altogether in favor of having `~/*` simply point to `app/*`, and all the imports become `~/table/cells/DateCell` or whatever.


libs/pagination/Pagination.tsx
```diff
@@ -1,57 +0,0 @@
-/*
```
david-crespo, line 1: accidentally left this duplicate file in in #1970

.prettierrc.js
```diff
@@ -18,6 +18,7 @@ export default {
     '',
     '^@oxide/(.*)$',
     '',
+    '^~/(.*)$',
     '^app/(.*)$',
```
david-crespo, line 22: This is about sorting `~/` into the right spot when prettier sorts and groups imports.

PR #1981, opened 2024-02-26
Get rid of `@oxide/table` alias and barrel file

david-crespo:
Also got rid of my ill-conceived `useReactTable` wrapper.


app/table/react-table.ts
```diff
@@ -1,31 +0,0 @@
...
-  type TableOptions,
-} from '@tanstack/react-table'
-
-/**
- * Exactly the same as the default `useReactTable` except it handles
- * `getCoreRowModel` for you.
- */
-export const useReactTable = <TData extends RowData>(
-  options: Omit<TableOptions<TData>, 'getCoreRowModel'>
-) => useReactTableOrig({ ...options, getCoreRowModel: getCoreRowModel() })
```
david-crespo, line 23: This was my idea, and I have concluded it was more confusing than helpful. The only reason any of this existed was to avoid writing `getCoreRowModel: getCoreRowModel()` in the calling code.

PR #1982, opened 2024-02-26
Update OMICRON_VERSION and regenerate api

charliepark:
Updated OMICRON_VERSION and ran `npm run gen-api`.

Also updated `oxide.ts`, in https://github.com/oxidecomputer/oxide.ts/pull/229


app/forms/floating-ip-create.tsx
```diff
@@ -123,7 +123,8 @@ export function CreateFloatingIpSideModalForm() {
             placeholder="Select pool"
           />
           <TextField
-            name="address"
+            name="ip"
+            label="Address"
```
david-crespo, line 127: This hasn’t gone out to any real users, so we should just name it whatever is best. Even if it had I think I’d say the same thing. That may still be “Address”, but consistency wouldn’t be the reason.

charliepark: I think `IP address` works well and makes it clear (and maybe more clear than just `IP`); will switch to that, but happy to update again if we want it to map precisely to the API.

david-crespo (approved):

PR #2011, opened 2024-02-29
Fix rounding bug on `n >= 1000` 

### LINKED ISSUE #2010: Memory utilization shows up as NaN in some cases
This was seen on rack2 for the "oxide" silo. The current quota and utilization from the API show:
```
oxide api /v1/utilization
{
  "capacity": {
    "cpus": 960,
    "memory": 8367025789338,
    "storage": 38654705664000
  },
  "provisioned": {
    "cpus": 771,
    "memory": 2177548419072,
    "storage": 17444009672704
  }
}
```
The UI however shows both numbers as `NaN` GiB:
image

It seems to be a problem with the unit of measure conversion from bytes to GiB.

david-crespo:
Closes #2010 

I'm suffering.

### DIFF
--- a/app/util/math.ts
+++ b/app/util/math.ts
@@ -29,6 +29,10 @@ export function splitDecimal(value: number): [string, string] {
 }
 
 export function round(num: number, digits: number) {
-  const nf = Intl.NumberFormat(navigator.language, { maximumFractionDigits: digits })
+  const nf = Intl.NumberFormat(navigator.language, {
+    maximumFractionDigits: digits,
+    // very important, otherwise turning back into number will fail on > 1000 due to commas
+    useGrouping: false,
+  })
   return Number(nf.format(num))
 }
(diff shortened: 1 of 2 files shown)


app/util/math.ts
```diff
@@ -29,6 +29,10 @@ export function splitDecimal(value: number): [string, string] {
 }
 
 export function round(num: number, digits: number) {
-  const nf = Intl.NumberFormat(navigator.language, { maximumFractionDigits: digits })
+  const nf = Intl.NumberFormat(navigator.language, {
+    maximumFractionDigits: digits,
+    // very important, otherwise turning back into number will fail on > 1000 due to commas
+    useGrouping: false,
```
benjaminleonard, line 35: Does this need to be returned as a number? I say because there's a few cases where having the grouping is preferred. `1,234 KiB` is easier to read than `1234 KiB`.

The one place we're using it could be a string: https://github.com/oxidecomputer/console/blob/25fb506d0527f5508dc90b118672c0bf8ce14dd3/app/pages/system/UtilizationPage.tsx#L250

david-crespo: That is a very good point. Rounding is a display thing.

PR #2016, opened 2024-02-29
Remove Badge component from barrel file

charliepark:
Part of #1964 

Wanted to start with a simple migration from the existing Badge component imports over to importing directly from the component file.


david-crespo (approved):
good one!

PR #2022, opened 2024-03-02
More barrel file component extraction

charliepark:
More #1964 work.

Throwing down barrels like Donkey Kong

donkey_kong


david-crespo (approved):
🚂

PR #2026, opened 2024-03-04
Optimize rack image

david-crespo:
* [main](https://console-pi7pkf6lj-oxidecomputer.vercel.app/login/oxide/saml/google)
* [first attempt](https://console-prlix0en8-oxidecomputer.vercel.app/login/oxide/saml/google) (2/3 reduction)
* [second attempt](https://console-2bcf6ywfz-oxidecomputer.vercel.app/login/oxide/saml/google) (1/3 reduction, looks better)
* [Ben's final version](https://console-3jxb48vnv-oxidecomputer.vercel.app/login/oxide/saml/google) (half size and _better_ than the original! yes!)


david-crespo:
Ok, I've found the catch. I'm sure the size can be cut substantially without this downside.

image

david-crespo:
I notice what I think are compression artifacts even on the `main` copy — I wonder if we'd see better results if we started from an original uncompressed version.

image

Issue #2021, opened 2024-03-01
Edit floating IP name and description


david-crespo:
Now that we have an update endpoint for floating IPs (https://github.com/oxidecomputer/omicron/pull/5176), we might as well have an edit form in the console. I'd put it under `Edit` in the more menu because that's where most of these currently are, rather than having the floating IP name link directly to it.

https://console-fp16ezpjw-oxidecomputer.vercel.app/projects/mock-project/floating-ips

PR #2033, opened 2024-03-05
Add floating IP edit form

charliepark:
This PR uses the newly-updated API that enables updating Floating IP name/descriptions, and updates the UI to enable Floating IP updates.

Screenshot 2024-03-05 at 11 26 40 AM


mock-api/msw/handlers.ts
```diff
@@ -246,6 +246,20 @@ export const handlers = makeHandlers({
     return paginated(query, ips)
   },
   floatingIpView: ({ path, query }) => lookup.floatingIp({ ...path, ...query }),
+  floatingIpUpdate: ({ path, query, body }) => {
+    const floatingIp = lookup.floatingIp({ ...path, ...query })
+    if (body.name) {
+      // only check for existing name if it's being changed
+      if (body.name !== floatingIp.name) {
+        errIfExists(db.floatingIps, { name: body.name })
```
david-crespo, line 254: needs `project_id: floatingIp.project_id` to enforce name uniqueness for the project only

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -122,6 +123,14 @@ export function FloatingIpsPage() {
...
         }
     return [
+      {
+        label: 'Edit',
+        onActivate: () => {
+          navigate(pb.floatingIpEdit({ project, floatingIp: floatingIp.name }), {
+            state: floatingIp,
+          })
+        },
+      },
```
david-crespo, line 133: I don't think this `state` bit is meant to be in there, right?

Also: I had forgotten about this trick, but it's probably worth doing. I bet we've forgotten to do it on other edit forms. By taking the `floatingIp` object we already have on hand and putting it in RQ's cache, we can make the form open instantly because the loader's data is already there.

https://github.com/oxidecomputer/console/blob/3ad46af4f1923bbee040cea798d608dc4c059302/app/pages/ProjectsPage.tsx#L68-L77

test/e2e/floating-ip-update.e2e.ts
```diff
@@ -0,0 +1,61 @@
...
+    name: updatedName,
+    description: updatedDescription,
+  })
+})
+
+test('updating a Floating IP description works (even without updating name)', async ({
+  page,
+}) => {
+  await page.goto(floatingIpsPage)
+  await clickRowAction(page, originalName, 'Edit')
```
david-crespo, line 44: Just for coverage, I would change this test to navigate directly to the edit form by URL instead of clicking.

app/pages/project/floating-ips/FloatingIpsPage.tsx
charliepark: Ah, that's nice.

david-crespo (approved):

charliepark:
Fixes #2021

Issue #1895, opened 2024-01-19
Use top-level `disabled` prop on `useForm` to disable forms


david-crespo:
#1893 handles instance create, which is the worst offender because the POST takes a long time to come back. But strictly speaking, every other form has the same problem. We could manually wire up every form with `disabled` props on every field, like in #1893, but it would probably be better if we could find a more automatic way.

Someone asked about this here but didn't get an answer:

https://github.com/orgs/react-hook-form/discussions/11171
https://github.com/orgs/react-hook-form/discussions/11057

RHF recently (in [v7.48.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.48.0)) added a way to pass `disabled` to `useForm`, and then it might automatically get passed into all registered fields. Worth exploring! 

https://github.com/react-hook-form/react-hook-form/pull/10496

charliepark:
Crespo and I looked at this a bit today, and determined that — once we upgrade RHF — this should work in SideModalForm:
```
  // Disable side modal forms while submitting
  useEffect(() => {
    form.control._disableForm(loading)
  }, [form.control, loading])
```

`_disableForm` is currently undocumented, so Crespo also asked the RHF team directly about future plans for it: https://github.com/orgs/react-hook-form/discussions/11462

PR #2043, opened 2024-03-06
disable form elements when submitting instance create form

### LINKED ISSUE #2005: Disabled radio and upload fields on submit on instance create
Unsure if these are just visually not disabled – or actually interactive. But should be updated to match the others. Found in the advanced accordion.

charliepark:
Fixes #2005 

This PR expands the contexts where we use the `isSubmitting` value to disable form elements.

### DIFF
--- a/app/components/form/fields/FileField.tsx
+++ b/app/components/form/fields/FileField.tsx
@@ -25,6 +25,7 @@ export function FileField<
   required = false,
   accept,
   description,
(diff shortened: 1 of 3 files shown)


david-crespo (approved):
Looks great. This reminds me, we should try https://github.com/oxidecomputer/console/issues/1895 at some point.

PR #2046, opened 2024-03-07
Add CLI command button to settings group

### LINKED ISSUE #1991: Instances: Add equivalent CLI command for serial console
Can copy from the button inside the serial console page itself

benjaminleonard:
Fixes #1991 

I initially tried changing `cta` to take a react component, that way I could just pass the `<EquivalentCliCommand />` component together with the "Connect" button. But I think the issue with that approach is it's very easy to pass the wrong button style or size.

Instead I broke the modal into its own compound that way I could pass a `secondaryCta` prop and just call it directly from that.

CleanShot 2024-03-07 at 12 38 55

### DIFF
--- a/app/pages/project/instances/instance/tabs/ConnectTab.tsx
+++ b/app/pages/project/instances/instance/tabs/ConnectTab.tsx
@@ -6,36 +6,33 @@
  * Copyright Oxide Computer Company
  */
 
-import { useState } from 'react'
+import { Link } from 'react-router-dom'
 
-import { EquivalentCliCommandModal } from '~/components/EquivalentCliCommand'
+import { EquivalentCliCommand } from '~/components/EquivalentCliCommand'
 import { useInstanceSelector } from '~/hooks'
+import { buttonStyle } from '~/ui/lib/Button'
 import { SettingsGroup } from '~/ui/lib/SettingsGroup'
+import { cliCmd } from '~/util/cli-cmd'
 import { pb } from '~/util/path-builder'
-
-import { serialConsoleCliCommand } from '../SerialConsolePage'
 
 export function ConnectTab() {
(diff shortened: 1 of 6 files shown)


app/ui/lib/SettingsGroup.tsx
```diff
@@ -21,9 +21,20 @@ type Props = {
...
-export const SettingsGroup = ({ title, docs, children, cta, ctaText }: Props) => {
+export const SettingsGroup = ({
+  title,
+  docs,
+  children,
+  cta,
+  ctaText,
+  secondaryCta,
+  secondaryCtaText,
+}: Props) => {
```
david-crespo, line 37: I'm thinking this is an excessive level of abstraction for something that's used in one place. I think I can break this up into a few reusable containers and cut this down a lot.

app/pages/project/instances/instance/tabs/ConnectTab.tsx
```diff
@@ -6,20 +6,33 @@
...
-    >
-      Connect to your instance&rsquo;s serial console
-    </SettingsGroup>
+    <SettingsGroup.Container>
+      <SettingsGroup.Body>
+        <SettingsGroup.Title>Serial console</SettingsGroup.Title>
+        Connect to your instance&rsquo;s serial console
+      </SettingsGroup.Body>
+      <SettingsGroup.Footer>
+        <EquivalentCliCommand command={cliCmd.serialConsole({ project, instance })} />
```
david-crespo, line 28: with the refactors separated out, this PR would have been adding this one line

david-crespo:
In general I don't like the big props API approach in order to avoid mistakes at the call site. It eliminates some categories of mistake, but at great expense in terms of flexibility and you can still make all kinds of other mistakes.

PR #2048, opened 2024-03-08
Make SideModal submit button copy consistent

### LINKED ISSUE #984: Make edit submit button text consistent
I noticed a few patterns w/ the edit buttons.

`Save`, `Save Changes`, `Update X`

charliepark:
Fixes #984 

We had a few different strings we were using for the various edit forms. This PR aligns them, by adding a new `formType` (`'create' | 'edit'`) prop. For _edit_ forms, we have a default button label of "Save changes" (which we were already using in a few places).

This PR also removes an `id` prop, which we can generate dynamically, allowing us to clean the props up a little.

I'm ambivalent on whether `formType` should be a _required_ prop, as we're only using it to set the default label on `edit` forms. We could make it optional and clean up the various `create` form callsites.

### DIFF
--- a/app/components/form/SideModalForm.tsx
+++ b/app/components/form/SideModalForm.tsx
@@ -23,9 +23,11 @@ type ResourceName =
   | 'IP range'
   | 'network interface'
   | 'project'
+  | 'project image'
   | 'role'
   | 'rule'
   | 'silo'
+  | 'silo image'
   | 'snapshot'
   | 'SSH key'
   | 'subnet'
@@ -34,11 +36,13 @@ type ResourceName =
 type CreateFormProps = {
   formType: 'create'
   submitLabel?: string
+  title?: string
 }
 
 type EditFormProps = {
   formType: 'edit'
   submitLabel?: never
+  title?: never
 }
 
 type SideModalFormProps<TFieldValues extends FieldValues> = {
@@ -59,7 +63,6 @@ type SideModalFormProps<TFieldValues extends FieldValues> = {
   /** Error from the API call */
   submitError: ApiError | null
   loading?: boolean
-  title: string
   subtitle?: ReactNode
   onSubmit?: (values: TFieldValues) => void
 } & (CreateFormProps | EditFormProps)
@@ -98,7 +101,7 @@ export function SideModalForm<TFieldValues extends FieldValues>({
     }
   }, [submitError, form])
   const defaultTitle = formType === 'edit' ? `Edit ${resourceName}` : title
-  const defaultLabel = formType === 'edit' ? `Update ${resourceName}` : submitLabel || title
+  const label = formType === 'edit' ? `Update ${resourceName}` : submitLabel || title
 
   return (
     <SideModal
@@ -141,7 +144,7 @@ export function SideModalForm<TFieldValues extends FieldValues>({
             loading={loading || isSubmitting}
             form={id}
           >
-            {submitLabel || defaultLabel}
+            {label}
           </Button>
         )}
       </SideModal.Footer>
--- a/app/components/form/SideModalForm.tsx
+++ b/app/components/form/SideModalForm.tsx
@@ -14,6 +14,23 @@ import type { ApiError } from '@oxide/api'
 import { Button } from '~/ui/lib/Button'
 import { SideModal } from '~/ui/lib/SideModal'
 
+type ResourceName =
+  | 'disk'
+  | 'floating IP'
+  | 'identity provider'
+  | 'image'
+  | 'IP pool'
+  | 'IP range'
+  | 'network interface'
+  | 'project'
+  | 'role'
+  | 'rule'
+  | 'silo'
+  | 'snapshot'
+  | 'SSH key'
+  | 'subnet'
+  | 'VPC'
+
 type SideModalFormProps<TFieldValues extends FieldValues> = {
   form: UseFormReturn<TFieldValues>
   formType: 'create' | 'edit'
@@ -27,6 +44,7 @@ type SideModalFormProps<TFieldValues extends FieldValues> = {
    */
   children: ReactNode
   onDismiss: () => void
+  resourceName: ResourceName
   /** Must be provided with a reason describing why it's disabled */
   submitDisabled?: string
   /** Error from the API call */
@@ -53,6 +71,7 @@ export function SideModalForm<TFieldValues extends FieldValues>({
   formType,
   children,
   onDismiss,
+  resourceName,
   submitDisabled,
   submitError,
   title,
@@ -70,13 +89,14 @@ export function SideModalForm<TFieldValues extends FieldValues>({
       form.setError('name', { message: 'Name already exists' })
     }
   }, [submitError, form])
-  const defaultLabel = formType === 'edit' ? 'Save changes' : title
+  const defaultTitle = formType === 'edit' ? `Edit ${resourceName}` : title
+  const defaultLabel = formType === 'edit' ? `Update ${resourceName}` : submitLabel || title
 
   return (
     <SideModal
       onDismiss={onDismiss}
       isOpen
-      title={title}
+      title={defaultTitle}
       animate={useShouldAnimateModal()}
       subtitle={subtitle}
       errors={submitError ? [submitError.message] : []}
--- a/app/components/form/SideModalForm.tsx
+++ b/app/components/form/SideModalForm.tsx
@@ -108,7 +108,7 @@ export function SideModalForm<TFieldValues extends FieldValues>({
           <Button
             type="submit"
             size="sm"
-            disabled={!form.formState.isDirty || !!submitDisabled}
+            disabled={!!submitDisabled}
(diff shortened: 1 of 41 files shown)


david-crespo:
I'm on the fence too, but I think `formType` should be required so you have to make a decision rather than, for example, accidentally leaving it off on an edit form and getting the create behavior by default.

app/components/form/SideModalForm.tsx
```diff
@@ -15,8 +15,8 @@ import { Button } from '~/ui/lib/Button'
 import { SideModal } from '~/ui/lib/SideModal'
 
 type SideModalFormProps<TFieldValues extends FieldValues> = {
-  id: string
   form: UseFormReturn<TFieldValues>
+  formType: 'create' | 'edit'
```
david-crespo, line 19: Check this out — you can make TS error if an edit form is given a `submitLabel`. We can always change it later if we decide we want to be able to override it. And look! It caught one!

```diff
diff --git a/app/components/form/SideModalForm.tsx b/app/components/form/SideModalForm.tsx
index 3ad1ed84..28c34d4c 100644
--- a/app/components/form/SideModalForm.tsx
+++ b/app/components/form/SideModalForm.tsx
@@ -16,7 +16,6 @@ import { SideModal } from '~/ui/lib/SideModal'
 
 type SideModalFormProps<TFieldValues extends FieldValues> = {
   form: UseFormReturn<TFieldValues>
-  formType: 'create' | 'edit'
   /**
    * A function that returns the fields.
    *
@@ -35,8 +34,17 @@ type SideModalFormProps<TFieldValues extends FieldValues> = {
   title: string
   subtitle?: ReactNode
   onSubmit?: (values: TFieldValues) => void
-  submitLabel?: string
-}
+} & (
+  | {
+      formType: 'edit'
+      // submitLabel is not allowed on edit forms. we want them all to say "save changes"
+      submitLabel?: never
+    }
+  | {
+      formType: 'create'
+      submitLabel?: string
+    }
+)
 
 /**
  * Only animate the modal in when we're navigating by a client-side click.
diff --git a/app/forms/subnet-edit.tsx b/app/forms/subnet-edit.tsx
index 98137a6c..3fe0d924 100644
--- a/app/forms/subnet-edit.tsx
+++ b/app/forms/subnet-edit.tsx
@@ -48,7 +48,6 @@ export function EditSubnetForm({ onDismiss, editing }: EditSubnetFormProps) {
       }}
       loading={updateSubnet.isPending}
       submitError={updateSubnet.error}
-      submitLabel="Update subnet"
     >
       <NameField name="name" control={form.control} />
       <DescriptionField name="description" control={form.control} />
```

david-crespo (commented):
I like the change — it fixes the inconsistency and reduces choices we have to make at dev time.

It's a small thing, but since we're giving all the edit forms the same button label, I want to make sure we considered all possibilities. The plausible options I can think of are:

| Label | Comment |
|---|---|
| Save changes | Fine, though something about "changes" bothers me. "Save" makes me think of saving progress for later rather than submitting my update permanently |
| Save | Kinda short, plus same problem as "save" above | 
| Update | A little better than "save", I think, but similar | 
| Update [resource] | I might like this the best, but it's also the most annoying. We could do it by making the edit form take a resource name prop and have it plug it into both the title with `Edit ${resource}` and the button label with `Update ${resource} |

charliepark:
I'm on board with the `Edit`/`Update` `${resource}` option. Adding the `resource` prop won't be too painful, and I think it'll give the best effect.

app/components/form/SideModalForm.tsx
charliepark: nice

app/components/form/SideModalForm.tsx
```diff
@@ -14,8 +14,38 @@ import type { ApiError } from '@oxide/api'
...
+  | 'project'
+  | 'project image'
+  | 'role'
+  | 'rule'
+  | 'silo'
+  | 'silo image'
+  | 'snapshot'
+  | 'SSH key'
+  | 'subnet'
+  | 'VPC'
```
david-crespo, line 34: I think this should just be a string. I'm not worried about typoing `projec` or whatever.

app/forms/disk-create.tsx
```diff
@@ -100,9 +100,10 @@ export function CreateDiskSideModalForm({
 
   return (
     <SideModalForm
-      id="create-disk-form"
       title="Create Disk"
+      resourceName="disk"
```
david-crespo, line 104: Seems like we can get rid of `title` on almost all the creates too, that's cool

app/components/form/SideModalForm.tsx
```diff
@@ -69,12 +80,15 @@ export function SideModalForm<TFieldValues extends FieldValues>({
       form.setError('name', { message: 'Name already exists' })
     }
   }, [submitError, form])
+  const defaultTitle = title || `${formType === 'edit' ? 'Edit' : 'Create'} ${resourceName}`
+  const label =
+    submitLabel || title || `${formType === 'edit' ? 'Update' : 'Create'} ${resourceName}`
```
charliepark, line 85: this feels a little precious to me; I imagine there's a cleaner way to do this

charliepark: Actually, looking at where we use it, this is what we had before (`submitLabel || title`), plus the string literal (`${formType === 'edit' ? 'Update' : 'Create'} ${resourceName}`), so it isn't _too_ precious.

app/components/form/SideModalForm.tsx
```diff
@@ -27,16 +36,16 @@ type SideModalFormProps<TFieldValues extends FieldValues> = {
...
   children: ReactNode
   onDismiss: () => void
+  resourceName: string
   /** Must be provided with a reason describing why it's disabled */
   submitDisabled?: string
   /** Error from the API call */
   submitError: ApiError | null
   loading?: boolean
-  title: string
+  title?: string
```
david-crespo, line 45: I’d add a doc comment here explaining that you only need this when you want to override the default create or edit title. Same with `submitLabel` I think.

test/e2e/silo-access.e2e.ts
```diff
@@ -67,7 +67,7 @@ test('Click through silo access page', async ({ page }) => {
     .click()
   await page.click('role=menuitem[name="Change role"]')
 
-  await expectVisible(page, ['role=heading[name*="Change user role"]'])
+  await expectVisible(page, ['role=heading[name*="Edit role"]'])
```
david-crespo, line 70: “Change user role” is a little more descriptive? I could go either way. Also if it shows up for editing a group role too it’s misleading.

charliepark: Yeah, I agree that `Change` is probably more accurate than `Edit`, as they aren't editing the _role_, per se. `Change permissions`?

charliepark: On the actions modal we say `Change role`, so I think that makes sense. Should we determine whether it's a `user` or `group` and pass in the appropriate term? `Change user role` / `Change group role`

david-crespo: Permissions isn’t a term in the API, so I prefer to avoid it. Really what they’re doing in this case is updating the _policy_ on the resource, and the UI is sort of deliberately obscuring that (uncharacteristically) because it is so much more natural to think about operating on the user.

charliepark: Hold the phone. There was a TODO in the `project-access` and `silo-access` files noting that we should `// TODO: show user name in header or SOMEWHERE` … this seems like as good a time as any to update that in the modal's title.

charliepark: Screenshot 2024-03-08 at 5 42 00 PM

This copy feels a little clunky to me, but dynamic headers might just be like that.

david-crespo: Yea, it’s a bit weird but it’s still an improvement.

david-crespo (approved):

test/e2e/silo-access.e2e.ts
charliepark: Though it worked if the name was hypenated, it overran the boundary without hyphens …
Screenshot 2024-03-11 at 10 39 13 AM
… so I added `word-break: break-all` to the class for that header:
Screenshot 2024-03-11 at 10 50 42 AM

charliepark: Though …
Screenshot 2024-03-11 at 11 19 18 AM


I have an idea I'll try here.

david-crespo: what about `break-words` instead of `break-all`?

charliepark: I tried that, plus wrapping a container in `display: inline-block` and individual words in both `break-words` and `break-all`, but neither worked as desired 100% of the time. I think users will run into more situations with something like `Name ofPerson` or `hyphenated-group-name` as opposed to `unhyphenatedLongGroupName`, so `break-words` is probably the best solution for now.

Issue #2047, opened 2024-03-07
Add explanatory copy above IAM tables


david-crespo:
Effective role logic is somewhat opaque!

![image](https://uploads.linear.app/b7545512-bb0c-4743-adc4-e5797212a353/89b3f85b-6472-4c55-8352-965c6d0f11d8/640b2951-b462-48c3-b412-385cad732200?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2I3NTQ1NTEyLWJiMGMtNDc0My1hZGM0LWU1Nzk3MjEyYTM1My84OWIzZjg1Yi02NDcyLTRjNTUtODM1Mi05NjVjNmQwZjExZDgvNjQwYjI5NTEtYjQ2Mi00OGMzLWI0MTItMzg1Y2FkNzMyMjAwIiwiaWF0IjoxNzEwMjk2MzAzLCJleHAiOjMzMjgwODU2MzAzfQ.QSNSTpwv4EyuXfcsIefmmMUqYCA8GGUBI6Ups7-eUnw)

charliepark:
At some point, we'll want to think of a mechanism beyond color for highlighting the effective role, as (I believe) color alone is insufficient from an accessibility context.

One option: We could add an "Effective role" (or whatever we want to call it) column either before or after the existing role columns.

charliepark:
<img width="1215" alt="Screenshot 2024-03-08 at 9 38 42 AM" src="https://github.com/oxidecomputer/console/assets/22547/81933a07-5b87-473c-81e6-b7474ddba90d">

david-crespo:
We just talked about this, and I think what we want to convey is:

* You can get a role directly on your user or from a group membership
* A user or group can have a role on this resource and/or a parent
* The effective role is the strongest role from any of those sources

charliepark:
We also need to make sure that the group membership permissions are accurately represented as part of the individual user's effective role in console. I'm not sure that that's currently the case.

benjaminleonard:
We need to figure out a designed solution for this – some inline documentation would be helpful but the text directly above the tables does not work for me. Will review with Eugene and come back.

charliepark:
We are moving to show the "effective" role on the IAM tables; do we think we still need explanatory copy for this page? If it's a design pattern that we aren't happy with, I think it might be fine to close this issue.

The soon-to-be-merged UI:
[...]

PR #2049, opened 2024-03-08
Add column for effective role on access table

charliepark:
Open to discussion on whether we want to add this or not.

We have an "effective role", derived from the silo and project roles, which is essentially the "most permissive" role assigned to a user or group. We already highlight the effective role by marking it in green on the table, but it could be clearer, and accessibility best practices dictate that … "Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element." ([WCAG](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html))

#2047 requests explanatory copy above the table (and we can still add that), but I thought this might help make it more clear.

image

### DIFF
--- a/app/pages/project/access/ProjectAccessPage.tsx
+++ b/app/pages/project/access/ProjectAccessPage.tsx
@@ -129,6 +129,10 @@ export function ProjectAccessPage() {
   const columns = useMemo(
     () => [
       colHelper.accessor('name', { header: 'Name', cell: AccessNameCell }),
+      colHelper.accessor('effectiveRole', {
+        header: 'Effective role',
+        cell: RoleBadgeCell,
+      }),
       colHelper.accessor('siloRole', {
         header: 'Silo role',
         cell: RoleBadgeCell,


benjaminleonard:
I think that's a fair point, but rather than rotating the current implementation further I think we'd be better served working on implementing the [outstanding designs](https://www.figma.com/file/NcUHVumfS9qzinnZhanL4F/Access-%26-IAM?type=design&node-id=1%3A19&mode=design&t=Va9m1vC4CWJXnK0Y-1).

They're not completely up to date (we've got orgs in there which are no longer present) but I think it might solve this.

image

See how we only show the most permissive role, the others are shown in a tooltip (would also be visible in full on the user's pane).

image

charliepark:
Ah, that makes sense, and is definitely a nicer presentation of it.

david-crespo:
Yeah, I think we should just do the rework. It doesn't seem that hard to do, could be worth aiming to get it into v7.

PR #2055, opened 2024-03-12
Use list fetches  + `setQueryData` to prefetch items for `*NameFromId` cells

### LINKED ISSUE #1946: Combine list fetch and individual fetch for getting names from IDs
We have a number of places where we list resource A that points a resource B by ID, but we would obviously prefer to show the names of B:

- Instance attached to a disk
- VPC for network interface
- Silo linked to IP pool
- Disk source of snapshot

In those cases, we fetch the names one by one asynchronously, after the table has loaded. Example `*FromId` cell component:

https://github.com/oxidecomputer/console/blob/cfda1636ffb97ae986e994f5cf15f698128c26f0/app/pages/project/disks/DisksPage.tsx#L43-L59

This is not bad — our page size is 25, so in the worst case of a full page with 25 unique IDs, we're making 25 little requests in parallel, and most are below the fold. However! Inspired by @charliepark's work on #1871, we realized we could fetch the list of that thing (in the loader, in parallel!) and use it for the names, making only one call instead of 25. We could use a decent-sized limit (500?) to make it likely we're getting all of that thing, and then we can still fall back to fetching individually if the ID for resource B is not found in the list.

david-crespo:
Closes #1946

I love this. Using React Query's cache as a go-between, we can do the list prefetch idea with perfect fallback behavior (if the disk is not in the list we fetched, fetch it directly) without having to change the calling code.

I decided to limit each fetch to 200 items, as in general that should cover us. We could easily bump this higher if we wanted — no other code changes are required. In both FF and Chrome on M1 Mac, I timed how long it takes to loop through the items and call `setQueryData` on 200 items and it took 2-4ms. With a possibly more typical 10 items, it takes under 1ms as one would expect.

<details>
<summary>diff showing code changes for testing perf</summary>

```diff
diff --git a/app/pages/project/snapshots/SnapshotsPage.tsx b/app/pages/project/snapshots/SnapshotsPage.tsx
index 0d09cd89..9216acc4 100644
--- a/app/pages/project/snapshots/SnapshotsPage.tsx
+++ b/app/pages/project/snapshots/SnapshotsPage.tsx
@@ -67,6 +67,7 @@ SnapshotsPage.loader = async ({ params }: LoaderFunctionArgs) => {
     apiQueryClient
       .fetchQuery('diskList', { query: { project, limit: 200 } })
       .then((disks) => {
+        const t = performance.now()
         for (const disk of disks.items) {
           apiQueryClient.setQueryDataErrorsAllowed(
             'diskView',
@@ -74,6 +75,7 @@ SnapshotsPage.loader = async ({ params }: LoaderFunctionArgs) => {
             { type: 'success', data: disk }
           )
         }
+        console.log(performance.now() - t, 'ms')
       }),
   ])
   return null
diff --git a/mock-api/disk.ts b/mock-api/disk.ts
index c58ff18e..4b3cab6a 100644
--- a/mock-api/disk.ts
+++ b/mock-api/disk.ts
@@ -147,13 +147,13 @@ export const disks: Json<Disk>[] = [
     block_size: 2048,
   },
   // put a ton of disks in project 2 so we can use it to test pagination
-  ...new Array(55).fill(0).map((_, i) => {
-    const numStr = (i + 1).toString().padStart(2, '0')
+  ...new Array(1000).fill(0).map((_, i) => {
+    const numStr = (i + 30).toString().padStart(2, '0')
     return {
       id: '9747d936-795d-4d76-8ee0-15561f4cbb' + numStr,
       name: 'disk-' + numStr,
       description: '',
-      project_id: project2.id,
+      project_id: project.id,
       time_created: new Date().toISOString(),
       time_modified: new Date().toISOString(),
       state: { state: 'detached' as const },
```
</detail>

### DIFF
--- a/app/api/hooks.ts
+++ b/app/api/hooks.ts
@@ -219,6 +219,11 @@ export const wrapQueryClient = <A extends ApiClient>(api: A, queryClient: QueryC
     queryClient.invalidateQueries({ queryKey: [method], ...filters }),
   setQueryData: <M extends keyof A>(method: M, params: Params<A[M]>, data: Result<A[M]>) =>
     queryClient.setQueryData([method, params], data),
+  setQueryDataErrorsAllowed: <M extends keyof A>(
+    method: M,
+    params: Params<A[M]>,
+    data: ErrorsAllowed<Result<A[M]>, ApiError>
+  ) => queryClient.setQueryData([method, params, ERRORS_ALLOWED], data),
   fetchQuery: <M extends string & keyof A>(
     method: M,
     params: Params<A[M]>,
--- a/app/pages/project/disks/DisksPage.tsx
+++ b/app/pages/project/disks/DisksPage.tsx
@@ -44,9 +44,27 @@ const EmptyState = () => (
 )
 
 DisksPage.loader = async ({ params }: LoaderFunctionArgs) => {
-  await apiQueryClient.prefetchQuery('diskList', {
-    query: { ...getProjectSelector(params), limit: 25 },
-  })
+  const { project } = getProjectSelector(params)
+  await Promise.all([
+    apiQueryClient.prefetchQuery('diskList', {
+      query: { project, limit: 25 },
+    }),
+
+    // fetch instances and preload into RQ cache so fetches by ID in
+    // InstanceLinkCell can be mostly instant yet gracefully fall back to
+    // fetching individually if we don't fetch them all here
+    apiQueryClient
+      .fetchQuery('instanceList', { query: { project, limit: 200 } })
+      .then((instances) => {
+        for (const instance of instances.items) {
+          apiQueryClient.setQueryData(
+            'instanceView',
+            { path: { instance: instance.id } },
+            instance
+          )
+        }
+      }),
+  ])
   return null
 }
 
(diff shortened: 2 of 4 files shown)


app/api/hooks.ts
```diff
@@ -219,6 +219,11 @@ export const wrapQueryClient = <A extends ApiClient>(api: A, queryClient: QueryC
     queryClient.invalidateQueries({ queryKey: [method], ...filters }),
   setQueryData: <M extends keyof A>(method: M, params: Params<A[M]>, data: Result<A[M]>) =>
     queryClient.setQueryData([method, params], data),
+  setQueryDataErrorsAllowed: <M extends keyof A>(
+    method: M,
+    params: Params<A[M]>,
+    data: ErrorsAllowed<Result<A[M]>, ApiError>
+  ) => queryClient.setQueryData([method, params, ERRORS_ALLOWED], data),
```
david-crespo, line 226: This is needed because there's no other way to sneak `ERRORS_ALLOWED` into the query key to match `useApiQueryErrorsAllowed`.

charliepark (approved):
Amazing!

Also, thanks for including the diff w/ code changes for perf testing

Issue #2057, opened 2024-03-12
Disk: Display the image or snapshot name from which it is created


askfongjojo:
I find that I often need to recall from which snapshot or image a disk is made from. The information is currently available only in the fetch/view disk query via API or CLI. We can perhaps show this information in the UI, in a way similar to how disk names are displayed in the snapshot list <img width="1385" alt="image" src="https://github.com/oxidecomputer/console/assets/10874909/15176a8b-df9c-41ac-99c8-8482f42fa4b5">

There are two places to which such information can be added:
1. Instance - storage tab
<img width="1381" alt="image" src="https://github.com/oxidecomputer/console/assets/10874909/592b3754-c48a-4f66-b1aa-22e20acea21e">

2. Disk list 
<img width="1379" alt="image" src="https://github.com/oxidecomputer/console/assets/10874909/4a51e546-04fc-4351-93e6-03f819ec759e">

david-crespo:
Thank you, this sounds very useful.

askfongjojo:
@david-crespo: Wow, thanks for the quick turnaround on the prototype! In the PR, you raised a very interesting question about how/if we can present silo image information. I can think of the corner case that the silo image being demoted back into a project  subsequently. It'll not be appropriate to show its details for another project user at that point.

One other thing I am also wondering is the usefulness of disk/snapshot/image "genealogy". I have seen that kind of information in VMware for snapshots. In our case, it may look something like this:

disk aaa-1
|_ from image aaa, created on 2024-01-23 09:30 AM
|_ from snapshot yyy-3, created on 2024-01-22 12:30 PM
|_ from disk yyy, created on 2024-01-22 10:15 AM
|_ from snapshot xxx-1, created on 2024-01-15 11:28 PM
|_ from disk xxx, created on 2024-01-10 12:39: PM
...

The history can go on and on if the user keeps iterating on the same disk/snapshot/image family. The feature also has the caveat of exposing information from other projects. We can perhaps keep it bound by capping it some maximum number of generations and stop the traversal as soon as the origin falls outside of the current project. If you think this may be useful and not prone to major issues, I can make a separate ticket for such a feature and get some UX opinions on the presentation.

david-crespo:
Travis just requested this feature and I had totally forgotten about this issue and the prototype. It would be really useful. I'm going to put it in the Unscheduled milestone to make it slightly harder to forget.

PR #2058, opened 2024-03-12
Show disk origin

david-crespo:
Was curious what #2057 would look like. I think this has some issues to resolve with design.

* We don't have a snapshot detail URL to link to from the snapshot cell
* Snapshot may no longer exist — do we fall back to deleted like we do with a snapshot of a deleted disk? probably
* We do have an image detail URL to link to, but the image may be silo or project image and the link URL depends on which one it is, but we don't actually know from the `imageView` response which it is

image

### DIFF
--- a/app/pages/project/disks/DisksPage.tsx
+++ b/app/pages/project/disks/DisksPage.tsx
@@ -12,6 +12,7 @@ import {
   diskCan,
   genName,
   useApiMutation,
+  useApiQuery,
   useApiQueryClient,
   type Disk,
 } from '@oxide/api'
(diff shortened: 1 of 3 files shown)


benjaminleonard:
Like it – I can see it being especially useful when we finish the filtering work.

1. Adding a snapshot pane would be a relatively small piece even without edit we can just show the resource long-form. I think it's useful to reinforce the UX pattern of clickable resources in tables.
2. If the snapshot is deleted do we just get a 404? Might make the case for the disk audit log as suggested by Angela (https://github.com/oxidecomputer/console/issues/2057#issuecomment-1992719212). That way we can include the source but just not link to it.
3. There's a gross solution which is to do a full silo image and project image fetch. We might consider adding the image type to the response.

david-crespo:
Closing this since it was just an experiment.

PR #2060, opened 2024-03-12
Highlight sidebar nav links when in create form

### LINKED ISSUE #2038: `Instances` in sidebar not highlighted on instance create page
Instance create is `/projects/mock-project/instances-new` and the sidebar link is `/projects/mock-project/instances`.

image

charliepark:
Fixes #2038 

This PR updates how the Sidebar evaluates the nav links and how their `to` destination compares to the current URL.

With this change, a create form for a resource (e.g. `/instances-new`, `/disks-new`, etc.) will now show up highlighted, to conform to the parent resource page.
The Instances page and `-new` page:
Screenshot 2024-03-12 at 3 58 19 PM
Screenshot 2024-03-12 at 3 58 28 PM
And the Disks page and `-new` page:
Screenshot 2024-03-12 at 3 58 40 PM
Screenshot 2024-03-12 at 3 58 59 PM

When the form is contained in a SideModal (see `/disks-new`, above), the nav link is dimmed a bit.


app/components/Sidebar.tsx
```diff
@@ -88,20 +88,30 @@ export const NavLinkItem = (props: {
...
-)
+}) => {
+  // "New" resource create forms have a url that doesn't match the top-level resource name, so `isActive` won't ever fire as true.
+  // Determine which resource in the Sidebar should be highlighted as active when on a create-form page.
+  const location = useLocation()
+  const resourcePath = location.pathname.split('/').pop()
+  const resourceName = resourcePath?.includes('-new')
+    ? resourcePath.replace('-new', '')
+    : ''
+  const isLinkToThisResource = resourceName.length > 0 && props.to.includes(resourceName)
```
david-crespo, line 99: You should extract this chunk of logic into a util function that takes `to` and `pathname` and returns the bool, then you can unit test it

app/util/links.ts
```diff
@@ -10,3 +10,9 @@ export const links: Record<string, string> = {
   cloudInitExamples: 'https://cloudinit.readthedocs.io/en/latest/reference/examples.html',
   ipPoolsDocs: 'https://docs.oxide.computer/guides/operator/ip-pool-management',
 }
+
+// Resource creation pages (e.g. /instances-new) don't have a route that matches the top-level resource name,
+// so `isActive` won't ever fire as true, and the Sidebar won't highlight the correct resource.
+// Determine, for a given Sidebar URL, whether it's the matching top-level resource URL for the current page.
+export const isRootResourceLink = (navUrl: string, locationPathname: string): boolean =>
+  locationPathname === `${navUrl}-new`
```
david-crespo, line 18: I hate to reverse course here, but now that the logic is so simple, I think the unit test is probably unnecessary and I find all these names in the extracted version more confusing than helpful. Kinda don’t think the unit tests are necessary either. Just moving the key bit inline in `NavLinkItem` with a comment would be easier to understand.

```ts
const onCreatePage = useLocation().pathname === `${props.to}-new`
```

app/components/Sidebar.tsx
```diff
@@ -88,20 +89,24 @@ export const NavLinkItem = (props: {
...
-      }
-      end={props.end}
-    >
-      {props.children}
-    </NavLink>
-  </li>
-)
+}) => {
+  const location = useLocation()
+  const isLinkToRootResource = isRootResourceLink(props.to, location.pathname)
```
david-crespo, line 94: I see why you picked the name `isRootResourceLink`, but to me, because it’s inside the nav link component, it’s kind of backwards. This name says: from the point of view of the create route, is this link relevant to me due to being my root resource? But we’re inside the link here, not the create form, so to me the more intuitive way to put it would be: from the point of view of this link, are we currently on the corresponding create form route? Hence `onCreatePage` in my other suggestion. Could be `onCreateFormRoute` or whatever.

charliepark:
Ah, crap. A rebase possibly messed this up. This should have been simplified down to just one or two file changes, total, so it might be easier for me to just do this in a new branch.

PR #2061, opened 2024-03-13
Add Type column to access table

charliepark:
We'll be making a few changes to the Access table; one small one is pulling the "type" badge out of the Name column and creating a new Type column.

Here's the end state from Figma:
image

This PR extracts the badge out to its own column on the silo and project access pages:
Screenshot 2024-03-12 at 8 43 03 PM
Screenshot 2024-03-12 at 8 43 11 PM


app/components/AccessNameCell.tsx
```diff
@@ -7,28 +7,9 @@
...
-      <span>{name}</span>
-      {identityType === 'silo_group' ? (
-        <Badge color="neutral" className="ml-2">
-          Group
-        </Badge>
-      ) : null}
-    </>
-  )
-}
+) => <span>{info.getValue()}</span>
```
david-crespo, line 15: Most likely this doesn't need to exist, unless you need the `<span>` there for some reason. If you leave `cell` unspecified, React Table will just stick `info.getValue()` in there for you.

app/components/AccessTypeCell.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Display whether this row displays a User or Group.
+ */
+export const AccessTypeCell = <RowData extends { identityType: IdentityType }>(
+  info: CellContext<RowData, string>
+) => {
+  const identityType = info.row.original.identityType
+  return <span>{identityType === 'silo_group' ? 'Group' : 'User'}</span>
+}
```
david-crespo, line 20: Now that this cell is dedicated to `identityType` and used in a column where the accessor is specified as `identityType`, you don't have to do shenanigans to get that value — it's now in `info.getValue()`. It can also be a fragment instead of a span.

```suggestion
export const AccessTypeCell = <RowData extends { identityType: IdentityType }>(
  info: CellContext<RowData, IdentityType>
) => {
  return <>{info.getValue() === 'silo_group' ? 'Group' : 'User'}</>
}
```

Another approach, driven by my desire to get rid of all the noise in the type signature, would be to decouple this from React Table altogether and do something like this:

```ts
export const accessTypeLabel = (identityType: IdentityType) =>
  identityType === 'silo_group' ? 'Group' : 'User'
```

```ts
colHelper.accessor('identityType', {
  header: 'Type',
  cell: (props) => accessTypeLabel(props.getValue()),
}),
```

That's how they tend to do it in the RT docs: 

https://tanstack.com/table/latest/docs/guide/column-defs#cell-formatting

We in the past have tended toward these `Cell` components that let you pass them directly as `cell: WhateverCell` because it looks cleaner at the call site, but I think the complexity of this refactor shows the downside.

david-crespo (approved):
now we're talking. beautiful

PR #2062, opened 2024-03-13
Add access roles column

charliepark:
This is a follow-on for https://github.com/oxidecomputer/console/pull/2061.

This PR changes from the current "Silo role" and "Project role" columns to a single "Role" column, showing the "effective" role for each user or group.

Screenshot 2024-03-19 at 5 21 29 PM

It also adds a new `ExpandedCountWithDetails` component, which gives us a count and a tooltip that, when hovered over, gives us some additional content.
Screenshot 2024-03-19 at 5 23 00 PM
This is a fairly simple first-run component, but we can add more details down the road if we need to. (For example, a design in Figma includes multiple (divided) sections of content as one possible scenario.)


benjaminleonard:
> Figma shows a +2, as well as a mix of "User roles" and "Inherited roles". I haven't yet sussed out whether that's an out-of-date data model or whether that still applies and I should account for it in the logic here.

The idea was originally that a user role is one applied directly to a single user but the inherited roles comes from a group the user is a member of. Perhaps the issue with that model is that almost all roles come from a group – not even sure if we allow direct role application right now. We can skip that.

app/components/AccessBadge.tsx
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import type { RoleKey } from '~/api'
+import { Badge } from '~/ui/lib/Badge'
+import { getBadgeColor } from '~/util/access'
+
+type AccessBadgeProps = { labelPrefix: 'silo' | 'project'; role: RoleKey }
+
+export const AccessBadge = ({ labelPrefix, role }: AccessBadgeProps) => (
+  <Badge color={getBadgeColor(role)}>{`${labelPrefix}.${role}`}</Badge>
```
david-crespo, line 16: unless I'm missing something

```suggestion
  <Badge color={getBadgeColor(role)}>{labelPrefix}.{role}</Badge>
```

app/components/ProjectAccessRolesCell.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <div className="flex items-baseline gap-1">
+      {effectiveRole}
+      {siloRole && projectRole ? (
+        <Tooltip content={alternateRole} placement="top">
+          <>+1</>
+        </Tooltip>
+      ) : undefined}
+    </div>
```
david-crespo, line 53: Keep in the back of your mind whether it would make sense to pull out a component for this +1 list concept. I was thinking, for example, of adding a `silos` column on IP pools, which would benefit from this treatment.

image

app/components/AccessBadge.tsx
charliepark: 🤦 womp

app/components/ProjectAccessRolesCell.tsx
charliepark: Sounds good. Have added ExpandedCountWithDetails, which we can adjust further as future needs dictate.

david-crespo:
52198863d30d53bff9c43a90f3bb25f2c33bae76 adds this. It's still not ideal because as @charliepark noted, we're not factoring in roles the user gets from groups because we don't know the groups. But I think it's helpful on net, and it lets us avoid putting text above the table.

Screenshot 2024-03-22 at 10 29 51 AM

david-crespo (approved):

Issue #2063, opened 2024-03-13
Firewall rules: Provide capability to easily set up new rules by copying an existing one


askfongjojo:
Setting up new firewall rules can be complicated for first-time cloud users. The top use case that necessitate users to fiddle with firewall is to open ports to applications running in the instances. Since we already have a default rule to allow SSH, it'll be much easier for users to simply copy that firewall rule and modify the port number section.

askfongjojo:
Yes, the pattern looks intuitive enough to me. Once the user knows the clone feature lives under "...", they should be able to find it everywhere.

I wonder if we can have a little note near the top of the "Create Instance" page to mention something along the lines of
> If you would like to create a new instance by cloning from another instance, you can select the "New Similar Instance" action next to the instance to clone from. The action presets the new instance configurations with the values of that existing instance.

PR #2067, opened 2024-03-14
Instance template

benjaminleonard:
Resurrecting this experiment as a discussion / reference for #2063. Came out of a suggestion from @taspelund.

The idea being that the user can use a resource as a template to create a new one. In this case the instance create form is pre-filled with the appropriate fields from another one. It essentially passes another instance in the URL param, which is then retrieved and sets the values in the form.

This is a quick test implementation, not a model for actually doing it. In practice we might be able to use get the instance in the loader and pass the values to `defaultValues`, avoiding the use of `setValues` all together.

CleanShot 2024-01-26 at 11 49 16


david-crespo:
Very cool, love the use of the query param. Definitely going to want to do the loader + defaultValues approach.

benjaminleonard:
I think we can close this and redo it later based on #2250

PR #2073, opened 2024-03-15
Add TableResourceDescription

### LINKED ISSUE #1967: Link to docs from floating IPs page
https://docs.oxide.computer/guides/managing-floating-ips

We could also embed the link in a couple of sentences of explanation at the top of the list of floating IPs, like on the IP pool silos list.

### LINKED ISSUE #2056: Create standard component for "descriptive copy above a table"
We currently have a few places where we use `<p className="mr-8 max-w-2xl text-sans-md text-secondary">` (or a close variation; check!), and we should standardize that and just pass in / wrap ReactNode contents.

charliepark:
Fixes #1967 
Fixes #2056 

This PR sets up a few components to standardize the templates around tables. As Eugene and Ben have both pointed out, the "descriptive copy" above the table isn't something we want to keep around for long, and will remove as soon as we have a better design in place for it.

The "TableControls" naming for the components is a _bit_ future-looking, as we intend to add various controls — filters, sorting options, etc. — down the road, and @david-crespo and I figured we could use this pattern as a foundation. Obviously, if the design changes radically regarding the table intro, we can revisit that decision.

Screenshot 2024-03-15 at 4 49 06 PM

Screenshot 2024-03-15 at 4 48 53 PM


app/pages/system/silos/SiloIpPoolsTab.tsx
```diff
@@ -142,19 +143,22 @@ export function SiloIpPoolsTab() {
...
   ]
 
+  const resourceInformation = (
+    <>
+      Users in this silo can allocate external IPs from these pools for their instances. A
+      silo can have at most one default pool. IPs are allocated from the default pool when
+      users ask for one without specifying a pool. Read the docs to learn more about{' '}
+      <ExternalLink href={links.ipPoolsDocs}>managing IP pools</ExternalLink>.
+    </>
+  )
```
david-crespo, line 153: I think the need or least the temptation to pull this out to make it look clean is a point against the props approach. Compare to

```tsx
<AboveTable>
  <AboveTable.Text>
    Users in this silo can allocate external IPs from these pools for their instances.
    A silo can have at most one default pool. IPs are allocated from the default pool
    when users ask for one without specifying a pool. Read the docs to learn more
    about <ExternalLink href={links.ipPoolsDocs}>managing IP pools</ExternalLink>.
  </AboveTable.Text>
  <AboveTable.Button onClick={() => setShowLinkModal(true)}>
    Link pool
  </AboveTable.Button>
</AboveTable>
```

And all the definitions get small and simple:

```tsx
const AboveTable = classed.div`mb-4 flex items-end justify-between space-x-8`
AboveTable.Text = classed.p`max-w-2xl text-sans-md text-secondary`
AboveTable.Button = (props: ButtonProps) => <Button size="sm" className="shrink-0" {...props} />
AboveTable.Link = (props: LinkProps) => <Link className={buttonStyle({ size: 'sm' }) {...props} />
```

app/ui/lib/TableActionButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const TableActionButton = ({ label, linkTo, onClick }: TableActionButtonProps) =>
+  linkTo ? (
+    <Link to={linkTo} className={buttonStyle({ size: 'sm' })}>
+      {label}
+    </Link>
+  ) : (
+    <Button onClick={onClick} size="sm" className="shrink-0">
+      {label}
+    </Button>
+  )
```
david-crespo, line 29: I do kind of like this `ButtonOrLink` concept, but I feel like maybe it should be more general, so we could use it in other places. I almost want to be able to do something like this:

```tsx
AboveTable.LinkOrButton = makeLinkOrButtonComponent({ size: 'sm' })
```

on the other hand, once you have this concept of the component factory, you don't _really_ have to combine them into one component. If the point is to make clear in one spot what styling you're supposed to use regardless of whether it's a link or button, you can just have two components, as long as they're lightweight to define and intuitive to call. So, like I said above, think something like this would be ok:

```tsx
export const AboveTableButton = (props: ButtonProps) => (
  <Button size="sm" className="shrink-0" {...props} />
)
export const AboveTableLink = (props: LinkProps) => (
  <Link className={buttonStyle({ size: 'sm' })} {...props} />
)
```

though the props spreads always make me nervous.

benjaminleonard:
As Eugene says [here](https://github.com/oxidecomputer/console/issues/2056#issuecomment-1992141217), I think this is fine as temporary solution. But we should aim to design the text above tables away when we can, it feels like the UX equivalent of a code smell.

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -161,11 +162,16 @@ export function FloatingIpsPage() {
...
       </PageHeader>
-      <TableActions>
-        <Link to={pb.floatingIpsNew({ project })} className={buttonStyle({ size: 'sm' })}>
+      <TableControls>
+        <TableControlsText>
+          Floating IPs are public IP addresses that can be attached to instances. They allow
+          your instances to be reachable from the internet. Find out more about{' '}
+          <ExternalLink href={links.floatingIpsDocs}>managing floating IPs</ExternalLink>.
+        </TableControlsText>
+        <TableControlsButton onClick={() => navigate(pb.floatingIpsNew({ project }))}>
```
david-crespo, line 171: this should be a link haha

david-crespo (approved):
looks great minus that one link

app/pages/project/floating-ips/FloatingIpsPage.tsx
charliepark: image

charliepark: Thank you for spotting that. Will patch that up.

Issue #2075, opened 2024-03-18
Reconsider active selected fields across multiple tabs


benjaminleonard:
Found on the instance create form – for example with the boot disk. In this case the bar bar becomes its own input – "Source type", but the interaction is unclear.

<img src="https://uploads.linear.app/b7545512-bb0c-4743-adc4-e5797212a353/d4f0fb34-955c-4862-992b-5b4c6a88dd40/d4b2d9cc-080c-4cd3-b42c-e9317cde72a0?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2I3NTQ1NTEyLWJiMGMtNDc0My1hZGM0LWU1Nzk3MjEyYTM1My9kNGYwZmIzNC05NTVjLTQ4NjItOTkyYi01YjRjNmE4OGRkNDAvZDRiMmQ5Y2MtMDgwYy00Y2QzLWI0MmMtZTkzMTdjZGU3MmEwIiwiaWF0IjoxNzEwNzU3MzM1LCJleHAiOjMzMjgxMzE3MzM1fQ.b8bBCbnU8nDA0ZZYoBaYqT55OT5_B7M6g3pqYx02Qig " alt="image.png" width="690" height="602" />

The easiest fix would be to use a regular radio group. But that makes switching type feel smaller than it should.

Also considering opening a modal, and clicking save or select to apply the boot disk. And the result can show up summarised in a mini table. It's clearer I think – but does add an extra step to open the modal first.

PR #2076, opened 2024-03-18
Create instance from existing boot disk

### LINKED ISSUE #1060: Choose existing disk as a boot disk on instance create
... but I should be able to, right?

Steps to repro:

- Create an instance, use an existing image as the disk of choice (I used debian, you can use whatever)
- Observe the disk and instance exist
- Stop the instance, delete it
- Observe the boot disk still exists
- Try to create a new instance using the old boot disk
- Observe that "boot disk" requires one of "{distros, images, snapshots}". Old boot disks do not show up here. Notably it can be attached as an "additional disk", but not as an old boot disk.

benjaminleonard:
Fixes #1060

This exacerbates the confusion with #2075 – where it is unclear what happens to the selected value when a user switches tabs. That doesn't necessarily need to be a blocker for this, would prefer to think through a design solution first.

CleanShot 2024-03-18 at 12 32 10

Also seeing a flash from the navigation blocker after submission. Unsure if that's something I introduced with this PR by accident.

- [x] Fix click from number input to tab bar not setting value
- [x] Text that explains which disks are usable


benjaminleonard:
Also running into an issue where the disk size input does not apply if you exit focus by clicking on the tab bar. Maybe the `setValue` on `bootDiskSource` interrupts the disk size input applying its own value. This is causing a failed E2E test.

CleanShot 2024-03-18 at 12 36 44

app/forms/instance-create.tsx
```diff
@@ -95,11 +106,23 @@ const baseDefaultValues: InstanceCreateInput = {
   userData: null,
 }
 
+const useBootDiskItems = (projectSelector: PP.Project) => {
+  const { data: disks } = useApiQuery('diskList', {
+    query: { ...projectSelector, limit: 1000 },
+  })
+  return (
+    disks?.items
+      .filter(diskCan.attach)
```
benjaminleonard, line 115: We might want to include disks that cannot be attached, or add a tooltip that lets the user know which disks are included and why.

charliepark:
> Also running into an issue where the disk size input does not apply if you exit focus by clicking on the tab bar. Maybe the `setValue` on `bootDiskSource` interrupts the disk size input applying its own value. This is causing a failed E2E test.

This appears to be fixed now

charliepark (approved):
This looks good; sorry for the delayed approval. Was running into issues that stemmed from me not restarting vite, but this is solid.

app/forms/instance-create.tsx
```diff
@@ -334,8 +367,14 @@ export function CreateInstanceForm() {
         className="full-width"
         // default to the project images tab if there are only project images
         defaultValue={
-          projectImages.length > 0 && siloImages.length === 0 ? 'project' : 'silo'
+          siloImages.length > 0 ? 'silo' : projectImages.length > 0 ? 'project' : 'disk'
```
david-crespo, line 370: nice

app/forms/instance-create.tsx
```diff
@@ -344,12 +383,15 @@ export function CreateInstanceForm() {
...
+            Existing disks
+          </Tabs.Trigger>
         </Tabs.List>
-        {allImages.length === 0 && (
+        {allImages.length === 0 && disks.length === 0 && (
           <Message
             className="mb-8 ml-10 max-w-lg"
             variant="notice"
-            content="Images are required to create a boot disk."
+            content="Images or disks are required to create a boot disk."
```
david-crespo, line 394: Not that important because very few people should ever see this. But this is confusing because if you're picking a disk you're not creating. The only thing I can think of is

> Images or disks are required to create or attach a boot disk.

app/forms/instance-create.tsx
```diff
@@ -388,29 +430,57 @@ export function CreateInstanceForm() {
...
           )}
         </Tabs.Content>
+
+        <Tabs.Content value="disk" className="space-y-4">
+          {disks.length === 0 ? (
+            <div className="flex max-w-lg items-center justify-center rounded-lg border p-6 border-default">
+              <EmptyMessage
+                icon={<Storage16Icon />}
+                title="No detached disks found"
+                body="A disk needs to be created and not attached to an instance to be used as a boot disk"
```
david-crespo, line 440: I get where this comes from, but "A disk needs to be created" is a bit computer-brained. Gonna tweak it to this

image

app/forms/instance-create.tsx
```diff
@@ -95,11 +106,24 @@ const baseDefaultValues: InstanceCreateInput = {
   userData: null,
 }
 
+const useBootDiskItems = (projectSelector: PP.Project) => {
+  const { data: disks } = useApiQuery('diskList', {
+    query: { ...projectSelector, limit: 1000 },
```
david-crespo, line 111: Tested this with 500 disks in the response and it doesn't seem to get bogged down at all.

benjaminleonard:
Great, copy changes look good to me

charliepark:
I have a few fixes in https://github.com/oxidecomputer/console/pull/2097 that I'm close to merging here

app/forms/instance-create.tsx
```diff
@@ -134,22 +176,45 @@ export function CreateInstanceForm() {
...
+  const disks = useMemo(
+    () => allDisks.filter(diskCan.attach).map(({ name }) => ({ value: name, label: name })),
+    [allDisks]
+  )
+
   const { data: sshKeys } = usePrefetchedApiQuery('currentUserSshKeyList', {})
   const allKeys = useMemo(() => sshKeys.items.map((key) => key.id), [sshKeys])
 
+  const defaultSource =
+    siloImages.length > 0 ? 'siloImage' : projectImages.length > 0 ? 'projectImage' : 'disk'
```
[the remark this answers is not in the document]
charliepark: Actually, we overwrite it on [line 195](https://github.com/oxidecomputer/console/pull/2076/files#diff-9083077524dfe071711dffdfb0fa64e1ef80a5518a494a1a1bdfc75af0b1ec9aR195), so probably fine.

david-crespo (approved):
Great work @benjaminleonard and @charliepark

PR #2078, opened 2024-03-18
IP pool utilization

### LINKED ISSUE #2070: Show IP pool utilization on IP pool list and detail
Console side of https://github.com/oxidecomputer/omicron/pull/5258

david-crespo:
Closes #2070 

- [x] Show v4 and v6 counts in a column on IP pools table
- [x] Fancy display logic for big numbers
- [x] Show v4 ~~and v6~~ counts in properties table on IP pool detail
- [x] Make mock handler use the actual ranges
- [x] Put utilization on IP pool detail page

https://github.com/oxidecomputer/console/assets/3612203/8eabd4d7-5b8c-415f-a770-32661a01520d


app/util/math.spec.ts
```diff
@@ -59,6 +59,20 @@ describe('with default locale', () => {
...
+    [155n, '155'],
+    [999999n, '999,999'],
+    [9999999n, '10M'],
+    [492038458320n, '492B'],
+    [894283412938921, '894.3T'],
+    [1293859032098219, '1.3E15'],
+    [23094304823948203952304920342n, '23.1E27'],
+  ])('displayBigNum %d -> %s', (input, output) => {
+    expect(displayBigNum(input)).toEqual(output)
+  })
```
david-crespo, line 75: these bigint literals are the reason for the `target: es2020` bump in the tsconfig

tsconfig.json
```diff
@@ -21,7 +21,7 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es2015",
+    "target": "es2020",
```
david-crespo, line 24: This is so we can use bigint literals, e.g., `1123901234n`. `target` seems primarily relevant in that it determines `lib` if you're not also setting `lib` (we are not). There is no change in our bundle size from this, which is weird. Will add more detail to this comment later.

https://www.typescriptlang.org/tsconfig#target
https://www.typescriptlang.org/tsconfig#lib

david-crespo: According to the [compat table](https://compat-table.github.io/compat-table/es2016plus/#firefox120) (ironically, works poorly in FF but well in Chrome), these are the features for each version:

image

These are the ones for es2020 specifically, and they all have been supported since 2019 or early 2020.

* https://caniuse.com/mdn-javascript_builtins_string_matchall
* https://caniuse.com/bigint
* https://caniuse.com/mdn-javascript_builtins_promise_allsettled
* https://caniuse.com/mdn-javascript_builtins_globalthis
* https://caniuse.com/mdn-javascript_operators_optional_chaining
* https://caniuse.com/mdn-javascript_operators_nullish_coalescing

I looked through most of the other features and I think we could go as high as es2022, but I don't think there's much advantage to doing that.

david-crespo: Ah. Of course. This explains why it had no effect on bundle size.

> We should also note that TypeScript's target only impacts your runtime code if you're using TypeScript to transpile your source files to JavaScript. If you're using another tool to transpile, such as [Babel](https://babeljs.io/), [ESBuild](https://esbuild.github.io/), or [SWC](https://swc.rs/), then that tool's settings will be what determines the syntax used in your output JavaScript code.

https://www.learningtypescript.com/articles/why-increase-your-tsconfig-target#target-might-not-impact-your-runtime-code

I'm embarrassed not to have known this, but here's how Vite handles it.

> **build.target**
>
> Browser compatibility target for the final bundle. The default value is a Vite special value, 'modules', which targets browsers with [native ES Modules](https://caniuse.com/es6-module), [native ESM dynamic import](https://caniuse.com/es6-module-dynamic-import), and [import.meta](https://caniuse.com/mdn-javascript_operators_import_meta) support. Vite will replace 'modules' to ['es2020', 'edge88', 'firefox78', 'chrome87', 'safari14']

> Note the build will fail if the code contains features that cannot be safely transpiled by esbuild. See [esbuild docs](https://esbuild.github.io/content-types/#javascript) for more details.

https://vitejs.dev/config/build-options.html#build-target

> Note that by default, Vite only handles syntax transforms and does not cover polyfills.

https://vitejs.dev/guide/build.html#browser-compatibility

The linked page from the esbuild docs about what features are in what version is better than anything I found yesterday while researching this stuff. 

https://esbuild.github.io/content-types/#javascript

It also gave me a good example of too-new syntax to test it with:

```
x Build failed in 7.97s
error during build:
Error: [vite:esbuild-transpile] Transform failed with 1 error:
assets/app-!~{001}~.js:62912:0: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)

Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
62910|  };
62911|  const colHelper = createColumnHelper();
62912|  await new Promise((res) => res(1));
   |  ^
62913|  function SiloAccessPage() {
62914|    const [addModalOpen, setAddModalOpen] = reactExports.useState(false);
```

### Conclusion

So in theory, Vite has us covered for making sure we are okay on syntax, and TS `target` should prevent us from using new built-in things we can't use. That said, I couldn't find a good example of something like that being rejected by TS.

app/components/CapacityBar.tsx
```diff
@@ -60,16 +65,16 @@ export const CapacityBar = ({
...
               <span className="normal-case">{includeUnit ? ' ' + unit : ''}</span>
             </div>
           </div>
           <div className="p-3 text-mono-sm">
-            <div className="text-quaternary">{allocatedLabel}</div>
-            <div className="text-secondary">
-              {allocated.toLocaleString()}
+            <div className="text-quaternary">{capacityLabel}</div>
+            <div className="!normal-case text-secondary">
+              {displayBigNum(capacity)}
```
david-crespo, line 77: This component is getting out of hand. I will make an issue about breaking it up into composable components.

app/components/CapacityBar.tsx
```diff
@@ -6,30 +6,30 @@
  * Copyright Oxide Computer Company
  */
 
-import { splitDecimal } from '~/util/math'
+import { BigNum } from '~/ui/lib/BigNum'
+import { percentage, splitDecimal } from '~/util/math'
 
-export const CapacityBar = ({
+export const CapacityBar = <T extends number | bigint>({
```
david-crespo, line 12: generic is so we can let provisioned and allocated be `number | bigint` while enforcing that they are the same type

app/components/IpPoolUtilization.tsx
```diff
@@ -0,0 +1,50 @@
...
+export function IpUtilCell(util: IpPoolUtilization) {
+  const { ipv4, ipv6 } = parseIpUtilization(util)
+
+  if (ipv6.capacity === 0n) {
+    return (
+      <div className="space-y-1">
+        <IpUtilFrac {...ipv4} />
+      </div>
+    )
+  }
```
david-crespo, line 29: This should be what users see because they almost certainly have no IPv6 ranges.

app/pages/system/networking/IpPoolPage.tsx
```diff
@@ -93,6 +103,43 @@ export function IpPoolPage() {
...
+          provisionedLabel="Allocated"
+          capacityLabel="Capacity"
+          unit="IPs"
+          includeUnit={false}
+        />
+      )}
+    </div>
+  )
+}
+
```
david-crespo, line 142: took a slightly different approach here than the possibly too clever "if there is IPv6, show both, otherwise just show IPv4 with no v4 label" in the table cell. Here we just show whichever ones are present, including possibly both or none.

app/util/math.spec.ts
```diff
@@ -99,10 +149,62 @@ describe('with de-DE locale', () => {
...
     Object.defineProperty(global.navigator, 'language', {
       value: originalLanguage,
       writable: true,
     })
   })
 })
+
+// the point of these tests is to make sure the toLowerCase shenanigan in
+// toEngNotation doesn't go horribly wrong due some obscure locale's concept of
+// engineering notation
```
charliepark, line 178: Love that you included these

charliepark (approved):
Really nice!

PR #2085, opened 2024-03-21
Refactor CapacityBar

### LINKED ISSUE #2081: Break up `CapacityBar` into composable components
The usual thing where you start with a props API and then eventually you add more props for overrides, plus default values for the original call site, and you have a bunch of conditional logic inside the component.

charliepark:
Fixes #2081 

I left everything inside the `CapacityBar` component, rather than creating a bunch of other files, though if it'd be cleaner if we broke it up further, I'm happy to do that.

It's possible I'm being too clever with the, say, `titleOptions`, `unitOptions`, `iconOptions`, if we decide to use these bars for wholly other kinds of bar charts. At the same time, adding new options to each `Options` object is pretty straightforward, and keeps the props design simpler than it had been before. Open to feedback, as always, on that.

The end UI isn't any different than the current layout. These two screenshots are using the updated component:
Screenshot 2024-03-20 at 5 27 25 PM
Screenshot 2024-03-20 at 5 27 32 PM

### DIFF
--- a/app/components/CapacityBar.tsx
+++ b/app/components/CapacityBar.tsx
@@ -6,16 +6,58 @@
  * Copyright Oxide Computer Company
  */
 
-import { Cpu16Icon, Ram16Icon, Ssd16Icon } from '@oxide/design-system/icons/react'
+import {
+  Cpu16Icon,
+  IpGlobal16Icon,
+  Ram16Icon,
+  Ssd16Icon,
+} from '@oxide/design-system/icons/react'
 
 import type { VirtualResourceCounts } from '~/api'
 import { classed } from '~/util/classed'
 import { splitDecimal } from '~/util/math'
 import { bytesToGiB, bytesToTiB } from '~/util/units'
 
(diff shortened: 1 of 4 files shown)


app/components/CapacityBar.tsx
```diff
@@ -6,75 +6,141 @@
...
+  kind: CapacityBarKind
+  provisioned: VirtualResourceCounts
+  allocated: VirtualResourceCounts
+  allocatedLabel: 'Quota' | 'Quota (Total)'
+}
+
+/**
+ * Shows a visual representation of the capacity and utilization of a resource.
+ * "provisioned" is the amount of the resource being used at this layer of the service
+ * "allocated" is the total amount of the resource available to this layer, as defined by the parent layer
```
charliepark, line 115: Also open to suggestions on this copy

charliepark:
There are enough other changes that'll be in play that I think it makes sense to close this PR for now.

PR #2087, opened 2024-03-21
Mobile and tablet layouts

benjaminleonard:
![Uploading CleanShot 2024-03-21 at 14.15.09.gif…](

Long overdue, but not as painful as I had expected. There are definitely refinements to be made to mobile layouts, but this is a good start. Feels a little magic to create an instance so easily and connect to it from your phone.

At some point we'll want to think about the actions that people are more likely to take on mobile devices and optimise for that. I'm thinking about dashboards, fault management, checking and increasing utilization – aka someone has pinged me about something and as an operator I need to check it out.

Some things are not disabled but unlikely to be usable on a mobile device. For example uploading an image – props to the first person to successfully upload a raw bootable image and boot from it.


benjaminleonard:
Priority for feedback on this PR will be ensuring we haven't regressed the desktop version.

app/components/EquivalentCliCommand.tsx
```diff
@@ -31,7 +31,12 @@ export function EquivalentCliCommand({ command }: { command: string }) {
 
   return (
     <>
-      <Button variant="ghost" size="sm" className="ml-2" onClick={() => setIsOpen(true)}>
+      <Button
+        variant="ghost"
+        size="sm"
+        className="md-:hidden"
```
benjaminleonard, line 37: Can be disabled universally I think. Users will not be using the CLI from their phones.

app/components/MswBanner.tsx
```diff
@@ -29,10 +29,18 @@ function ExternalLink({ href, children }: { href: string; children: ReactNode })
 export function MswBanner() {
   const [isOpen, setIsOpen] = useState(false)
   const closeModal = () => setIsOpen(false)
+
+  useEffect(() => {
+    document.body.classList.add('msw-banner')
```
benjaminleonard, line 34: Perhaps convoluted but only appears on the preview so not too concerned about it. We use it to add extra padding for the banner at the bottom. Banner has been moved to the bottom, mostly because it breaks less stuff in the navigation.

app/components/Sidebar.tsx
```diff
@@ -55,16 +67,113 @@ const JumpToButton = () => {
...
-    <div className="flex flex-col border-r text-sans-md text-default border-secondary">
-      <div className="mx-3 mt-4">
-        <JumpToButton />
-      </div>
-      {children}
+    <>
+      {transitions(
+        ({ x }, item) =>
+          item && (
+            <Dialog.Root
```
benjaminleonard, line 86: Feels slightly weird to be using a dialog here but it means we can use the same sidebar across both layouts. If there's a place that needs a bit more of a discerning eye it's probably here and the state logic to make sure it's bulletproof.

app/components/Sidebar.tsx
```diff
@@ -55,16 +67,113 @@ const JumpToButton = () => {
...
+                {pathname.split('/')[1] !== 'settings' && <ProfileLinks />}
+              </AnimatedDialogContent>
+            </Dialog.Root>
+          )
+      )}
+    </>
+  )
+}
+
+export const ProfileLinks = () => {
```
benjaminleonard, line 118: Adds profile info into the navigation, hidden on desktop because it's visible in the nav.

app/components/TopBar.tsx
```diff
@@ -104,6 +131,24 @@ export function TopBar({ children }: { children: React.ReactNode }) {
           </div>
         </div>
       </div>
-    </>
+    </div>
   )
 }
+
+const Menu12Icon = ({ className }: { className: string }) => (
```
benjaminleonard, line 138: Likewise, will also add to the design system

app/hooks/use-menu-state.ts
```diff
@@ -0,0 +1,40 @@
...
+  const { size } = useWindowSize()
+  const location = useLocation()
+
+  useEffect(() => {
+    closeSidebar()
+  }, [location.pathname])
+
+  return useStore(
+    (store) => ({
+      isOpen: size.width >= 1024 ? true : store.isOpen,
```
benjaminleonard, line 36: I tend to not like to target window sizes with javascript, but this felt like the easiest option. Forces the sidebar to be open on larger screen sizes since it's always visible.

app/layouts/helpers.tsx
```diff
@@ -14,20 +14,24 @@ import { useScrollRestoration } from '~/hooks/use-scroll-restoration'
 import { SkipLinkTarget } from '~/ui/lib/SkipLink'
 import { classed } from '~/util/classed'
 
-export const PageContainer = classed.div`grid h-screen grid-cols-[14.25rem,1fr] grid-rows-[60px,1fr]`
+export const PageContainer = classed.div`h-full pt-[var(--navigation-height)] [overscroll-behavior-y:none]`
```
benjaminleonard, line 17: Switched back to good old salt of the earth non-grid CSS. It was a great solution but we ran into some awkward browser behaviour where a forced scroll would mess up the layout irredeemably.

app/ui/styles/index.css
```diff
@@ -44,11 +44,34 @@
 
 :root {
   --content-gutter: 2.5rem;
+  --sidebar-width: 14.25rem;
+  --navigation-height: 3.75rem;
```
benjaminleonard, line 48: We should do more of this stuff I think

tailwind.config.js
```diff
@@ -28,19 +28,14 @@ module.exports = {
   theme: {
     extend: {
       screens: {
-        'xl-': { max: '1535px' },
-        'lg-': { max: '1279px' },
-        'md-': { max: '1023px' },
-        'sm-': { max: '767px' },
         'sm+': { min: '640px' },
```
benjaminleonard, line 31: I've never used this `-` / `+` for responsive Tailwind before and I must say I am a fan, means we're not forced to class them mobile first.

test/e2e/scroll-restore.e2e.ts
```diff
@@ -8,13 +8,13 @@
 import { expect, test, type Page } from './utils'
 
 async function expectScrollTop(page: Page, expected: number) {
-  const container = page.getByTestId('scroll-container')
+  const container = page.locator('#content_pane')
```
benjaminleonard, line 11: I was adding an ID to content pane already, so we no longer need the `test-id`

david-crespo:
Obviously we need to test it properly, but I just tried it on my phone and it feels incredible.

benjaminleonard:
Refactoring the dialog out on desktop, its causing issues with side modals.

test/e2e/scroll-restore.e2e.ts
```diff
@@ -5,30 +5,34 @@
...
   await page.goto('/projects/mock-project/disks')
+  // this helps us wait till the page is ready to scroll
+  await expectVisible(page, ['role=heading[name*="Disks"]'])
   await expectScrollTop(page, 0)
   await scrollTo(page, 143)
 
   // nav to snapshots
   await page.getByRole('link', { name: 'Snapshots' }).click()
+  await expectVisible(page, ['role=heading[name*="Snapshots"]'])
+  await expect(page).toHaveURL('/projects/mock-project/snapshots')
```
benjaminleonard, line 35: The scroll restore test had broken, if when changing it to use `window.scrollTo` and `window.scrollY`. The conclusion I came to was the page was not ready to scroll yet when the scroll was triggered. We didn't hit this before because we needed to wait for the scrollable element before continuing. My solution was to wait for something in the new screen to arrive before scrolling.

tailwind.config.js
```diff
@@ -28,19 +28,14 @@ module.exports = {
...
         'md+': { min: '768px' },
         'lg+': { min: '1024px' },
         'xl+': { min: '1280px' },
-        '2xl+': { min: '1536px' },
-        sm: { min: '640px', max: '767px' },
-        md: { min: '768px', max: '1023px' },
```
david-crespo, line 41: `md:` is used here. Couldn't find any of the others around:

https://github.com/oxidecomputer/console/blob/0114288974fb9e36bf392277fe59b0de357ec137/app/ui/styles/components/login-page.css#L9-L11

Without it:

image

david-crespo: Oh, it doesn't error because it still has a definition, the default one. That might not be the reason for the above bug then.

benjaminleonard: Fixed in latest commit – I was applying a gradient to the layout too soon.

benjaminleonard:
No idea why this test is suddenly failing and it passes on my local machine.

david-crespo:
Couple things. 

1. I'm not sure why it suddenly became reproducible locally after I merged main. Could be that I re-ran `npm install`
2. The failure here was simply that the element in _was_ very slightly in the viewport (not sure why that changed). The error message from playwright about that happens to be very opaque.
3. The fix is to scroll 5px farther.

image

app/layouts/helpers.tsx
```diff
@@ -14,20 +14,24 @@ import { useScrollRestoration } from '~/hooks/use-scroll-restoration'
...
+      id="content_pane"
+    >
+      <div className="flex flex-grow flex-col pb-8 md-:pb-16">
         <SkipLinkTarget />
         <main className="[&>*]:gutter">
           <Outlet />
         </main>
       </div>
-      <div className="sticky bottom-0 flex-shrink-0 justify-between overflow-hidden border-t bg-default border-secondary empty:border-t-0">
+      <div className="sticky bottom-0 z-popover flex-shrink-0 justify-between overflow-hidden border-t bg-default border-secondary empty:border-t-0">
```
david-crespo, line 34: this might conflict with actual popovers

benjaminleonard: image

image

This seems to be fine because all of the popovers are in portals and therefore appears later in the DOM. I _could_ add a `z-bottomBar` or something that is `z-index: 5;` (popover is currently 10).

benjaminleonard:
Would love to get this into release 8 – in part because it touches so many files the longer it sits the more painful any merge conflicts are going to be.

david-crespo:
This is beautiful work but because it's hard to say it's a priority and it has so many moving parts (and is getting more and more out of date), I think we'll want to break it into chunks (as far as that's possible) and redo each one as a new PR on top of main.

benjaminleonard:
The important part here is fixing the parts that break on smaller desktop sizes currently. And removing the grid layout is necessary I think for fixing the scrolling bug on the instance create (and other screens) – which is probably the most invasive change.

Happy for this to be split out, though I'd be doubtful that would be less work than spending a bit of time testing this. If we deployed this as early on the dogfood cycle before the next release, and did some extra checks on our end then the scope for harm is pretty low. Especially since our e2e coverage is pretty good.

david-crespo:
I merged main and it all seems to work, so we're in better shape than I expected. Here's one easy thing to fix — needs a max-w or a slightly smaller hard width.

image

app/components/TopBar.tsx
```diff
@@ -26,20 +34,53 @@ export function TopBar({ children }: { children: React.ReactNode }) {
...
 
-  // The height of this component is governed by the `PageContainer`
-  // It's important that this component returns two distinct elements (wrapped in a fragment).
-  // Each element will occupy one of the top column slots provided by `PageContainer`.
+  const { isOpen } = useMenuState()
+
   return (
-    <>
-      <div className="flex items-center border-b border-r px-3 border-secondary">
+    <div className="fixed top-0 z-topBar col-span-2 grid h-[var(--navigation-height)] w-full grid-cols-[min-content,auto] bg-default lg+:grid-cols-[var(--sidebar-width),auto]">
```
david-crespo, line 40: this stuff freaks me out. wonder if we can drop the grid cols thing

PR #2089, opened 2024-03-21
Access roles refactor suggestions

david-crespo:
My suggestions for #2062 were too complicated to put into comments. The resulting UI is the same except I called the tooltip "other roles" instead of figuring out which is in there.


app/pages/project/access/ProjectAccessPage.tsx
```diff
@@ -92,26 +92,23 @@ export function ProjectAccessPage() {
...
 
-        const roles = [siloRole, projectRole].filter(isTruthy)
+        const siloAccessRow = userAssignments.find((a) => a.roleSource === 'silo')
+        const projectAccessRow = userAssignments.find((a) => a.roleSource === 'project')
 
-        const { name, identityType } = userAssignments[0]
+        const roleBadges = sortBy(
+          [siloAccessRow, projectAccessRow].filter(isTruthy),
+          (r) => roleOrder[r.roleName] // sorts strongest role first
+        )
```
david-crespo, line 103: This filter and sort takes care of what the ternaries were doing.

PR #2091, opened 2024-03-22
Bundle splitting proof of concept

### LINKED ISSUE #2019: More aggressive bundle splitting
The main bundle is 300k gzipped and it's quite a lot of JS to parse and execute up front when maybe 20% of is needed for a given page.  React Router makes it extremely easy to split on a per-page basis. Remix SPA mode seems to split every page by default. We probably don't have to go that far, but there's not that much downside to it. Converting to the `lazy` format will get us very close to the file structure we'd need for Remix SPA, so we can consider this an investment in that transition as well.

https://reactrouter.com/en/main/route/lazy

david-crespo:
Closes #2019 

See https://reactrouter.com/en/main/route/lazy#multiple-routes-in-a-single-file

Details to be added.

### DIFF
--- a/app/pages/system/lazy.ts
+++ b/app/pages/system/lazy.ts
@@ -1,0 +1,16 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+
+type Page = keyof typeof import('./index.ts')
+
+export const lazySystemPage = (page: Page) => () =>
+  import('./index.ts').then((mod) => mod[page])
+
+/** for the weird little redirect routes where we only want the loader */
+export const lazySystemPageLoader = (page: Page) => () =>
+  import('./index.ts').then((mod) => ({ loader: mod[page].loader }))
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -36,7 +36,10 @@ module.exports = {
   rules: {
     '@typescript-eslint/array-type': 'off',
     '@typescript-eslint/consistent-type-definitions': 'off',
-    '@typescript-eslint/consistent-type-imports': ['error', { prefer: 'type-imports' }],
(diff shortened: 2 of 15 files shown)


app/pages/system/lazy.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+
+type Page = keyof typeof import('./index.ts')
```
david-crespo, line 9: didn't know you can do this, it rules

david-crespo:
We'll figure this out later.

david-crespo:
I pulled out more system pages into the system pages bundle and the results speak for themselves: it's a complete waste of time. Lmao. We're cutting barely more than 1% out of the main bundle.

### Before

```
dist/assets/Terminal-k-txfx7j.js                            287.56 kB │ gzip:  72.15 kB │ map:   590.21 kB
dist/assets/TimeSeriesChart-DNJqSGFX.js                     384.83 kB │ gzip: 107.41 kB │ map: 1,759.67 kB
dist/assets/app-pxHDd_km.js                               1,078.75 kB │ gzip: 337.50 kB │ map: 5,189.44 kB
```

### After

```
dist/assets/index-C3sI8Wdb.js                                30.82 kB │ gzip:   8.10 kB │ map:   107.52 kB
dist/assets/Terminal-Cz_nk_5N.js                            287.56 kB │ gzip:  72.15 kB │ map:   590.21 kB
dist/assets/TimeSeriesChart-DdtGqp6q.js                     384.83 kB │ gzip: 107.41 kB │ map: 1,759.67 kB
dist/assets/app-BILsoTVx.js                               1,050.90 kB │ gzip: 332.45 kB │ map: 5,086.27 kB
```

### Why so underwhelming

`npx vite-bundle-visualizer` tells the story: after pulling out xterm and recharts (which we already do) the entire main app.js bundle is taken up by large dependencies that are used in both system and silo pages, like the date picker related stuff. So our best bet is probably to try pulling those things in with dynamic imports so they get their own bundles.

image

PR #2097, opened 2024-03-26
Fix issue with sourceType overload on project vs silo images and disks

charliepark:
This proposal addresses an issue we're seeing in the instance create form.

Currently, because the create-instance form has a default image selected when the form is initialized, the form will use that image as the foundation for the instance, even if the user clicks to a different image source tab, if they don't select a new image. So they might select the "Project images" tab but end up creating an instance using the default Silo image. This is also a problem when the user selects the new "existing disk" tab.

This PR addresses the issue by setting up each of the tabs as their own piece of the form, and setting up a `bootDiskSourceType` that keeps track of the desired boot disk source.

This also allows us to maintain state across tab changes, so if the user has selected a source image and then clicks to a different tab, then clicks back, they don't lose their work.


app/forms/instance-create.tsx
```diff
@@ -363,14 +367,20 @@ export function CreateInstanceForm() {
...
           setValue(
             'bootDiskSourceType',
-            val === 'silo' || val === 'project' ? 'image' : 'disk'
+            val === 'silo' ? 'siloImage' : val === 'project' ? 'projectImage' : 'disk'
+          )
+          // if the user switches to the disk tab, clear the selected image;
+          // if they switch to the same tab that has the defaultImage, select it
+          setValue(
+            'bootDiskSource',
+            defaultSource === 'disk' ? '' : val === defaultSource ? defaultImage?.id : ''
```
charliepark, line 380: I can clean this up

app/forms/instance-create.tsx
```diff
@@ -363,14 +367,17 @@ export function CreateInstanceForm() {
...
         // default to the project images tab if there are only project images
-        defaultValue={
-          siloImages.length > 0 ? 'silo' : projectImages.length > 0 ? 'project' : 'disk'
-        }
+        defaultValue={defaultSource}
         onValueChange={(val) => {
           setValue(
             'bootDiskSourceType',
-            val === 'silo' || val === 'project' ? 'image' : 'disk'
+            val === 'silo' ? 'siloImage' : val === 'project' ? 'projectImage' : 'disk'
```
charliepark, line 374: This and line 168 have some overlap that can get cleaned up a bit, probably

david-crespo, line 374: This could just be `setValue('bootDiskSourceType', val)` if the strings were the same. Unfortunately there is no cute way to get TS to check that you're not messing it up somehow, but we're already lacking that, so it's not getting any worse.

app/forms/instance-create.tsx
```diff
@@ -450,7 +457,8 @@ export function CreateInstanceForm() {
         </Tabs.Content>
       </Tabs.Root>
 
-      {sourceType === 'image' && (
+      {/* siloImage and projectImage sourceTypes have additional fields */}
+      {sourceType.includes('Image') && (
```
david-crespo, line 461: a less evil approach would be defining chunk of JSX up around line 188

```
const bootDiskSizeAndName = (....)
```

and then inserting it directly into the two image tabs with `{bootDiskSizeAndName}` and leaving it out of the disk one.

app/forms/instance-create.tsx
david-crespo: Best I could do. Kind of horrible. Probably best to instead just make sure the e2es exercise all three tabs.

```diff
diff --git a/app/forms/instance-create.tsx b/app/forms/instance-create.tsx
index 7156e5ae..4c55aeee 100644
--- a/app/forms/instance-create.tsx
+++ b/app/forms/instance-create.tsx
@@ -60,6 +60,8 @@ import { links } from '~/util/links'
 import { pb } from '~/util/path-builder'
 import { GiB } from '~/util/units'
 
+type BootDiskTab = 'disk' | 'projectImage' | 'siloImage'
+
 export type InstanceCreateInput = Assign<
   // API accepts undefined but it's easier if we don't
   SetRequired<InstanceCreate, 'networkInterfaces'>,
@@ -68,7 +70,7 @@ export type InstanceCreateInput = Assign<
     disks: DiskTableItem[]
     bootDiskName: string
     bootDiskSize: number
-    bootDiskSourceType: 'disk' | 'projectImage' | 'siloImage'
+    bootDiskSourceType: BootDiskTab
     bootDiskSource: string
     userData: File | null
     // ssh keys are always specified. we do not need the undefined case
@@ -161,11 +163,11 @@ export function CreateInstanceForm() {
   const allKeys = useMemo(() => sshKeys.items.map((key) => key.id), [sshKeys])
 
   const defaultSource =
-    siloImages.length > 0 ? 'silo' : projectImages.length > 0 ? 'project' : 'disk'
+    siloImages.length > 0 ? 'siloImage' : projectImages.length > 0 ? 'projectImage' : 'disk'
 
   const defaultValues: InstanceCreateInput = {
     ...baseDefaultValues,
-    bootDiskSourceType: defaultSource === 'disk' ? 'disk' : `${defaultSource}Image`,
+    bootDiskSourceType: defaultSource,
     bootDiskSource: defaultImage?.id || '',
     sshPublicKeys: allKeys,
     // Use 2x the image size as the default boot disk size
@@ -369,10 +371,7 @@ export function CreateInstanceForm() {
         // default to the project images tab if there are only project images
         defaultValue={defaultSource}
         onValueChange={(val) => {
-          setValue(
-            'bootDiskSourceType',
-            val === 'silo' ? 'siloImage' : val === 'project' ? 'projectImage' : 'disk'
-          )
+          setValue('bootDiskSourceType', val as BootDiskTab)
           // if the user switches to the disk tab, clear the selected image;
           // if they switch to the same tab that has the defaultImage, select it
           setValue('bootDiskSource', val === defaultSource ? defaultImage?.id : '')
@@ -381,13 +380,16 @@ export function CreateInstanceForm() {
         }}
       >
         <Tabs.List aria-describedby="boot-disk">
-          <Tabs.Trigger value="silo" disabled={isSubmitting}>
+          <Tabs.Trigger value={'siloImage' satisfies BootDiskTab} disabled={isSubmitting}>
             Silo images
           </Tabs.Trigger>
-          <Tabs.Trigger value="project" disabled={isSubmitting}>
+          <Tabs.Trigger
+            value={'projectImage' satisfies BootDiskTab}
+            disabled={isSubmitting}
+          >
             Project images
           </Tabs.Trigger>
-          <Tabs.Trigger value="disk" disabled={isSubmitting}>
+          <Tabs.Trigger value={'disk' satisfies BootDiskTab} disabled={isSubmitting}>
             Existing disks
           </Tabs.Trigger>
         </Tabs.List>
@@ -398,7 +400,7 @@ export function CreateInstanceForm() {
             content="Images or disks are required to create or attach a boot disk."
           />
         )}
-        <Tabs.Content value="silo" className="space-y-4">
+        <Tabs.Content value={'siloImage' satisfies BootDiskTab} className="space-y-4">
           {siloImages.length === 0 ? (
             <div className="flex max-w-lg items-center justify-center rounded-lg border p-6 border-default">
               <EmptyMessage
@@ -415,7 +417,7 @@ export function CreateInstanceForm() {
             />
           )}
         </Tabs.Content>
-        <Tabs.Content value="project" className="space-y-4">
+        <Tabs.Content value={'projectImage' satisfies BootDiskTab} className="space-y-4">
           {projectImages.length === 0 ? (
             <div className="flex max-w-lg items-center justify-center rounded-lg border p-6 border-default">
               <EmptyMessage
@@ -435,7 +437,7 @@ export function CreateInstanceForm() {
           )}
         </Tabs.Content>
 
-        <Tabs.Content value="disk" className="space-y-4">
+        <Tabs.Content value={'disk' satisfies BootDiskTab} className="space-y-4">
           {disks.length === 0 ? (
             <div className="flex max-w-lg items-center justify-center rounded-lg border p-6 border-default">
               <EmptyMessage
```

charliepark: Eh, I liked it; folded it in. Working on the tests.

david-crespo:
This is looking really good. It fixes the bug. As we talked about in chat, I think it would be slightly better if your selection on each tab was preserved as you switch between tabs, so you could check what's on another tab without losing what you've done. This would require splitting `bootDiskSource` into a separate value for each tab and then tweaking what we do in `onSubmit` to construct the request body. I'm not sure if it makes validation harder, hopefully it just works.

test/e2e/instance-create.e2e.ts
```diff
@@ -229,3 +230,58 @@ test('shows object not found error on no default pool', async ({ page }) => {
...
+  await expect(page.getByText('arch-2022-06-01')).toBeVisible()
+  // change to a different tab
+  await page.getByRole('tab', { name: 'Existing disks' }).click()
+  // the image should no longer be visible
+  await expect(page.getByText('arch-2022-06-01')).toBeHidden()
+  // change back to the tab with the image
+  await page.getByRole('tab', { name: 'Silo images' }).click()
+  // arch should still be selected
+  await expect(page.getByText('arch-2022-06-01')).toBeVisible()
+})
```
david-crespo, line 287: These look great. Too bad there's not really a way to check after create that the instance got the right image.

app/forms/instance-create.tsx
```diff
@@ -68,8 +70,10 @@ export type InstanceCreateInput = Assign<
     disks: DiskTableItem[]
     bootDiskName: string
     bootDiskSize: number
-    bootDiskSourceType: 'disk' | 'image'
-    bootDiskSource: string
+    bootDiskSourceType: BootDiskSourceType
+    siloImageSource: string
+    projectImageSource: string
+    diskSource: string
```
david-crespo, line 76: I'd put newlines around these 4 lines and a comment above saying `type` is a switch picking between the other three

app/forms/instance-create.tsx
```diff
@@ -160,29 +166,88 @@ export function CreateInstanceForm() {
...
+        ? values.siloImageSource
+        : values.projectImageSource
+    return {
+      type: 'create' as const,
+      name: values.bootDiskName || genName(values.name, source),
+      description: `Created as a boot disk for ${values.name}`,
+      size: values.bootDiskSize * GiB,
+      diskSource: { type: 'image' as const, imageId: source },
+    }
+  }
```
david-crespo, line 249: Since as far as I can tell, this takes all relevant values as an argument, I would move it out of the component body. I would also rename it to `getBootDisk` or `getBootDiskAttachment` because `create` is easy to confuse with the `type: 'create' thing here. You can also put a return type of `InstanceDiskAttachment` on the function, more for clarity than anything else since it's checked regardless when you put it in the request body. It will also let you get rid of those `as const` bits, and gives slightly easier to understand type errors if we mess something up in here.

test/e2e/instance-create.e2e.ts
charliepark: Looks like we can verify that a substring from the original image's ID was used as part of the new disk's name.

david-crespo (approved):
sweet. let’s do final review in #2076

PR #2100, opened 2024-03-27
Maintain state of fields across tab navigation

charliepark:
In order to have the form retain the input values into the various tab's forms, I had to update a few fields in the form.

I'll want to take a look at this in the morning, as I suspect there are a few gnarly bits in it.

If this works as I believe it does, though, we'd merge it back into `fix-form-sourceType-bug`, which is, itself, waiting to get merged to `existing-boot-disk`. I was wary enough about the changes here that I didn't want to just Leeroy Jenkins it into `fix-form-sourceType-bug`.

### DIFF
--- a/app/forms/instance-create.tsx
+++ b/app/forms/instance-create.tsx
@@ -55,8 +55,8 @@ import { RadioCard } from '~/ui/lib/Radio'
 import { Tabs } from '~/ui/lib/Tabs'
 import { TextInputHint } from '~/ui/lib/TextInput'
 import { readBlobAsBase64 } from '~/util/file'
-import { invariant } from '~/util/invariant'
 import { links } from '~/util/links'
+import { nearest10 } from '~/util/math'
 import { pb } from '~/util/path-builder'
 import { GiB } from '~/util/units'
 
@@ -71,7 +71,6 @@ export type InstanceCreateInput = Assign<
     bootDiskName: string
     bootDiskSize: number
     bootDiskSourceType: BootDiskSourceType
-    bootDiskSource: string
     siloImageSource: string
     projectImageSource: string
     diskSource: string
@@ -97,7 +96,6 @@ const baseDefaultValues: InstanceCreateInput = {
   bootDiskSize: 10,
 
   bootDiskSourceType: 'siloImage',
-  bootDiskSource: '',
   siloImageSource: '',
   projectImageSource: '',
   diskSource: '',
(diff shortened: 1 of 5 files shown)


app/forms/instance-create.tsx
```diff
@@ -398,11 +429,15 @@ export function CreateInstanceForm() {
...
-          form.clearErrors('bootDiskSource')
+          if (val === 'siloImage') {
+            setValue('bootDiskSource', siloImageSource)
+          }
+          if (val === 'projectImage') {
+            setValue('bootDiskSource', projectImageSource)
+          }
+          if (val === 'disk') {
+            setValue('bootDiskSource', bootDiskSource)
+          }
```
david-crespo, line 440: Why do we need to keep the common `bootDiskSource` at all? Couldn't both the onSubmit boot disk logic and the size updater look at `siloImageSource`, etc. dynamically based on `bootDiskSourceType`?

app/forms/instance-create.tsx
```diff
@@ -177,10 +186,33 @@ export function CreateInstanceForm() {
...
+  const bootDiskSourceType = useWatch({ control: control, name: 'bootDiskSourceType' })
+  const bootDiskSource = useWatch({ control: control, name: 'bootDiskSource' })
+  const bootDiskSize = useWatch({ control: control, name: 'bootDiskSize' })
+  const image = allImages.find((i) => i.id === bootDiskSource)
   const imageSize = image?.size ? Math.ceil(image.size / GiB) : undefined
 
+  useEffect(() => {
+    bootDiskSourceType === 'siloImage' && setValue('bootDiskSource', siloImageSource)
+    bootDiskSourceType === 'projectImage' && setValue('bootDiskSource', projectImageSource)
+    bootDiskSourceType === 'disk' && setValue('bootDiskSource', diskSource)
```
david-crespo, line 201: these are redundant with the ones in `onValueChange`, but like I said there, I don't think either is necessary

app/forms/instance-create.tsx
```diff
@@ -177,10 +186,33 @@ export function CreateInstanceForm() {
...
+    if (imageSize && imageSize > bootDiskSize) {
+      console.log({ imageSize, bootDiskSize })
+      setValue('bootDiskSize', imageSize * 2)
+    }
```
david-crespo, line 205: This should be able to go in `onValueChange`, I think, then you avoid this whole `useEffect`

app/forms/instance-create.tsx
charliepark: There was an issue I was seeing where some cached state wasn't clearing when changing tabs and so the intended source wasn't getting updated, but I'll look into this a bit more.

david-crespo (approved):

PR #2102, opened 2024-03-27
Change all uses of RHF `<Controller>` to `useController`

david-crespo:
TIL `<Controller>` is just `useController` in a trenchcoat. Remove the trenchcoat. Render props are so 2016.

```ts
const Controller = <
  TFieldValues extends FieldValues = FieldValues,
  TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
>(
  props: ControllerProps<TFieldValues, TName>,
) => props.render(useController<TFieldValues, TName>(props));
```

https://github.com/react-hook-form/react-hook-form/blob/306627e385738e322cad8dff9e752678f8210020/src/controller.tsx#L46-L51


charliepark (approved):
nice

PR #2108, opened 2024-03-28
update firewall filter type/value visual treatment

### LINKED ISSUE #1932: Firewall filter alignment

charliepark:
In #1932 @augustuswm reported that the alignment and spacing was strange when multiple lines were present. @benjaminleonard noted that the table's design hadn't been fully realized according to the spec in Figma. This PR updates the elements of the filter column, as well as the design of the `TypeValueCell` component, to better match the design.

Current:
Screenshot 2024-03-28 at 11 02 03 AM

Updated:
Screenshot 2024-03-28 at 11 50 52 AM

Fixes #1932


charliepark:
Just noticed a capitalization change we'll need to make to conform to designs. Updating.

augustuswm:
This looks much more readable. In this version is each filter always on a new line? or will small filters attempt to collapse.

charliepark:
Great question. I had them on multiple lines, but as our table rows have some CSS constraining the row height, things get a bit crowded when multiple filters are in play. I've updated it so filters will wrap:
Screenshot 2024-03-28 at 4 40 03 PM

david-crespo (approved):
Clearly an improvement. At some point we should do one of those + popup things for overflow when there are too many.

One note, I think the letter spacing on the mono font is meant for upper case and is too wide for normal-cased text. We do the `!normal-case` override in so many places; it might be worth adding a style of `text-mono-md` that is meant to support lower case.

2024-04-02-lower-upper

It doesn't look that bad in the gif, but I think it looks worse in context.

image

Here it is with `!tracking-normal` on the badge, which shrinks `letter-spacing` from 0.04rem (from `text-mono-sm`) to 0. Worth considering in a followup.

image

PR #2110, opened 2024-03-29
Proof of concept: Take column logic out of QueryTable

### LINKED ISSUE #2107: Error on nav from IP pools to IP pool detail before utilization fetch completes

david-crespo:
Closes #2107
Supersedes #2109

In #2109 I had a big outstanding mystery:

> **Unsolved:** why does unmount happen before the loader is complete? Page transition should only happen after loader is done.

To investigate, I tried to reproduce the #2107 problem on StackBlitz with a very minimal setup:

* https://stackblitz.com/edit/vitejs-vite-l4jbyr?file=src%2Fmain.tsx&terminal=dev
* https://stackblitz.com/edit/vitejs-vite-59zkej?file=src%2Fmain.tsx&terminal=dev

But it did not have the issue, so I had to go back to our code and ask myself what we're doing to cause it. The breakthrough came when I replaced the IP pools table with this and the problem went away.

```tsx
<Link to={pb.ipPool({ pool: 'ip-pool-1' })}>ip-pool-1</Link>
<UtilizationCell pool="ip-pool-1" />
```

So most likely it was `QueryTable` causing the `UtilizationCell` to unmount and remount at a weird time. Then I remembered: `QueryTable` is full of shenanigans, it would totally do that. The culprit was my least favorite part of `QueryTable`:

https://github.com/oxidecomputer/console/blob/44c646f973ce9951c2819165582f0db0eb6abee0/app/table/QueryTable.tsx#L132-L167

Turning `QueryTable` inside out and getting rid of all the `any` bits has been a long-term goal of mine, but it is pretty hard to untangle. What hadn't occurred to me until now is that we can pretty easily stop doing the `columns` part altogether, passing in a native React Table `columns` array as a prop instead of doing the `<Column />` business.

So in this PR, I've introduced `useQueryTable2`, which is the same as `useQueryTable` but with the bad `columns` thing deleted.

I think we should convert all existing `useQueryTable` calls to `useQueryTable2`, then delete the original and rename `useQueryTable2` back to `useQueryTable`. There are 21 `useQueryTable` calls (not including the IP pools one I changed here), and some will be pretty involved, so we probably don't want to do them all in one PR.

### DIFF
--- a/app/pages/project/disks/DisksPage.tsx
+++ b/app/pages/project/disks/DisksPage.tsx
@@ -6,7 +6,7 @@
  * Copyright Oxide Computer Company
  */
 import { createColumnHelper } from '@tanstack/react-table'
-import { useCallback, useMemo } from 'react'
+import { useCallback } from 'react'
 import { Link, Outlet, type LoaderFunctionArgs } from 'react-router-dom'
 
 import {
@@ -26,7 +26,7 @@ import { DateCell } from '~/table/cells/DateCell'
 import { defaultCell } from '~/table/cells/DefaultCell'
 import { InstanceLinkCell } from '~/table/cells/InstanceLinkCell'
 import { SizeCell } from '~/table/cells/SizeCell'
-import { getActionsCol, type MenuAction } from '~/table/columns/action-col'
+import { useColsWithActions, type MenuAction } from '~/table/columns/action-col'
 import { useQueryTable2 } from '~/table/QueryTable2'
 import { buttonStyle } from '~/ui/lib/Button'
 import { EmptyMessage } from '~/ui/lib/EmptyMessage'
@@ -158,7 +158,7 @@ export function DisksPage() {
     [addToast, createSnapshot, deleteDisk, project]
   )
 
-  const columns = useMemo(() => [...staticCols, getActionsCol(makeActions)], [makeActions])
+  const columns = useColsWithActions(staticCols, makeActions)
 
   return (
     <>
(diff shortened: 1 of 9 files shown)


david-crespo:
The same test failed twice in a row in Safari, so that’s significant. Will take a look.

charliepark (approved):
Good sleuthing on this. No concerns I can come up with; stoked it's addressing the issue. We might extract some of the common columns from different tables into helper functions, though I also get it if the benefits there are minimal. Regardless, this is good.

david-crespo:
Thanks for the quick review. Definitely with you on cleaner helper functions. I did a bit of that with `defaultCell` and `makeLinkCell`.

david-crespo:
The test passed in Safari but now flaked in Firefox and in the same spot. Very interesting....

PR #2111, opened 2024-03-29
Convert more tables to `useQueryTable2`

david-crespo:
These diffs will be a lot nicer to read with whitespace changes hidden.

https://github.com/oxidecomputer/console/pull/2111/files?w=1

### DIFF
--- a/app/table/cells/DefaultCell.tsx
+++ b/app/table/cells/DefaultCell.tsx
@@ -13,6 +13,6 @@ export const DefaultCell = ({ value }: Cell<string>) => (
   <span className="text-secondary">{value}</span>
 )
 
-export const defaultCell = <T, U extends React.ReactNode>(props: CellContext<T, U>) => (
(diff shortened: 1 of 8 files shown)


app/table/cells/DefaultCell.tsx
```diff
@@ -5,14 +5,8 @@
...
-
 import type { Cell } from './Cell'
 
 export const DefaultCell = ({ value }: Cell<string>) => (
   <span className="text-secondary">{value}</span>
 )
-
-export const defaultCell = <T, U extends React.ReactNode>(props: CellContext<T, U>) => (
-  <span className="text-secondary">{props.getValue()}</span>
-)
```
david-crespo, line 18: Turns out we don't need this. Worth keeping an eye out for situations where it behaves differently because it's not in a span, but that seems pretty unlikely.

https://github.com/oxidecomputer/console/blob/9485ca2392ba7e4e5b65d370e4020a95ad46b1bd/app/ui/styles/components/table.css#L31-L36

PR #2112, opened 2024-03-29
Wrap all the `makeActions` functions in `useCallback`

david-crespo:
This is a very noisy but empty change, so we're pulling it out, so the rest of our `useQueryTable` conversions will not be as noisy as #2111.


charliepark (approved):
Thank you!

david-crespo:
Still flaking. It’s probably time to really figure this out, or if that takes too long, just change the test.

PR #2113, opened 2024-03-30
Reduce e2e flake

david-crespo:
Will only merge if any of it appears to work.


david-crespo:
Out of 4 runs, none failed outright, and fewer retries seemed to be required.

PR #2115, opened 2024-04-01
More table conversions

charliepark:
This PR finishes the QueryTable move from `~/table/QueryTable` to `~/table/QueryTable2` (though — not to be confusing — it also renames the file `~/table/QueryTable2` back to `~/table/QueryTable`.

All tests are passing and I clicked through to check that the tables are still rendering their data correctly. Will look for a few more optimizations / refactors in this, but I it's ready for review.


charliepark:
Oh, capitalization; fun times.

david-crespo:
Looks great pending the casing fixes. You should also be able to delete `DefaultCell` and `linkCell`. I think.

app/table/cells/DefaultPoolCell.tsx
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Copyright Oxide Computer Company
+ */
+import { Success12Icon } from '@oxide/design-system/icons/react'
+
+import { Badge } from '~/ui/lib/Badge'
+
+import type { Cell } from './Cell'
+
+export const DefaultPoolCell = ({ value }: Cell<boolean>) =>
```
david-crespo, line 14: Nitpick, but I think this `Cell` type is pointless. I want to get rid of it in a followup. Might as well not use it here.

```suggestion
export const DefaultPoolCell = ({ value }: { value: boolean }) =>
```

And a nitpick on the nitpick, since we're no longer constrained by `QueryTable` to have the prop named `value`, it might be nice to name it something more descriptive like `isDefault`.

app/table/cells/DefaultPoolCell.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+import { Success12Icon } from '@oxide/design-system/icons/react'
+
+import { Badge } from '~/ui/lib/Badge'
+
+import type { Cell } from './Cell'
```
david-crespo, line 12:
```suggestion

```

charliepark:
Good thoughts on removing Cell. Since I was updating the prop at the callsites, I just put it all into one commit.

app/pages/system/SiloImagesPage.tsx
```diff
@@ -52,12 +53,26 @@ SiloImagesPage.loader = async () => {
...
+    colHelper.accessor('name', {
+      cell: makeLinkCell((image) => pb.siloImageEdit({ image })),
+    }),
+    colHelper.accessor('description', {}),
+    colHelper.accessor('size', { cell: (info) => <SizeCell value={info.getValue()} /> }),
+    colHelper.accessor('timeCreated', {
+      header: 'Created',
+      cell: (info) => <DateCell value={info.getValue()} />,
+    }),
+  ]
```
david-crespo, line 74: In the interest of not doing things inside the render loop when we can avoid it, cases like these where all the columns are fully static should be moved outside the component. This is true here and for projects, sleds, and floating IPs.

app/pages/project/vpcs/VpcsPage.tsx
```diff
@@ -54,6 +55,17 @@ export function VpcsPage() {
...
+  const staticCols = [
+    colHelper.accessor('name', { cell: makeLinkCell((vpc) => pb.vpc({ project, vpc })) }),
+    colHelper.accessor('dnsName', { header: 'DNS name' }),
+    colHelper.accessor('description', {}),
+    colHelper.accessor('timeCreated', {
+      header: 'created',
+      cell: (info) => <DateCell value={info.getValue()} />,
+    }),
+  ]
+
```
david-crespo, line 68: This one, on the other hand, has a column that can't be defined outside of render because it relies on `project`. In this case, we need to memoize.

```ts
const columns = useMemo(() => [
  colHelper.accessor('name', { cell: makeLinkCell((vpc) => pb.vpc({ project, vpc }))}),
  colHelper.accessor('dnsName', { header: 'DNS name' }),
  colHelper.accessor('description', {}),
  colHelper.accessor('timeCreated', {
    header: 'created',
    cell: (info) => <DateCell value={info.getValue()} />,
  }),
  getActionsCol(makeActions)
], [project, makeActions])
```

There's no point using `useColsWithActions` here — all it really does is wrap that `useMemo`. We may find we're using `useColsWithActions` in so few places after this change that we should just get rid of it.

In theory we could move the four statically definable columns out of here and spread them in with `...staticCols`, but I don't think the misdirection is worth it. Because of the `useMemo` they are only ever defined once anyway.

app/pages/system/inventory/sled/SledInstancesTab.tsx
```diff
@@ -43,38 +44,49 @@ const makeActions = (): MenuAction[] => []
...
+    colHelper.accessor('timeCreated', {
+      header: 'created',
+      id: 'created',
+      cell: (info) => <DateCell value={info.getValue()} />,
+    }),
+    colHelper.accessor('timeModified', {
+      header: 'modified',
+      id: 'modified',
+      cell: (info) => <DateCell value={info.getValue()} />,
+    }),
```
david-crespo, line 86: The `id`s here can all be removed — if the accessor is a string key, RT will use that as the ID, and if it's a function it will use `header` as the ID.

https://tanstack.com/table/latest/docs/guide/column-defs#unique-column-ids

app/pages/ProjectsPage.tsx
```diff
@@ -45,11 +46,23 @@ ProjectsPage.loader = async () => {
   return null
 }
 
+const colHelper = createColumnHelper<Project>()
+const projectsPageStaticCols = [
```
david-crespo, line 50: Nitpick, but for future reference I think the long name is overkill in files where there's only one, which is nearly all of them. It's actually kind of nice for them all to have the same name; it's like `export const loader` in a remix app.

david-crespo:
We did it.

Issue #1469, opened 2023-05-09
Image upload hang (only in FF on macOS)


david-crespo:
@mkeeter got stuck here, nothing unexpected in browser console. So Something prevented this step from completing and moving on. Not much to go on but we need to keep an eye out for this happening again.

<img src="https://github.com/oxidecomputer/console/assets/3612203/be7a446e-e44e-4181-a955-24b4eadfe946" width="400">

This is the relevant section of code. The fact that the upload step remained in its loading state and we never triggered the next request suggests an error got thrown by the `pMap` that somehow resulted in a silent hang rather than an error.

https://github.com/oxidecomputer/console/blob/e9306b0b4498e9ab71c5e84391c63d527673f5f7/app/forms/image-upload.tsx#L408-L430

PR #2117, opened 2024-04-03
Timeouts on individual image upload chunk POSTs

david-crespo:
I forgot who suggested this, but it's such a good idea it seems obvious in hindsight. I'm not 100% sure whether we're still seeing the image upload hang from #1469 and #3559 — I think we are, occasionally. Having requests time out after some relatively long amount of time might mitigate the issue entirely.

The first commit here uses a timeout of 5 seconds and simulates a hang of 30 seconds in the mock API handler. With the timeout change commented out, the test fails. With the timeout change in, it passes.

This PR also improves something irrespective of hanging requests: before this PR, when the image upload is canceled, any outstanding chunk POSTs are allowed to finish. Here we pass that abort signal to every request, which means they get canceled immediately when the overall upload is canceled.

### DIFF
--- a/app/forms/image-upload.tsx
+++ b/app/forms/image-upload.tsx
@@ -212,12 +212,7 @@ export function CreateImageSideModalForm() {
 
   const createDisk = useApiMutation('diskCreate')
   const startImport = useApiMutation('diskBulkWriteImportStart')
-  const uploadChunk = useApiMutation(
-    'diskBulkWriteImport',
-    {},
-    // use both the general abort signal for the whole process and a per-request timeout
-    { signal: anySignal([AbortSignal.timeout(30000), abortController.current?.signal]) }
-  )
+  const uploadChunk = useApiMutation('diskBulkWriteImport')
 
   // synthetic state for upload step because it consists of multiple requests
   const [syntheticUploadState, setSyntheticUploadState] =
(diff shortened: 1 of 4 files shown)


app/forms/image-upload.tsx
```diff
@@ -211,7 +212,12 @@ export function CreateImageSideModalForm() {
 
   const createDisk = useApiMutation('diskCreate')
   const startImport = useApiMutation('diskBulkWriteImportStart')
-  const uploadChunk = useApiMutation('diskBulkWriteImport')
+  const uploadChunk = useApiMutation(
+    'diskBulkWriteImport',
+    {},
+    // use both the general abort signal for the whole process and a per-request timeout
+    { signal: anySignal([AbortSignal.timeout(30000), abortController.current?.signal]) }
+  )
```
david-crespo, line 220: TODO: While this seems to work, I'm pretty sure it should not, and it should be pretty easy to break with multiple hangs, because this is using the same timeout signal for all requests. Every mutate call should get its own.

app/forms/image-upload.tsx
```diff
@@ -392,7 +393,15 @@ export function CreateImageSideModalForm() {
       // all zeros. It turns out this happens a lot.
       if (!isAllZeros(base64EncodedData)) {
         await uploadChunk
-          .mutateAsync({ path, body: { offset, base64EncodedData } })
+          .mutateAsync({
+            path,
+            body: { offset, base64EncodedData },
+            // use both the abort signal for the whole upload and a per-request timeout
+            signal: anySignal([
+              AbortSignal.timeout(30000),
```
david-crespo, line 401: This could probably be shorter, I just wanted to be generous.

charliepark (approved):
This is clever. Nice.

PR #2142, opened 2024-04-12
use a content-security-policy in development

iliana:
In preparation for adding `content-security-policy` and other security headers to console endpoints in Nexus, we can add them here too to ensure they're applied when developing / running tests / etc.

The base headers are defined in `vercel.json` and then imported into `vite.config.ts` to avoid repeating them.

The `content-security-policy` is based on the recommendation by the [OWASP Secure Headers Project](https://owasp.org/www-project-secure-headers/index.html) (click the "Best Practices" tab). I am still tweaking this as I look at other sites for recommendations, but I wanted to see if this breaks any tests. The directives:
- `default-src 'self'`: By default, restrict all resources to same-origin.
- `style-src 'unsafe-inline' 'self'`: Restrict CSS to same-origin and inline use. We've fixed a number of these issues as part of this PR but we still have some, so we'll track those in #2183.
- `frame-src 'none'`: Disallow nested browsing contexts (`<frame>` and `<iframe>`).
- `object-src 'none'`: Disallow `<object>` and `<embed>`.
- `form-action 'none'`: Disallow submitting any forms with an `action` attribute (none of our forms appear to be the traditional kind and instead use events).
- `frame-ancestors 'none'`: Disallow embedding this site with things like `<iframe>`; used to prevent click-jacking attacks.

In development mode, an additional `script-src` CSP directive is added which references a randomly-generated nonce. [Vite injects this in the generated index.html](https://vitejs.dev/guide/features.html#content-security-policy-csp) so that the dev-mode scripts can load. We do this instead of allowing `'unsafe-inline'` because I'm not sure whether tests run against dev bits or not, and this helps get dev builds much closer to production.

Also set are `x-content-type-options: nosniff` and `x-frame-options: DENY`.

It would be nice to write tests (both vitest and playwright?) that verify this header is set in those environments so we know we're actually testing the CSP.


benjaminleonard:
We might not need the headers in `vercel.json` since all we're really using it for is the mock preview, though there's probably no harm in keeping them there.

david-crespo:
I was thinking the same thing. Will think about it.

iliana:
Yeah, it's certainly optional for us to add them to vercel.json, but I also figure why not (apart from the hot mess of importing the JSON file when loading the Vite config). My thinking was, if we see some regular usage out of that, that's more incidental testing on the CSP... but I don't know if we get bug reports about the console preview.

david-crespo:
Still looking through all the values but I think this is good. I don't see any errors, everything seems fine. Good idea using the vercel config as the single source of truth.

david-crespo:
One thing I probably want to do is find a way in the code itself to comment on each of the values and why they are what they are. It looks like we [can't](https://github.com/orgs/vercel/discussions/1485) convert vercel.json to .jsonc and we probably can't jam comments in there, so I might add it as a markdown thing in the docs dir.

david-crespo:
The vite doc says:

> Ensure that you replace the placeholder with a unique value for each request. This is important to prevent bypassing a resource's policy, which can otherwise be easily done.

I assume that's relevant in production — maybe we don't need to bother with this at all in local dev?

iliana:
Yes, that's primarily relevant for production. The reason I went ahead and randomized it for each config load is so that someone can't add a `<script nonce="fixed-dev-string">` and have it work in dev, but that's a bit silly (it wouldn't work on the Vercel preview or the real control plane). For instance doing it this way in dev mode means the nonce is fixed on each page load for the existence of that dev server.

The real purpose of `cspNonce` is to put a placeholder string there, which can be replaced and randomized by the actual web server on each request it serves. We don't need that because our prod builds don't have any inline scripts, they all come from `'self'` which is reasonably safe for our purposes.

image

(from https://csp-evaluator.withgoogle.com/)

But we could pretty easily have Nexus generate a nonce, if we want. I guess we wouldn't be able to apply the correct nonce in vercel.json though.

iliana:
I suppose how that would look is:
- We set `cspNonce` to always be something like `NexusPlaceholder`.
- We update the CSP to always have `script-src 'nonce-NexusPlaceholder'`, in server/preview/vercel.
- Over in Nexus we change the code to search and replace `NexusPlaceholder` in index.html with a random hex value and also put that in the CSP header on each request.

I haven't yet researched if that breaks any browser addons.

david-crespo:
Ok, I think the placeholder thing makes more sense than what we have now, but I suspect, as you say, `script-src 'self'` is sufficient and we can skip the nonce thing altogether both here and in omicron.

---

I'm thinking of adding more explanation to the doc of why we're doing this client-side. How close is this? Do you think there is much of a direct security problem we're solving? The only attack vector I can think of is a malicious npm dependency.

> The reason we are setting CSP headers not just in Nexus, where they affect production deployments, but also in local dev (which Playwright runs against) and in the Vercel previews is not primarily the security of those environments but rather so we can know as early as possible in the development process if a given CSP directive breaks the web console.

iliana:
I don't believe we're solving a security problem directly by putting the CSP here, so your writeup is correct to me. The primary reason for starting here is to catch issues early (and before we turn it on in Nexus, where clickjacking/XSS/etc are far more of a potential threat).

Given that the Playwright tests run against the dev server, I definitely want to keep the nonce setup in non-production mode; the alternative is that we must allow `script-src 'unsafe-inline'` and that would mean we're not testing that the console works _without_ that.

david-crespo:
This is looking very good. In my attempt at a final run-through, I found this error when opening any popover, like the action menu on a table row or the system/silo picker. I will look into this later.

>  Content-Security-Policy: The page’s settings blocked an inline style (style-src-elem) from being applied because it violates the following directive: “default-src 'self'”

iliana:
Looks like that might be https://github.com/radix-ui/primitives/issues/2057, which they fixed by adding a nonce prop :( Might have to actually do the nonce thing in Nexus, or otherwise figure out how to add the CSS they're using into our build and filter out the dynamically created style element (see https://github.com/radix-ui/primitives/pull/2728/files).

iliana:
Okay there are two problems I spotted while trying to find causes of `cssText` in the bundle. The first is pretty annoying but can use a nonce as a workaround; the second can be fixed with an upstream patch.

1: **react-style-singleton** [creates a style tag with arbitrary CSS strings](https://github.com/theKashey/react-style-singleton/blob/32ae0eb5a3af83ee52f205ffef0e7418abb9be43/src/singleton.ts).

```
$ npm why react-style-singleton
react-style-singleton@2.2.1
node_modules/react-style-singleton
  react-style-singleton@"^2.2.1" from react-remove-scroll@2.5.5
  node_modules/react-remove-scroll
    react-remove-scroll@"2.5.5" from @radix-ui/react-dialog@1.0.5
    node_modules/@radix-ui/react-dialog
      @radix-ui/react-dialog@"^1.0.5" from the root project
    react-remove-scroll@"2.5.5" from @radix-ui/react-menu@2.0.6
    node_modules/@radix-ui/react-menu
      @radix-ui/react-menu@"2.0.6" from @radix-ui/react-dropdown-menu@2.0.6
      node_modules/@radix-ui/react-dropdown-menu
        @radix-ui/react-dropdown-menu@"^2.0.6" from the root project
  react-style-singleton@"^2.2.1" from react-remove-scroll-bar@2.3.4
  node_modules/react-remove-scroll-bar
    react-remove-scroll-bar@"^2.3.3" from react-remove-scroll@2.5.5
    node_modules/react-remove-scroll
      react-remove-scroll@"2.5.5" from @radix-ui/react-dialog@1.0.5
      node_modules/@radix-ui/react-dialog
        @radix-ui/react-dialog@"^1.0.5" from the root project
      react-remove-scroll@"2.5.5" from @radix-ui/react-menu@2.0.6
      node_modules/@radix-ui/react-menu
        @radix-ui/react-menu@"2.0.6" from @radix-ui/react-dropdown-menu@2.0.6
        node_modules/@radix-ui/react-dropdown-menu
          @radix-ui/react-dropdown-menu@"^2.0.6" from the root project
```

This uses yet another package by the same author, [get-nonce](https://github.com/theKashey/get-nonce), to look for the global `__webpack_nonce__` and add it to the `<style>` elements it creates.

2: This particular line in **@radix-ui/react-focus-guards**: https://github.com/radix-ui/primitives/blob/b32a93318cdfce383c2eec095710d35ffbd33a1c/packages/react/focus-guards/src/FocusGuards.tsx#L35

This can be fixed with a pretty simple patch which I'll send upstream.

iliana:
I wrote a patch to remove the code from **react-remove-scroll** that creates a global stylesheet.

The path we're actually hitting goes through **react-remove-scroll-bar**. This code is enabled when `<RemoveScroll removeScrollBar={true}>`, which is the default. This generates [a global stylesheet](https://github.com/theKashey/react-remove-scroll-bar/blob/v2.3.4/src/component.tsx#L18-L67) that, allegedly, removes the scroll bar from the browser view. In my use (Firefox on macOS with "Show scroll bars" set to "Always") this doesn't actually do anything. The non-`body` rules in that stylesheet are so you can do other things downstream from that package, which neither radix-ui nor we do.

Instead of adapting **react-remove-scroll-bar** to apply styles directly to the body element, I decided that instead this is fragile behavior that probably just doesn't need to run in the first place.

Another path, when `<RemoveScroll inert={true}>` is used (false by default), adds two global classes that are used to enable/disable `pointer-events` on the body and other relevant elements. radix-ui does not use this. I removed this so we could remove **react-style-singleton** from the bundle in its entirety.

The other patch handles the `cssText` usage I spotted in **@radix-ui/react-focus-guards**. I don't know if this actually triggers on the console but this is a simple patch, which I've already sent upstream.

iliana:
... wow, Vercel's node_modules caching doesn't appear to behave nicely with patch-package?

package.json
```diff
@@ -11,6 +11,7 @@
     "start:msw": "API_MODE=msw vite",
     "start:nexus": "API_MODE=nexus vite",
     "start:dogfood": "API_MODE=dogfood vite",
+    "preview": "API_MODE=msw npm run build && cp mockServiceWorker.js dist/ && vite preview",
```
iliana, line 14: Added this because I added more content to an existing patch in a commit, which made Vercel break. Turns out it caches your `node_modules` at the end of the run, and `patch-package` only knows not to apply a patch if it's entirely applied.

https://github.com/ds300/patch-package/issues/37 has some related chatter.

vercel.json
```diff
@@ -1,6 +1,19 @@
 {
-  "buildCommand": "API_MODE=msw npm run build && cp mockServiceWorker.js dist/",
+  "buildCommand": "API_MODE=msw npm run build && cp mockServiceWorker.js dist/ && npx patch-package --reverse",
```
iliana, line 2: Added this because I added more content to an existing patch in a commit, which made Vercel break. Turns out it caches your `node_modules` at the end of the run, and `patch-package` only knows not to apply a patch if it's entirely applied.

https://github.com/ds300/patch-package/issues/37 has some related chatter.

iliana:
> I removed this so we could remove react-style-singleton from the bundle in its entirety.

Also, a much simpler patch to **react-remove-scroll** should be possible and have the same effect, without removing the offending code from our bundle, if a smaller patch is preferred:

```diff
diff --git a/node_modules/react-remove-scroll/dist/es2015/UI.js b/node_modules/react-remove-scroll/dist/es2015/UI.js
index 26c94a8..5991c96 100644
--- a/node_modules/react-remove-scroll/dist/es2015/UI.js
+++ b/node_modules/react-remove-scroll/dist/es2015/UI.js
@@ -26,7 +26,7 @@ var RemoveScroll = React.forwardRef(function (props, parentRef) {
 });
 RemoveScroll.defaultProps = {
     enabled: true,
-    removeScrollBar: true,
+    removeScrollBar: false,
     inert: false,
 };
 RemoveScroll.classNames = {
```

iliana:
I just remembered that I need to verify that WebSockets (serial console) still work...

iliana:
The console websocket works fine with this CSP on Firefox. To-spec source matching was apparently wrong in WebKit (Safari) until two years ago. Not sure what Safari versions we consider supported? https://bugs.webkit.org/show_bug.cgi?id=235873

Unfortunately there appear to be some inline styles in xterm.js that I'll need to hunt down. I didn't realize that was in a separate bundle.
image

iliana:
If we want this to land in R8, we should add 'unsafe-inline' back to style-src, I think (and do the same over on the Omicron PR). Getting xterm.js to play nicely will take a bit more time than the patches above did.

david-crespo:
I think that makes sense. Everything else is a big improvement on the status quo.

iliana:
Cool. I'll get that done today so we can land it!

david-crespo:
Think it's worth getting rid of the patches and redoing them later? Keeps things cleaner here.

iliana:
Personally I'd like to keep them so we can see if/when they break with upstream changes? I don't think the behavior we're patching out is something we really want anyway. Unless the reason they're pulled in (e.g. radix-ui) are being dropped at some point.

david-crespo:
Yeah, that's fine. We'll see conflicts if we bump the dep and it conflicts with changes they've made.

david-crespo (approved):
Thank you for doing this! Putting it on auto-merge and I'll go give a quick last look to the Omicron PR.

PR #2145, opened 2024-04-12
Inline identicon

david-crespo:
Console is the only app using @oxide/identicon. @iliana made a [PR](https://github.com/oxidecomputer/identicon/pull/4) over there, but there's no point figuring out how to push that to npm. This PR copies that in here and changes it to render svg directly instead of producing a string that we dangerouslySetHtml with.

Confirmed the resulting SVG is the same as before:

image


benjaminleonard:
Great, that seems neater too.

iliana (approved):

PR #2148, opened 2024-04-15
Session expiration popup

### LINKED ISSUE #1616: Notify user if the authentication situation has changed
Our SAML setup on the dogfood rack has a very low authorization time (like an hour?), and I often find myself poking at something on the console only to have the next form submission whisk me away to a login page.

There's also a potential situation where in another tab you log into another account that might not have the same access you did before.

It would be nice if the console were polling something periodically to determine if the currently logged-in user has changed, and notify the user in some way. (I do not like AWS's version of this where it takes over the console and forces you to reload; I would like to be able to re-log in from another tab and come back to this page and continue whatever it was I was doing. Bonus points if I can middle-click a button or link to open the login page in a new tab.)

david-crespo:
Closes #1616 

- [x] Put modal state in zustand store
- [x] Display modal when we get a 401
- ~~Figure out how to _not_ close other modals when we show this one~~
- [x] Convert to toast-like thing with its own `ToastStack`-like thing
- [x] Integrate into the main toast stack so it doesn't overlap other toasts
- [ ] Make sure nothing weird happens if we get multiple 401s in parallel
- [ ] See what happens on subsequent attempts to nav
  - [ ] See if we can disable that without getting too weird — maybe telling RQ we're offline ([docs](https://tanstack.com/query/v5/docs/reference/onlineManager))

image

### Followup work

A followup that more fully nails what #1616 wants would be to ping periodically or on tab focus (which RQ does by default, and which we are doing by default) so that the user sees they're been logged out immediately and doesn't e.g., start filling out a form or whatever, only to find out on submit it was all for nothing.

Another followup would to open sign in in a new window and figuring out how to poll from the original window for a restored session. The cookie is HttpOnly, so we can't check it directly, but we can do two things. The obvious thing is to try to hit the API. I'm not sure we want to ping the API a million times though. I think we can be clever and either use `postMessage` to have the child window ping back to the parent to tell it there was a successful login, or use some kind of flag in localStorage to indicate session status to tell the parent window when it would make sense to try hitting the API again to see if we're back.


app/api/hooks.ts
```diff
@@ -49,14 +49,10 @@ const handleResult =
...
-    // Exception: 401 on password login POST needs to be handled in-page
+    // if logged out, show session expired popup and prompt to log in again
+    // Exception: 401 on password login POST needs to be handled in-form
     if (result.response.status === 401 && method !== 'loginLocal') {
-      // TODO-usability: for background requests, a redirect to login without
-      // warning could come as a surprise to the user, especially because
-      // sometimes background requests are not directly triggered by a user
-      // action, e.g., polling or refetching when window regains focus
-      navToLogin({ includeCurrent: true })
+      setSessionExpired()
```
david-crespo, line 55: If we also do this here

```ts
onlineManager.setOnline(false)
```

https://tanstack.com/query/v5/docs/framework/react/guides/network-mode
https://tanstack.com/query/latest/docs/reference/onlineManager

that does work to prevent subsequent requests from going through, and therefore to prevent subsequent navs that require an API call, but it does that by making any queries and mutations sit in pending forever. Submitting a form starts a spinner that never stops, and trying to nav kicks of a loading bar that never completes. This is a bizarre experience.

Maybe this is an argument for putting some kind of modal background on the toast that prevents you from clicking anything else on screen. But that would prevent you from scrolling the page (unless we got creative and allowed scroll but not click — a very strange thing to do), which would be useful if you had filled out a longer form like instance create or firewall rule and wanted to note down what you had done.

PR #2149, opened 2024-04-15
Headless UI combobox

david-crespo:
This is a very early WIP. I haven't even extracted it into its own component yet; I'm just messing around inline in the disk attach modal. Fortunately, our listbox is already built on a Headless component, and listbox is very similar in style to combobox, so we can pull quite a lot from our existing ([quite involved](https://github.com/oxidecomputer/console/blob/73263a4e805cdf5757d5441b962a453bc1531c87/app/ui/lib/Listbox.tsx)) component.

### DIFF
--- a/app/forms/disk-attach.tsx
+++ b/app/forms/disk-attach.tsx
@@ -5,7 +5,12 @@
  *
  * Copyright Oxide Computer Company
  */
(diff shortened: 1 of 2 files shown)


david-crespo:
Closing, @charliepark will redo this on top of #2230

PR #2151, opened 2024-04-16
Add state and policy columns to physical disks table

### LINKED ISSUE #2124: Show more data about physical disks
In https://github.com/oxidecomputer/omicron/issues/5153, the API added `state` and `policy` fields to physical disks. We should probably show at least state in the table, if not also policy. `timeModified` might also be useful if it reflects when the state last changed. I also think could stand to add a detail view in a side modal that you get to by clicking the ID cell, and which would include more or less all the fields in a properties table.

image

charliepark:
Fixes #2124 

This PR adds two columns to the table listing physical disks:
Screenshot 2024-04-15 at 8 56 23 PM


david-crespo:
Nice. Most likely we don’t want those red because it’s not really an error state. It’s something that happens on purpose. Could do neutral or whatever the gray one is.

charliepark:
With the neutral badges:
Screenshot 2024-04-16 at 10 12 20 AM

david-crespo:
Yeah that looks better. I think the policy column should be first because from [Sean's comment](https://github.com/oxidecomputer/console/issues/2124#issuecomment-2035358192), it sounds like policy drives the state, in the sense that you set the policy to expunged and wait for the state to change to decommissioned.

Another idea, but I doubt it can be made to feel right, is that we signal that the expunged/active state is temporary with a yellow badge color, but it's kind of weird because the _text_ of the state is the same — `active` — so it would be stupid to signal it with a different color. I would say don't bother getting clever with this. We can mull it over and see if we come up with a cute approach. I think the issue is that while there are two fields here, there are really three states (as Sean's comment indicates) and the states are a composite of the two values. So I picture something like this:

* In service + active: green
* Expunged + active: yellow
* Expunged + decommissioned: gray 

But is that one column? What do you even call the column? `Policy/state`? Do you have two badges? One badge with a slash in it? Blech. Just some thoughts. I don't think any of this works, maybe @paryhin can make sense of it.

charliepark:
Ah, great point on the column ordering (or at least, the attribute ordering, if we move to one column). Do we know what the time delta is on expunging → decommissioned? We aren't live-updating the page right now, but I'm wondering how often they'd end up in an expunged + active situation.

charliepark:
Proper order:
Screenshot 2024-04-16 at 11 39 18 AM

david-crespo (approved):
Looks good, add these columns to the test here and it's good to go:

https://github.com/oxidecomputer/console/blob/73263a4e805cdf5757d5441b962a453bc1531c87/test/e2e/inventory.e2e.ts#L38-L50

test/e2e/inventory.e2e.ts
```diff
@@ -47,4 +47,10 @@ test('Disk inventory page', async ({ page }) => {
   const table = page.getByRole('table')
   await expectRowVisible(table, { id: physicalDisks[0].id, 'Form factor': 'U.2' })
   await expectRowVisible(table, { id: physicalDisks[3].id, 'Form factor': 'M.2' })
+  await expectRowVisible(table, { id: physicalDisks[3].id, policy: 'in service' })
+  await expectRowVisible(table, { id: physicalDisks[3].id, state: 'active' })
+  await expectRowVisible(table, { id: physicalDisks[4].id, policy: 'expunged' })
+  await expectRowVisible(table, { id: physicalDisks[4].id, state: 'active' })
+  await expectRowVisible(table, { id: physicalDisks[5].id, policy: 'expunged' })
+  await expectRowVisible(table, { id: physicalDisks[5].id, state: 'decommissioned' })
```
david-crespo, line 55: These can be done in one line each:

```ts
await expectRowVisible(table, { id: physicalDisks[0].id, 'Form factor': 'U.2' })
await expectRowVisible(table, { id: physicalDisks[3].id, 'Form factor': 'M.2', policy: 'in service', state: 'active' })
await expectRowVisible(table, { id: physicalDisks[4].id, policy: 'expunged', state: 'active' })
await expectRowVisible(table, { id: physicalDisks[5].id, policy: 'expunged', state: 'decommissioned' })
```

david-crespo (approved):

PR #2153, opened 2024-04-16
Convert date cell to single line

charliepark:
We're working to tighten up the data tables (vertically), so are moving the date cell from two lines to one line.

Currently, it looks like this:
Screenshot 2024-04-16 at 12 07 04 PM

With this PR, it now looks like this:
Screenshot 2024-04-16 at 12 01 11 PM

On smaller screens, the date wraps:
Screenshot 2024-04-16 at 12 48 21 PM

### DIFF
--- a/app/components/TimeAgo.tsx
+++ b/app/components/TimeAgo.tsx
@@ -6,10 +6,9 @@
  * Copyright Oxide Computer Company
  */
 import type { Placement } from '@floating-ui/react'
-import { format } from 'date-fns'
 
 import { Tooltip } from '~/ui/lib/Tooltip'
-import { timeAgoAbbr } from '~/util/date'
+import { timeAgoAbbr, toLocaleDateString } from '~/util/date'
 
 export const TimeAgo = ({
   datetime,
@@ -23,7 +22,7 @@ export const TimeAgo = ({
   const content = (
     <div className="flex flex-col">
       <span className="text-tertiary">{tooltipText}</span>
-      <span>{format(datetime, 'MMM d, yyyy p')}</span>
+      <span>{toLocaleDateString(datetime)}</span>
     </div>
   )
   return (
--- a/app/components/TimeAgo.tsx
+++ b/app/components/TimeAgo.tsx
@@ -8,7 +8,7 @@
 import type { Placement } from '@floating-ui/react'
 
 import { Tooltip } from '~/ui/lib/Tooltip'
-import { timeAgoAbbr, toLocaleDateString } from '~/util/date'
+import { timeAgoAbbr, toLocaleDateTimeString } from '~/util/date'
 
 export const TimeAgo = ({
   datetime,
@@ -22,7 +22,7 @@ export const TimeAgo = ({
   const content = (
     <div className="flex flex-col">
       <span className="text-tertiary">{tooltipText}</span>
-      <span>{toLocaleDateString(datetime)}</span>
+      <span>{toLocaleDateTimeString(datetime)}</span>
     </div>
   )
   return (
--- a/app/table/columns/common.tsx
+++ b/app/table/columns/common.tsx
@@ -6,9 +6,9 @@
  * Copyright Oxide Computer Company
  */
 
-import { format } from 'date-fns/format'
 import { filesize } from 'filesize'
 
+import { DateTime } from '~/ui/lib/DateTime'
 import { Truncate } from '~/ui/lib/Truncate'
 
 import { EmptyCell } from '../cells/EmptyCell'
@@ -18,13 +18,7 @@ import { EmptyCell } from '../cells/EmptyCell'
 type Info<T> = { getValue: () => T }
 
 function dateCell(info: Info<Date>) {
-  const date = info.getValue()
-  return (
-    <time dateTime={date.toISOString()} className="flex flex-wrap gap-x-2">
-      <span>{format(date, 'PP')}</span>
-      <span className="text-quaternary">{format(date, 'p')}</span>
-    </time>
-  )
+  return <DateTime date={info.getValue()} />
 }
 
 function sizeCell(info: Info<number>) {
--- a/app/table/columns/common.tsx
+++ b/app/table/columns/common.tsx
@@ -6,7 +6,6 @@
  * Copyright Oxide Computer Company
  */
 
-import { formatISO } from 'date-fns'
 import { format } from 'date-fns/format'
 import { filesize } from 'filesize'
 
@@ -21,7 +20,7 @@ type Info<T> = { getValue: () => T }
 function dateCell(info: Info<Date>) {
   const date = info.getValue()
   return (
-    <time dateTime={formatISO(date)} className="flex flex-wrap gap-x-2">
+    <time dateTime={date.toISOString()} className="flex flex-wrap gap-x-2">
       <span>{format(date, 'PP')}</span>
       <span className="text-quaternary">{format(date, 'p')}</span>
     </time>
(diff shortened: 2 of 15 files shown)


app/table/columns/common.tsx
```diff
@@ -6,21 +6,26 @@
...
 
 // the full type of the info arg is CellContext<Row, Item> from RT, but in these
 // cells we only care about the return value of getValue
 type Info<T> = { getValue: () => T }
 
 function dateCell(info: Info<Date>) {
   const date = info.getValue()
-  return <TwoLineCell value={[format(date, 'MMM d, yyyy'), format(date, 'p')]} />
+  return (
+    <time dateTime={formatISO(date)} className="flex flex-wrap gap-x-2">
```
david-crespo, line 24: `date.toISOString()` should cover this. Not sure why they bothered reimplementing the whole thing themselves. I guess it takes options that let you do date only or time only. 

https://github.com/date-fns/date-fns/blob/5c1adb5369805ff552737bf8017dbe07f559b0c6/src/formatISO/index.ts#L47

charliepark: Oh, good point. Merp.

app/table/columns/common.tsx
```diff
@@ -12,15 +12,19 @@ import { filesize } from 'filesize'
...
 
 function dateCell(info: Info<Date>) {
   const date = info.getValue()
-  return <TwoLineCell value={[format(date, 'MMM d, yyyy'), format(date, 'p')]} />
+  return (
+    <time dateTime={date.toISOString()} className="flex flex-wrap gap-x-2">
+      <span>{format(date, 'PP')}</span>
+      <span className="text-quaternary">{format(date, 'p')}</span>
+    </time>
+  )
```
david-crespo, line 27: I'm realizing `date-fns` is very literal and doesn't seem to do any locale-based formatting. I think we should use `toLocaleString()` and get decent results, i.e., the same as we're currently getting in English but better for other languages. Something like this:

```ts
const dateStr = date.toLocaleString(undefined, { month: 'short', day: 'numeric', year: 'numeric' })
const timeStr = date.toLocaleString(undefined, { hour: 'numeric', minute: 'numeric' }))
```

image

In the math function we use `navigator.language` where I've put `undefined` here, but it seems like `undefined` is really the right way. The only problem is that the way I've tested this is to mock `navigator.language`, so if we use `undefined` instead that approach won't work in the tests. Maybe we should be using playwright instead to test the locale rendering. They make it really easy:

```ts
import { test, expect } from '@playwright/test';

test.describe('french language block', () => {

  test.use({ locale: 'fr-FR' });

  test('example', async ({ page }) => {
    // ...
  });
});
```

https://playwright.dev/docs/test-use-options#configuration-scopes

To summarize, I think you should:

1. Replace these `date-fns` `format` calls with `toLocaleString(undefined, ...)`
2. Check a displayed date in exactly one table in `de-DE` or whatever using the Playwright local setting feature
3. Don't worry about math, I'll make an issue to switch those to use `undefined` instead of `navigator.language` and change the way we test it

app/components/TimeAgo.tsx
```diff
@@ -23,7 +22,7 @@ export const TimeAgo = ({
   const content = (
     <div className="flex flex-col">
       <span className="text-tertiary">{tooltipText}</span>
-      <span>{format(datetime, 'MMM d, yyyy p')}</span>
+      <span>{toLocaleDateString(datetime)}</span>
```
david-crespo, line 25: This should include the time too. You might need another helper, or maybe `toLocaleString()` by itself would do it?

### `main`

Screenshot 2024-04-16 at 7 08 51 PM

### This branch

Screenshot 2024-04-16 at 7 08 46 PM

app/ui/lib/DateTime.tsx
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+import { toLocaleDateString, toLocaleTimeString } from '~/util/date'
+
+export const DateTime = ({ date, locale }: { date: Date; locale?: string }) => (
+  <time dateTime={date.toISOString()} className="flex flex-wrap gap-x-2">
+    <span>{toLocaleDateString(date, locale)}</span>
+    <span className="text-quaternary">{toLocaleTimeString(date, locale)}</span>
+  </time>
+)
```
david-crespo, line 16: stylish

app/util/date.spec.ts
```diff
@@ -54,4 +54,20 @@ describe('timeAgoAbbr', () => {
...
+    expect(toLocaleDateString(baseDate, 'fr-FR')).toEqual('7 juin 2021')
+    expect(toLocaleDateString(baseDate, 'de-DE')).toEqual('7. Juni 2021')
+    expect(toLocaleDateString(baseDate, 'ja-JP')).toEqual('2021年6月7日')
+  })
+
+  it('formats toLocaleTimeString', () => {
+    expect(toLocaleTimeString(baseDate)).toEqual('12:00 AM')
+    expect(toLocaleTimeString(baseDate, 'en-US')).toEqual('12:00 AM')
+    expect(toLocaleTimeString(baseDate, 'fr-FR')).toEqual('0:00 AM')
+    expect(toLocaleTimeString(baseDate, 'de-DE')).toEqual('0:00 AM')
```
david-crespo, line 70: I don't think these should include AM. I think if you take out `hour12: true`, the `en-US` one will still have it but these won't.

charliepark: Oh, subtle. Nice.

charliepark:
Changing my system language to French worked correctly:
Capture d’écran 2024-04-16 à 20 10 38

And back to English:
Screenshot 2024-04-16 at 8 34 17 PM

app/components/TimeAgo.tsx
charliepark: The native formatting adds a comma after the year (for `en-US`), but I think it's fine
Screenshot 2024-04-16 at 8 40 20 PM

david-crespo (approved):
Looks great! Tests should be simple, I think just looking at one page where there is a date should be fine.

app/util/date.ts
```diff
@@ -47,4 +43,22 @@ export const timeAgoAbbr = (d: Date, options?: FormatDistanceToNowStrictOptions)
...
+const dateOptions = {
+  month: 'short',
+  day: 'numeric',
+  year: 'numeric',
+} satisfies Intl.DateTimeFormatOptions
+
+const timeOptions = {
+  hour: 'numeric',
+  minute: 'numeric',
+} satisfies Intl.DateTimeFormatOptions
```
david-crespo, line 55: Heh, check this out: I was just reading the MDN page and noticed the style shortcuts section: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#datestyle

I think we can reduce these to `timeStyle: 'short'` and `dateStyle: 'medium'`! That is clean.

```ts
const dateOptions = { dateStyle: 'medium' } satisfies Intl.DateTimeFormatOptions
const timeOptions = { timeStyle: 'short' } satisfies Intl.DateTimeFormatOptions
```

image

Oddly it does change one of the results, the French one. Seems fine to me!

image

charliepark: Oh, interesting. Funny that it changes the German syntax. I don't have strong feelings on it, but agree the shorter configuration syntax is nice. Will get that in.

charliepark: Eh, I maybe prefer the output of the earlier version, but am fine either way, and recognize this will likely be an edge case for some time. We can always change it back if we decide that's better.
Screenshot 2024-04-16 at 9 32 25 PM

david-crespo: Even though it looks worse (maybe) I like the idea of taking the details out of our hands — in theory they should know better than we do what’s idiomatic in each locale.

charliepark: Makes sense; I agree with that.

PR #2154, opened 2024-04-16
Get rid of single tab on IP pools

david-crespo:
Trying harder to be normal.

image


app/routes.tsx
```diff
@@ -178,10 +177,11 @@ export const routes = createRoutesFromElements(
         </Route>
         <Route path="health" element={null} handle={{ crumb: 'Health' }} />
         <Route path="update" element={null} handle={{ crumb: 'Update' }} />
-        <Route path="networking" element={<NetworkingPage />}>
+        <Route path="networking">
+          <Route index element={<Navigate to="ip-pools" replace />} />
```
david-crespo, line 181: Just in case, but no one should ever land on plain `/system/networking` unless they're manually messing with the URL.

PR #2155, opened 2024-04-17
Use explicit optional locale arg in math functions, rework tests accordingly

david-crespo:
This is basically a refactor. It should not change any functionality, though it does add a test we didn't have before.

While discussing changes to date functions in https://github.com/oxidecomputer/console/pull/2153#discussion_r1567920202, I learned that the more natural way to tell the built-in locale-aware functions you want to use whatever the browser's current locale is while also passing some options is to pass `undefined` as the locale instead of `navigator.language` (which should still work but is a bit weirder-feeling). The nice thing about passing undefined is it goes really well with an optional `locale` arg that defaults to `undefined`, and which we only explicitly use in the tests.

### DIFF
--- a/app/util/math.spec.ts
+++ b/app/util/math.spec.ts
@@ -5,16 +5,9 @@
  *
  * Copyright Oxide Computer Company
  */
-import { afterAll, beforeAll, describe, expect, it } from 'vitest'
+import { describe, expect, it } from 'vitest'
 
-import {
-  displayBigNum,
-  nearest10,
-  percentage,
-  round,
-  splitDecimal,
-  toEngNotation,
-} from './math'
+import { displayBigNum, nearest10, percentage, round, splitDecimal } from './math'
 import { GiB } from './units'
 
 function roundTest() {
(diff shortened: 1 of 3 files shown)


app/util/math.ts
```diff
@@ -58,9 +58,7 @@ export function round(num: number, digits: number) {
   return Number(nf.format(num))
 }
 
-// a separate function because I wanted to test it with a bunch of locales
-// to make sure the toLowerCase thing is ok
-export const toEngNotation = (num: number | bigint, locale = navigator.language) =>
+const toEngNotation = (num: number | bigint, locale?: string) =>
```
david-crespo, line 61: Don't have to test this directly anymore because displayBigNum takes a locale arg too, so we don't have to export it at all.

test/e2e/ip-pools.e2e.ts
```diff
@@ -33,6 +33,24 @@ test('IP pool list', async ({ page }) => {
   })
 })
 
+test.describe('german locale', () => {
+  test.use({ locale: 'de-DE' })
```
david-crespo, line 37: love that playwright!

app/util/math.spec.ts
```diff
@@ -150,17 +134,14 @@ describe('with de-DE locale', () => {
...
 
-  // rounding must work the same irrespective of locale
-  it('round', roundTest)
-
   it.each([
     [0n, ['0', false]],
     [1n, ['1', false]],
     [155n, ['155', false]],
-    [999999n, ['999,999', false]],
+    [999999n, ['999.999', false]],
```
david-crespo, line 144: This was sort of a bug, but I'm pretty sure it was only due to the way we're mocking this for the tests. I think in a browser it would have worked correctly.

PR #2158, opened 2024-04-17
Set table cells to have shorter default height

charliepark:
This PR sets the default table row height to the `h-12` height, rather than the larger `h-16`.

So we move from a table looking like this:
Screenshot 2024-04-17 at 1 30 27 PM

To something more compact, like this:
Screenshot 2024-04-17 at 1 30 55 PM

37.5% more data on the screen*!

There are still some tables that require a larger height, and we can pass in a `rowHeight` prop to enable that. Here we have the `InstancesPage` table, which uses the `useQueryTable` hook, which invokes the `app/table/Table.tsx` component. With this wrapper component, we can pass in `rowHeight` as a prop to the overall `Table` component and the component passes it through to the `Table.Cell` in `ui/lib/Table.tsx`.
Screenshot 2024-04-17 at 1 29 49 PM

We also sometimes use the `ui/lib/Table.tsx` file directly, as in the `UtilizationPage` table. In cases like that, we have to pass the `height` prop directly to each `Table.Cell`, as the `Table` wrapper component doesn't pass anything directly along to its children table cells.
Screenshot 2024-04-17 at 1 29 57 PM

I'd prefer, I think, for us to not be involved at all in the height of these elements, and to instead just pass the data in and let the table cells dynamically update, with the appropriate cell-padding. The table would then flex to handle whatever the tallest content in the cells is. At the moment, though, the way we're doing borders on the cells would be a little fussy to get sorted, and there are a few things on our plate for V8. Happy to revisit that down the road.

### DIFF
--- a/app/table/QueryTable.tsx
+++ b/app/table/QueryTable.tsx
@@ -73,7 +73,7 @@ const makeQueryTable = <Item extends Record<string, unknown>>(
     debug,
     pagination = 'page',
     pageSize = PAGE_SIZE,
-    rowHeight,
+    rowHeight = 'small',
     emptyState,
     columns,
   }: QueryTableProps<Item>) {
--- a/app/table/QueryTable.tsx
+++ b/app/table/QueryTable.tsx
@@ -56,7 +56,7 @@ type QueryTableProps<Item> = {
   /** Function that produces a list of actions from a row item */
   pagination?: 'inline' | 'page'
   pageSize?: number
-  rowHeight?: 'small' | 'large' | 'auto'
+  rowHeight?: 'small' | 'large'
   emptyState: React.ReactElement
   columns: ColumnDef<Item, any>[]
 }
--- a/app/table/QueryTable.tsx
+++ b/app/table/QueryTable.tsx
@@ -56,6 +56,7 @@ type QueryTableProps<Item> = {
   /** Function that produces a list of actions from a row item */
   pagination?: 'inline' | 'page'
   pageSize?: number
+  rowHeight?: 'small' | 'large' | 'auto'
   emptyState: React.ReactElement
   columns: ColumnDef<Item, any>[]
 }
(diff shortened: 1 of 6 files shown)


david-crespo:
This
Is
The
Best

david-crespo:
I think the problem with letting the table rows decide their own height based on the content is that you might not get the same height for all the rows. In practice maybe that's a small concern because we decide what all the cells are, and we would decide how tall they are. But it feels a little risky.

charliepark:
We could put a `height` on the cells to force a minimum height, but if you had something that was wildly larger, the row would flex to handle it as in [this codepen example](https://codepen.io/charlie-at-oxide/pen/gOyBgZG) (also different border colors and interior colors, for funsies)

app/table/QueryTable.tsx
```diff
@@ -56,6 +56,7 @@ type QueryTableProps<Item> = {
   /** Function that produces a list of actions from a row item */
   pagination?: 'inline' | 'page'
   pageSize?: number
+  rowHeight?: 'small' | 'large' | 'auto'
```
david-crespo, line 59: I think you can remove `auto` as an option here and in the other one and nothing would change.

david-crespo:
IP pools might need to be a tall one for now until we redo the treatment of the utilization.

image

david-crespo:
Well, on second thought that’s pretty silly since you can’t even have IPv6 ranges right now. I’ll take a look at tweaking that to fit on one line. Even if it’s mediocre, the fact is no one will ever see it.

david-crespo:
Talked to @benjaminleonard and we decided to do nothing with the IP pool cells since the two line cell will never show up in production and it's not total shit (my words, not his).

david-crespo:
Was about to say let's go but we have a problem on firewall rules. Dynamic height?

image

david-crespo (approved):
Firewall rules is non-ideal, but because we're leaving it tall, it's not any worse than it currently is, so that's fine. I'm testing out using `ListPlusCell` for the filters, that can be a followup.

PR #2161, opened 2024-04-17
Animate refresh button

### LINKED ISSUE #1995: Instances: Instances list refresh feedback
We can reuse the same pattern from the utilization and capacity page.
Do we also want to reuse the auto-refresh dropdown? Or just the spinner on refresh.

david-crespo:
Closes #1995 

Imagine my delight discovering that `invalidateQueries` returns a promise that completes when the actual query finishes refetching! That makes it extremely easy to turn on a spinner when it starts and turn it off when it ends. How great is that!

https://github.com/oxidecomputer/console/assets/3612203/49481520-22ef-4334-bd15-452c2d9bf7a6

### DIFF
--- a/app/components/RefreshButton.tsx
+++ b/app/components/RefreshButton.tsx
@@ -16,17 +16,14 @@ import { SpinnerLoader } from '~/ui/lib/Spinner'
 export function RefreshButton({ onClick }: { onClick: () => Promise<void> }) {
   const [refreshing, setRefreshing] = useState(false)
 
+  async function refresh() {
+    setRefreshing(true)
+    await onClick()
+    setRefreshing(false)
+  }
+
   return (
-    <Button
-      size="icon"
-      variant="ghost"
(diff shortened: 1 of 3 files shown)


charliepark (approved):
Oh, great! This looks good.

In an ideal world, I think we'd have a "done" state, like we have with the "copy the external IP" checkmark that fades, so you have some indication that it has completed successfully. I guess the timestamps on the page do change, so you get that, but if your focus is on the spinner, it'd be cool if it also gave you the confirmation that the action had resolved.

But that can wait for a future PR. Looks good.

david-crespo:
Yes! I was thinking about a tooltip on hover that says when it was last run. Kind of noisy? I don’t know.

charliepark:
I was playing with an animating checkmark (really just cribbing CopyToClipboard); will open a new PR and we can see if it works or not.

PR #2163, opened 2024-04-18
`ListPlusCell` for firewall rule filters

### LINKED ISSUE #2118: Firewall rules table: use + popup for overflow when there are too many targets/filters
Right now they stack and can overflow the cell.

david-crespo:
Just a fun idea to close #2118.

- [x] Better mock data
- [x] Basic concept
- [ ] Also use for targets
- [ ] Final polish

image


david-crespo:
This looks good and I think it works pretty well, almost certainly better than overflowing. The only problem I have with it is that it makes the table pretty wide to show two rules inline.

image

One way around that is to only show at most one inline, though that's kind of crappy, especially if there are exactly two.

image

I think my ideal setup would be to leave the table as a tall table and encourage these two wrap so we could fit at least 2 or 3 in there before having to overflowing into the popup.

david-crespo:
As discussed in chat, in a followup I will experiment with putting the whole list in the popover, not just the "rest".

PR #2164, opened 2024-04-18
Make messages closable, and persist closure to localStorage

charliepark:
This adds a new prop to the Message component that, if passed, makes the Message box hideable. This can be used for info boxes that explain concepts of the API, or UI elements within Console. When the user hides the box, the box's key is saved to localStorage, so subsequent reloads of the page won't render it.

Here's an example of the info box being shown for the Floating IPs page. It's hidden, then a subsequent reload shows that the box isn't rendered:
closable


app/ui/lib/Message.tsx
```diff
@@ -73,18 +76,24 @@ export const Message = ({
   variant = 'success',
   cta,
   icon,
+  hideableKey,
 }: MessageProps) => {
+  const [hidden, setHidden] = useState(false)
+  if (hidden || (hideableKey && hideableKey in localStorage)) {
```
david-crespo, line 82: I think we should nest these all under one key maybe? makes them much easier to nuke and stuff

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -188,9 +188,23 @@ export function FloatingIpsPage() {
...
+                <ExternalLink
+                  href={links.floatingIpsDocs}
+                  className="text-info-secondary hover:text-info"
+                >
+                  managing floating IPs
+                </ExternalLink>
+                .
+              </>
+            }
+          />
```
david-crespo, line 207: move this out of TableControle so it goes across the whole page, that's what @paryhin had in the design

charliepark:
Screenshot 2024-04-18 at 12 00 27 PM

Screenshot 2024-04-18 at 12 00 38 PM

charliepark:
I think a full-width box also makes the `X` a bit harder to spot.

charliepark:
Eh, there's a weird interplay between a few of the elements and negative margins that I need to resolve.

david-crespo:
Remaining negative margin issue on IP pool linked silos.

https://console-git-make-messages-dismissible-oxidecomputer.vercel.app/system/networking/ip-pools/ip-pool-1?tab=silos

image

test/e2e/hideableMessages.e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+  await page.getByLabel('Close information box').click()
+
+  // Go back to the Floating IPs page and make sure the box is still hidden
+  await page.goto('/projects/mock-project/floating-ips')
+  await expect(page.getByLabel('Close information box')).toBeHidden()
+
+  // Go back to the IP Pools Linked Silos page and make sure the box is still hidden
+  await page.goto('/system/networking/ip-pools/ip-pool-1?tab=silos')
+  await expect(page.getByLabel('Close information box')).toBeHidden()
+})
```
david-crespo, line 27: neat

app/ui/lib/Message.tsx
```diff
@@ -73,22 +76,29 @@ export const Message = ({
   variant = 'success',
   cta,
   icon,
+  hideableKey,
 }: MessageProps) => {
+  const [hidden, setHidden] = useState(false)
+  const closedMessages = localStorage.getItem('closedMessages')
```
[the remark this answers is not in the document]
david-crespo: There's kind of a lot going on here in the implementation, but the calling API here is really nice. If we plan on doing more with localStorage it might be worth getting this or similar in place.

https://usehooks-ts.com/react-hook/use-local-storage#hook

david-crespo:
The work looks good here, but I would vote for waiting a few days and doing the contextual docs links in a button like Eugene is working on. I feel somehow both that these are more obtrusive that what we have now when they're there, but then once you close them they're impossible to find again.

I think I agree with Ben and Eugene that it would be nice if we could avoid these chunks of text altogether. I think good links to the would be a fine replacement.

charliepark:
We're going to hold off on this for now, in favor of a button directing them to the docs. Will revisit this problem / solution thesis down the road.

PR #2166, opened 2024-04-18
Small table cell refactor

charliepark:
Our tables are constructed by dropping a `div` inside each `td` and then putting the contents of the cell into that div. I tried simplifying things by moving the borders to the cells, but it proved to be trickier than I thought it would be, and I wasn't confident that I wouldn't introduce edge case CSS bugs. So I've mostly left it as it was.

One change that I did make was to only apply the height class to the contained `<div>`, rather than to the `<div>` and the `<td>` that wraps it. There was no reason I could find to apply the class to both, as the `<td>` will occupy the proper size based on the child element (the `<div>`). I tried to see if there were any unexpected rendering issues with the class in only one place, but did not find anything. If @benjaminleonard or @david-crespo has any reason they can recall as to why we were applying the `height` to both the `<div>` _and_ the `<td>`, I'd love to hear it! But with the logic only needed for one place, we're able to simplify things a little bit.

Visually, the tables all still look the same.

Screenshot 2024-04-18 at 4 12 58 PM

I also found one spot where we should have the taller row height, so I embiggened the VPC Subnets table cells.

Screenshot 2024-04-18 at 3 55 56 PM

### DIFF
--- a/app/pages/project/vpcs/VpcPage/tabs/VpcSubnetsTab.tsx
+++ b/app/pages/project/vpcs/VpcPage/tabs/VpcSubnetsTab.tsx
@@ -83,7 +83,7 @@ export const VpcSubnetsTab = () => {
         {creating && <CreateSubnetForm onDismiss={() => setCreating(false)} />}
         {editing && <EditSubnetForm editing={editing} onDismiss={() => setEditing(null)} />}
       </div>
-      <Table columns={columns} emptyState={emptyState} />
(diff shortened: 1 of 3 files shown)


david-crespo:
The double height class was added in https://github.com/oxidecomputer/console/pull/579 but I don’t see any particular reason for it. Definitely better to get rid of it if possible!

PR #2168, opened 2024-04-19
Add plus icon to all "new" buttons above tables

charliepark:
This PR updates the buttons (and links, where appropriate) that sit just above tables in the UI, adding a plus icon (technically, a _roundel_) to the buttons.
Screenshot 2024-04-18 at 10 04 39 PM

Screenshot 2024-04-18 at 10 11 00 PM

I'm not totally sure that the icon makes sense when paired with the text reading "Promote image", but figured that consistency across the buttons was more important than the semantic quibbling on that one button. That being said, I'm open to input.

### DIFF
--- a/app/pages/ProjectsPage.tsx
+++ b/app/pages/ProjectsPage.tsx
@@ -7,7 +7,7 @@
  */
 import { createColumnHelper } from '@tanstack/react-table'
 import { useCallback, useMemo } from 'react'
-import { Link, Outlet, useNavigate } from 'react-router-dom'
+import { Outlet, useNavigate } from 'react-router-dom'
 
 import {
   apiQueryClient,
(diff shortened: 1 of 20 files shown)


david-crespo:
Nice. I think I lean against putting the plus on promote image. I'm surprised that every other one is a create action! I thought there would be more weird ones around. Since it really is a create button I'd make the name less generic to reflect that, maybe simply `CreateButton` and `CreateLink`.

There might be a couple other buttons here and there that need the plus, like `Add SSH key` on instance create. On the other hand it feels weird to put it on create and attach disk buttons, so maybe it's better to stay away from that entirely on the form.

david-crespo:
Looking great. Last thing: what about these? I feel like all the options are weird. If it's on create only, why is it not on attach? If it's on both, it looks kind of weird and busy. Maybe neither is the least attention-grabbing.

image

benjaminleonard:
Can we go for `text-accent-secondary` on these icons. Otherwise I think they end up a little too weighty compared to the text.

david-crespo:
Better.

image

PR #2170, opened 2024-04-19
make all external IPs in header copyable

### LINKED ISSUE #2171: Add copy button to each external IP
Also use `ListPlusCell` (or some version of it) to handle more IPs than fit in the box.

charliepark:
Fixes #2171 

We currently only allow for a "copy to clipboard" function when there's a single external IP in the list in the header. This PR makes it so that any IP in the list is copyable.
Screenshot 2024-04-19 at 12 07 26 PM

We should add a dropdown, a la ListPlusCell, for when there are more than … 3(? 2?) IPs, but that can be a subsequent PR.

### DIFF
--- a/app/components/ExternalIps.tsx
+++ b/app/components/ExternalIps.tsx
@@ -22,30 +22,18 @@ export function ExternalIps({ project, instance }: InstanceSelector) {
   if (isPending) return <SkeletonCell />
 
   const ips = data?.items
-    ? intersperse(
-        data.items.map((eip) => <IpLink ip={eip.ip} key={eip.ip} />),
+  if (!ips || ips.length === 0) return <EmptyCell />
+  return (
+    <div className="flex items-center gap-1">
+      {intersperse(
+        ips.map((eip) => (
+          <span className="flex items-center" key={eip.ip}>
+            {eip.ip}
+            <CopyToClipboard text={eip.ip} />
+          </span>
+        )),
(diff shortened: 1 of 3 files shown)


david-crespo:
I think we could stand to keep the IP as a link. It doesn't really cost us anything if the copy button is also there.

charliepark:
re-linked: Screenshot 2024-04-19 at 12 37 59 PM

app/components/ExternalIps.tsx
```diff
@@ -22,30 +22,25 @@ export function ExternalIps({ project, instance }: InstanceSelector) {
...
+          <span className="flex items-center" key={eip.ip}>
+            <a
+              className="link-with-underline text-sans-semi-md"
+              href={`https://${eip.ip}`}
+              target="_blank"
+              rel="noreferrer"
+            >
+              {eip.ip}
+            </a>
+            <CopyToClipboard text={eip.ip} />
```
david-crespo, line 39: I think this needs an `ml-1` to do what the `gap-1` was doing when this was at top level. I was afraid it would then look too close to the slash on the right, but it seems fine to me.

### before

image


### after

image

david-crespo (approved):
looks good to me pending an `ml-1`

app/components/ExternalIps.tsx
benjaminleonard: Yep +1

PR #2181, opened 2024-04-22
Add properties table to IP Pool page

### LINKED ISSUE #2177: Add mini table to IP pools page

charliepark:
Fixes #2177 

This PR adds a properties table to the top of the IP Pool page.
Screenshot 2024-04-22 at 12 41 26 PM

It also makes a slight tweak to the half-width capacity bar, to make sure that it aligns with the properties table.

The code differs from Figma in that the original design lists the linked silos in the bottom-right property slot. Because the linked silos are so available, though (a bit further down the page), @david-crespo and I decided to put something else in the bottom-right slot. I used the existing pattern of "last modified", but there might be a more useful one.

Eugene suggested possibly using "is this the default IP Pool for this Silo?"

### DIFF
--- a/app/pages/system/networking/IpPoolPage.tsx
+++ b/app/pages/system/networking/IpPoolPage.tsx
@@ -120,9 +120,9 @@ function UtilizationBars() {
 
   return (
     <>
-      <h2 className="mb-3 pl-3 text-mono-xs text-secondary">Capacity available</h2>
-      <div className="mb-8 flex min-w-min flex-col gap-3 lg+:flex-row">
-        {ipv4.capacity > 0 ? (
+      <h2 className="mb-3 pl-3 text-mono-xs text-secondary">Utilization</h2>
+      <div className="mb-8 flex min-w-min flex-col gap-4 lg+:flex-row">
+        {ipv4.capacity > 0 && (
           <CapacityBar
             icon={<IpGlobal16Icon />}
             title="IPv4"
@@ -133,11 +133,8 @@ function UtilizationBars() {
             unit="IPs"
             includeUnit={false}
           />
-        ) : (
-          // necessary for aligning capacity bar with Properties Table
-          <div className="w-full"></div>
         )}
-        {ipv6.capacity > 0 ? (
+        {ipv6.capacity > 0 && (
           <CapacityBar
             icon={<IpGlobal16Icon />}
             title="IPv6"
@@ -148,10 +145,10 @@ function UtilizationBars() {
             unit="IPs"
             includeUnit={false}
           />
-        ) : (
-          // necessary for aligning capacity bar with Properties Table
-          <div className="w-full"></div>
         )}
+        {/* necessary for aligning capacity bar with Properties Table */}
+        {((ipv4.capacity > 0 && ipv6.capacity === 0n) ||
+          (ipv4.capacity === 0 && ipv6.capacity > 0n)) && <div className="w-full"></div>}
       </div>
     </>
   )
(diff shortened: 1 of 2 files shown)


app/pages/system/networking/IpPoolPage.tsx
```diff
@@ -101,32 +119,41 @@ function UtilizationBars() {
...
+            capacity={ipv4.capacity}
+            provisionedLabel="Allocated"
+            capacityLabel="Capacity"
+            unit="IPs"
+            includeUnit={false}
+          />
+        ) : (
+          // necessary for aligning capacity bar with Properties Table
+          <div className="w-full"></div>
+        )}
```
david-crespo, line 139: This has a very funny result when there are only IPv6:

image

charliepark: Ah, good callout.

charliepark: Screenshot 2024-04-22 at 4 14 35 PM

charliepark:
@david-crespo and I talked through a cleaner way to handle the half-width charts, with a more precise `max-width` with `calc()`. Here's how it looks on wide and narrow screens:
Screenshot 2024-04-22 at 6 37 51 PM

Screenshot 2024-04-22 at 6 38 00 PM

app/pages/system/networking/IpPoolPage.tsx
```diff
@@ -101,32 +119,35 @@ function UtilizationBars() {
...
-          capacity={ipv6.capacity}
-          provisionedLabel="Allocated"
-          capacityLabel="Capacity"
-          unit="IPs"
-          includeUnit={false}
-        />
-      )}
-    </div>
+    <>
+      <h2 className="mb-3 pl-3 text-mono-xs text-secondary">Utilization</h2>
```
david-crespo, line 123: This heading is teensy tiny. I think it should be `text-mono-sm`. We don't use `xs` in many spots.

app/pages/system/networking/IpPoolPage.tsx
```diff
@@ -75,6 +77,22 @@ export function IpPoolPage() {
       <PageHeader>
         <PageTitle icon={<Networking24Icon />}>{pool.name}</PageTitle>
       </PageHeader>
+      <PropertiesTable.Group className="mb-12">
+        <PropertiesTable>
+          <PropertiesTable.Row label="Description">
+            {pool.description || <EmptyCell />}
+          </PropertiesTable.Row>
+          <PropertiesTable.Row label="ID">{pool.id}</PropertiesTable.Row>
+        </PropertiesTable>
```
david-crespo, line 86: Silo, the most similar existing one of these, puts ID first, and I think that looks better when the description is empty or short.

image

david-crespo:
I'm ambivalent about this change overall, even though it's consistent with other pages — I think the data in the table is rarely going to be useful and it adds noise and pushes everything interesting down. To me it points to the need for a different pattern for this resource metadata, maybe a collapsible one.

image

benjaminleonard:
cc @paryhin – could you chime in when you can?

Collapsible properties table is interesting but it has significant consequences across the entire console that we shouldn't take lightly. Though I'm less bothered about the page being pushed down, utilisation and the properties both seem like things you might check more frequently than the included ranges and scrolling is no great crime.

I do find the half width utilisation pretty unsatisfying – and even when we have IPv6 in there, the sheer size of the range makes the bar meaningless unless the user is using a crazy number of them. So as nice as this component is, I think it might benefit us to try something tabular instead.

We might also consider a metrics or utilisation tab, that shows not only current but historical utilisation. Is that useful? I don't know. There might be something in knowing that an IP pool is being exhausted more quickly at this point in time.

Not that it's a blocker to this PR, we can follow up.

david-crespo:
I do think historical utilization is useful, but we don't have it API side — it would have been way more work to add than the point-in-time data and I wanted to wait and see if anyone really wants it.

charliepark:
There are enough issues with this one, and it's been open long enough, I'm going to close it, and if we feel it was solving a problem that's still worth solving, we can revisit it.

PR #2184, opened 2024-04-23
Use OxQL for system metrics

### LINKED ISSUE #2122: Try `/v1/timeseries/query` endpoint
https://github.com/oxidecomputer/omicron/pull/5273 added some really powerful endpoints for pulling data. If we can get our graphs working with those, we can get rid of some of my least favorite API endpoints.

david-crespo:
Closes #2122

Pretty early draft, just wanted a break from other work. Tested against dogfood and it works.

- [x] Fetch relevant data
- [ ] Render data in the charts and remove existing calls


david-crespo:
Will pick this back up later.

Issue #2187, opened 2024-04-23
Improve copy on form validation errors


charliepark:
![image](https://github.com/oxidecomputer/console/assets/22547/07104c9a-526c-420b-94ee-eff1b239b390)

![image](https://github.com/oxidecomputer/console/assets/22547/527c1cf6-2752-4623-aae0-999938216f1a)

These could direct the user a little more clearly

david-crespo:
These are API errors, so there's not that much we can do about them, or at least, if we were going to spend time getting elaborate client-side it would be better spent on the API side so all clients can benefit.

david-crespo:
Actually.... the reason this gets to be an API error is because we fail to catch it in client-side validation in the form. So we can probably improve that.

david-crespo:
Thinking more about this, some of the validation problem goes away if we're using comboboxes to pick the VPC, subnet, or instance names (https://github.com/oxidecomputer/console/issues/2065), but because we will not be constraining those fields to values that already exist (the lists are just there to help), we will still have the validation problem.

PR #2190, opened 2024-04-24
Rearrange firewall rules form and add help text

david-crespo:
The diff is a nightmare compared to the amount of changes because it's a lot of shuffling and fiddly margin stuff. I'd focus on checking the rendered result.

https://console-git-firewall-rules-help-oxidecomputer.vercel.app/projects/mock-project/vpcs/mock-vpc

image

image
﻿
image

### DIFF
--- a/app/forms/firewall-rules-create.tsx
+++ b/app/forms/firewall-rules-create.tsx
@@ -247,7 +247,7 @@ export const CommonFields = ({ error, control }: CommonFieldsProps) => {
       <h3 className="mb-4 text-sans-2xl">Targets</h3>
       <Message
         variant="info"
-        content=" Specify instances this rule applies to. For example, target a VPC to make this rule apply to every instance in the VPC. Targets are additive: the rule applies to instances matching any of the specified targets."
+        content="Targets determine the instances to which this rule applies. You can target instances directly or specify a VPC, VPC subnet, IP, or IP subnet which will apply the rule to all matching instances. Targets are additive: the rule applies to instances matching any target."
       />
       {/* TODO: make ListboxField smarter with the values like RadioField is */}
       <ListboxField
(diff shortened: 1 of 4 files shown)


benjaminleonard:
Some of those notes could possible be moved to the help text / description directly on the input.

Whilst you're in here, some placeholder values would be wonderful.

app/forms/firewall-rules-create.tsx
```diff
@@ -182,8 +215,6 @@ export const CommonFields = ({ error, control }: CommonFieldsProps) => {
       <NameField name="name" control={control} />
       <DescriptionField name="description" control={control} />
 
-      <FormDivider />
-
       <NumberField
         name="priority"
```
benjaminleonard, line 219: Any thoughts about moving this to the end? I think having it first gives it undue importance.

david-crespo: Moved it to the bottom of the top section.

david-crespo:
I don't think I can squeeze the help into individual fields because it's really about the whole thing. Placeholders are also kind of tough because they depend on which type you have picked in the listbox.

david-crespo:
That's a cool idea. I'll see if there's a way to make it work.

david-crespo:
What you might be thinking of that came up before was making Enter submit the mini form when you're in one of those fields rather than submitting the whole form, we fixed that.

david-crespo:
I don't think I can make it make sense to block leftover stuff in the input. You wouldn't want to just disable submit, you'd want it to be a validation error you can go find, but you'd only want it to be triggered by submit. We'd have to get pretty elaborate and I don't think it's worth it.

I added validation for port range because it's easy, but for the target and host fields, it's a lot more complicated because the way you validate the value input depends on which type is selected in the listbox. That's captured in #2187

david-crespo:
I'm going to merge this as-is and we can redo the text boxes in a different style as a followup per design feedback.

PR #2194, opened 2024-04-24
Add DocsPopover with page-specific links to docs

### LINKED ISSUE #2114: Change docs link icon to ? or "docs" or "help"
Letter I icon is not evocative enough.

image

### LINKED ISSUE #2210: Animate `DocsPopover`
I don't want to block #2194 on this because it's already great without an animation, but I think an animation would be nice and it would match how dropdown menu works. Apparently we do it in CSS:

https://github.com/oxidecomputer/console/blob/042d5ef3036563967b1b58cea2cffe47df11151f/app/ui/styles/components/menu-button.css#L36-L55

charliepark:
Fixes #2114 
Fixes #2210

This PR introduces new page-specific links to the Oxide Docs. 

Here you can see the trigger for the popover in the upper-right corner …
Screenshot 2024-04-26 at 10 17 14 AM
Screenshot 2024-04-26 at 10 15 41 AM

And the popover when it shows up:
Screenshot 2024-04-26 at 10 17 23 AM
Screenshot 2024-04-26 at 10 15 52 AM


package.json
```diff
@@ -59,6 +59,7 @@
     "react-hook-form": "^7.51.3",
     "react-is": "^18.2.0",
     "react-merge-refs": "^2.1.1",
+    "react-popper": "^2.3.0",
```
david-crespo, line 62: we're using floating-ui elsewhere, which is the successor to popper. even though the headless docs say to use popper, you can probably make this work with floating-ui

david-crespo: Example here: https://github.com/tailwindlabs/headlessui/blob/main/playgrounds/react/pages/menu/menu-with-floating-ui.tsx

Linked from here: https://github.com/tailwindlabs/headlessui/issues/2780

paryhin:
@charliepark the icon next to a document name is slightly misplaced, it should be pushed down a pixel or two :)

app/pages/ProjectsPage.tsx
```diff
@@ -121,6 +123,17 @@ export function ProjectsPage() {
...
         <PageTitle icon={<Folder24Icon />}>Projects</PageTitle>
+        <ContextualDocsModal
+          heading="Projects"
+          icon={<Folder16Icon />}
+          summary="Projects are logical containers for managing compute, storage and network resources within a silo."
+          links={[
+            {
+              href: links.projectsDocs,
+              linkText: 'Managing Projects',
+            },
```
david-crespo, line 134: I was thinking these feel a little noisy, and it occurred to me — instead of putting just the href in the `links` object, you could put the whole thing. Then this would be as neat as

```ts
links={[docLinks.projects]}
```

charliepark:
Latest:
Screenshot 2024-04-25 at 5 25 21 PM

app/pages/system/inventory/InventoryPage.tsx
```diff
@@ -28,6 +30,12 @@ export function InventoryPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Racks24Icon />}>Inventory</PageTitle>
+        <ContextualDocsModal
+          heading="Sleds and Disks"
+          icon={<Racks24Icon />}
```
charliepark, line 35: We'll need a `<Racks16Icon />`, which I don't believe exists yet.

charliepark:
Screenshot 2024-04-25 at 5 40 36 PM

Just noticed this slightly different background color for the different buttons; will want to chat about best approach.

app/components/ContextualDocsModal.tsx
```diff
@@ -0,0 +1,80 @@
...
+)
+
+type ContextualDocsModalProps = {
+  heading: React.ReactNode
+  icon: JSX.Element
+  links: Array<ContextualDocsLinkProps>
+  summary: string
+}
+
+export const ContextualDocsModal = ({
```
david-crespo, line 40: This isn’t a modal and contextual is kind of a mouthful — how about `DocsPopover`?

charliepark: Makes sense!

app/pages/system/inventory/InventoryPage.tsx
charliepark: @paryhin or @benjaminleonard, if one of you would be able to create a Racks16Icon, (or show me how,) I'd appreciate it!

charliepark:
I noticed that the popover scrolls with the page, rather than staying locked to the trigger. You can see this in action here: http://localhost:4000/projects/mock-project/disks

I'll need to look into this tomorrow morning; I suspect there's an option we can adjust in the floating-ui configs.

david-crespo:
We are going to need some design guidance on the pile o' buttons situation as well.

image

app/components/DocsPopover.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { refs, floatingStyles } = useFloating({
+    placement: 'bottom-end',
+    middleware: [offset(6)],
+  })
+  return (
+    <Popover>
+      <Popover.Button
+        ref={refs.setReference}
+        className="ox-button btn-secondary flex h-8 w-9  shrink-0 items-center justify-center gap-2 rounded-lg border align-top text-mono-md bg-raise border-secondary elevation-1 svg:w-5"
+      >
```
david-crespo, line 50: You can make it use `Button`:

```ts
<Popover.Button as={Button} ref={refs.setReference} size="icon" variant="secondary">
```

david-crespo:
Looks like this works to keep it in place on scroll.

```ts
whileElementsMounted: autoUpdate,
```

https://floating-ui.com/docs/useFloating#whileelementsmounted

app/components/DocsPopover.tsx
```diff
@@ -0,0 +1,75 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+
+import { offset, useFloating } from '@floating-ui/react'
+import { Popover } from '@headlessui/react'
```
benjaminleonard, line 10: Are we testing the `@headlessui` alternative to the one we're using elsewhere? It's probably a good idea because we were seeing some performance issues with radix.

app/pages/system/inventory/InventoryPage.tsx
benjaminleonard: The 24 one was named wrong. It should be `Servers24` and you'll find `Servers16` already exists. I'll update the 24 one separately.

benjaminleonard:
> Just noticed this slightly different background color for the different buttons; will want to chat about best approach.

We should have been clearer. That's the hover state, when idle its the same background as the others.

benjaminleonard:
Alongside some small spacing tweaks. I also added:

*Truncating to a single line for longer guide names*
CleanShot 2024-04-26 at 10 40 48@2x

*Made entire row clickable and trigger hover state*
CleanShot 2024-04-26 at 10 41 10

app/components/DocsPopover.tsx
benjaminleonard: Oh I missed this, came to a similar conclusion by using `buttonStyle`

benjaminleonard:
One side effect of this change is we can strip out the `<TableControlsText>` in most (if not all) present usage.

david-crespo:
The border between the two links doesn’t make sense to me. It makes it look like “guides” only applies to the first one.

benjaminleonard:
That's fair, will revert to the non full-width version.

image

app/components/DocsPopover.tsx
david-crespo: Yeah, I have this hunch that headless is just better overall. I didn’t realize we were using Radix Popover already — looks like it’s only for the date picker. Would probably be pretty easy to switch it over.

david-crespo:
That’s design baby!

benjaminleonard:
Credit to Eugene he had it like that first before I ruined it.

app/components/DocsPopover.tsx
charliepark: Ah, nice. My approach had felt dirty; this is a lot better.

app/pages/SiloAccessPage.tsx
```diff
@@ -162,6 +164,12 @@ export function SiloAccessPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Access24Icon />}>Access &amp; IAM</PageTitle>
+        <DocsPopover
+          heading="Access"
+          icon={<Access16Icon />}
+          summary=""
```
charliepark, line 170: I didn't have a great summary for this; I think "Access" is pretty straightforward without a description, though.

charliepark: nvm

david-crespo: > Configure user and group access to this <project/silo>.

🤷

david-crespo: I like yours better!

app/components/DocsPopover.tsx
david-crespo: Looked into it and it's not Radix, it's React Aria's `usePopover`, which is pretty well integrated with their date picker stuff (the only context we use it). So I don't think it's worth teasing apart.

https://github.com/oxidecomputer/console/blob/e8e4f576e6ee5c3e3544eaf5230a71e07c79a428/app/ui/lib/Popover.tsx

app/pages/ProjectsPage.tsx
```diff
@@ -121,6 +123,12 @@ export function ProjectsPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Folder24Icon />}>Projects</PageTitle>
+        <DocsPopover
+          heading="Projects"
+          icon={<Folder16Icon />}
+          summary="Projects are logical containers for managing compute, storage and network resources within a silo."
```
david-crespo, line 129: Would prefer to avoid "logical", missing Oxford comma. Not sure whether should mention the silo because for end users the silo is their whole universe. On the other hand, maybe conveying that projects are children of silos is useful to do.

> Projects are containers for managing compute, storage, and network resources.

Lower priority, I think we could try being more concrete about what is in a project.

> Projects are containers for managing resources like instances, disks, and VPCs.

app/pages/SiloAccessPage.tsx
```diff
@@ -162,6 +164,12 @@ export function SiloAccessPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Access24Icon />}>Access</PageTitle>
+        <DocsPopover
+          heading="Access"
+          icon={<Access16Icon />}
+          summary="Silo-level access roles and permissions. Permissions set at the silo level cascade down to all projects within the silo."
+          links={[docLinks.keyConceptsIam, docLinks.access]}
+        />
```
david-crespo, line 172: This and project access should be more similar. Since "permissions" is not really an API concept, I would avoid it.

> Set silo-level roles for users and groups. Silo roles apply to all projects within the silo.

I wonder if we should also mention that you can override these with project roles. That might be too much, and covered by the docs. The problem is you also have to explain that you can only override if the role is stronger.

app/util/links.ts
```diff
@@ -6,8 +6,102 @@
...
+    href: links.firewallRulesDocs,
+    linkText: 'Firewall Rules',
+  },
+  floatingIps: {
+    href: links.floatingIpsDocs,
+    linkText: 'Floating IPs',
+  },
+  keyConceptsIam: {
+    href: links.keyConceptsIamPolicyDocs,
+    linkText: 'Key Concepts: Identity & Access Management Policy',
```
david-crespo, line 57: This truncates. Maybe you could abbreviate IAM?

Screenshot 2024-04-29 at 5 02 03 PM

app/util/links.ts
```diff
@@ -6,8 +6,102 @@
...
+    href: links.systemMetricsDocs,
+    linkText: 'System Metrics',
+  },
+  systemSilo: {
+    href: links.systemSiloDocs,
+    linkText: 'Silo Management',
+  },
+  vms: {
+    href: links.vmsDocs,
+    linkText: 'Virtual Machines and Deploying Workloads',
```
david-crespo, line 101: I'd just call this `Deploying Workloads` since that's the title of the linked doc.

app/pages/project/access/ProjectAccessPage.tsx
```diff
@@ -196,6 +198,12 @@ export function ProjectAccessPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Access24Icon />}>Access</PageTitle>
+        <DocsPopover
+          heading="Access"
+          icon={<Access16Icon />}
+          summary="Identity and access management roles determine who can view, edit, or administer this project. Silo roles are inherited from the organization, while project roles are specific to this project."
```
david-crespo, line 204: "organization" made me flinch. Guess I am traumatized. https://github.com/oxidecomputer/omicron/pull/2626

I do like the first sentence, so it might be nice to add that to the silo one too. I'd get rid of "Identity and access management" and just say 

> Roles determine who can view, edit, or administer this project.

And on silo access you could do

> Roles determine who can view, edit, or administer this silo and the projects within it.

The second sentence is pretty good, but I'd get rid of organization like I said. I also think it would be worth mentioning what happens when you have a role from each, even though it makes the text longer, because that is so fundamental.

> Silo roles are inherited from the silo. If a user or group has both a silo and project role, the stronger role takes precedence.

app/pages/project/disks/DisksPage.tsx
```diff
@@ -162,6 +164,12 @@ export function DisksPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Storage24Icon />}>Disks</PageTitle>
+        <DocsPopover
+          heading="Disks"
+          icon={<Storage16Icon />}
+          summary="Disks are persistent block-storage volumes that can be managed independently from VM instances."
```
david-crespo, line 170: I think it would be "block storage" or "block-level storage" but not "block-storage". I'll make a note to fix that in the docs too. Regardless, in this case maybe we can drop it. It's sort an implementation detail.

app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -184,17 +184,18 @@ export function FloatingIpsPage() {
 
   return (
     <>
-      <PageHeader className="!mb-12">
+      <PageHeader>
         <PageTitle icon={<IpGlobal24Icon />}>Floating IPs</PageTitle>
+        <DocsPopover
+          heading="Floating IPs"
+          icon={<IpGlobal16Icon />}
+          summary="Floating IPs can be assigned to target instances, making it easier to host services from a consistent address."
```
david-crespo, line 192: This is the right direction — I like the gesture at the purpose — but it's kind of vague somehow. This isn't the right wording but here is what I would want to convey.

> A floating IP is an IP address with an independent existence that can be attached and detached from instances.

"Independent existence" feels kind of ridiculous; "persistent" is more in the right direction but has a technical meaning that is misleading. Maybe independent existence is tolerable. Claude suggests other wordings that are the same words but feel better:

> Floating IPs exist independently of instances, allowing them to be dynamically attached and detached as needed.

app/pages/project/images/ImagesPage.tsx
```diff
@@ -98,6 +100,12 @@ export function ImagesPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Images24Icon />}>Images</PageTitle>
+        <DocsPopover
+          heading="Images"
+          icon={<Images16Icon />}
+          summary="Images are templates of virtual machines that can be used to create new instances."
```
david-crespo, line 106: They are templates of disks, not instances, and it's not exactly a template in that for me a template is like a form with slots you can fill in. "Snapshot" is a better word, but that is its own thing. I don't have a good suggestion here yet.

app/pages/project/instances/InstancesPage.tsx
```diff
@@ -124,6 +126,12 @@ export function InstancesPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Instances24Icon />}>Instances</PageTitle>
+        <DocsPopover
+          heading="Instances"
+          icon={<Instances16Icon />}
+          summary="Instances are virtual machines that run on Oxide."
```
david-crespo, line 132: We should ask Kevin and others if there's some canonical phrasing for "on Oxide", maybe "on the Oxide platform" or "on the Oxide rack"

david-crespo (commented):
Dumping my first pile of comments on you because I have to stop for the day

app/pages/project/instances/InstancesPage.tsx
david-crespo: “on the Oxide platform” was the consensus in the marketing channel.

app/pages/project/access/ProjectAccessPage.tsx
charliepark: Nice.

app/pages/project/images/ImagesPage.tsx
david-crespo: Claude suggests "point-in-time copy", that's kind of interesting, though we might reserve that for snapshots.

david-crespo:
On second thought, maybe “on the Oxide platform” is still too marketing-y for the console. It’s true of everything in the console anyway, so it’s not clear why only instances deserves to say it. “Instances are virtual machines.” is a little short thought, maybe worth padding out some other way.

app/pages/project/floating-ips/FloatingIpsPage.tsx
charliepark: Will go with that for now, though happy to revise if we come up with improvements.

app/pages/project/images/ImagesPage.tsx
charliepark: Maybe `blueprints`?
"Images serve as blueprints of disks, used to create new virtual machine instances."

david-crespo: Maybe we avoid it by saying what they do without saying what they are.

> Images let you create a new disk based on an existing one. Images can be uploaded directly or created from a snapshot.

The “you” feels a little too casual. I hope most people using the system should have some sense of what an image is.

app/pages/system/silos/SilosPage.tsx
```diff
@@ -112,6 +114,12 @@ export function SilosPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Cloud24Icon />}>Silos</PageTitle>
+        <DocsPopover
+          heading="Silos"
+          icon={<Cloud16Icon />}
+          summary="Silos are collections of resources, designed to provide strict tenancy separation. They have separate UI and API endpoints, separate resource limits, and separate access policies."
```
david-crespo, line 120: This is almost there, but I feel like the "collections of resources" bit hurts more than it helps. And "separate UI and API endpoints" could be phrased more concretely. Something like:

> Silos provide strict tenancy separation between groups of users. Each silo has its own resource limits and access policies as well as its own subdomain for the web console and API.

david-crespo (changes requested):
Second pile of notes

app/pages/system/silos/SiloPage.tsx
```diff
@@ -54,6 +56,12 @@ export function SiloPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Cloud24Icon />}>{silo.name}</PageTitle>
+        <DocsPopover
+          heading="Silos"
+          icon={<Cloud16Icon />}
+          summary="Silos are collections of resources, designed to provide strict tenancy separation. They have separate UI and API endpoints, separate resource limits, and separate access policies."
+          links={[docLinks.systemSilo]}
```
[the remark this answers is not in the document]
charliepark: Since these are the tabs on the Silo page, possibly also the Access docs?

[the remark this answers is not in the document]
david-crespo: Yeah, good idea.

app/pages/system/networking/IpPoolPage.tsx
```diff
@@ -73,7 +72,13 @@ export function IpPoolPage() {
...
     <>
       <PageHeader>
-        <PageTitle icon={<Networking24Icon />}>{pool.name}</PageTitle>
+        <PageTitle icon={<IpGlobal24Icon />}>{pool.name}</PageTitle>
+        <DocsPopover
+          heading="IP Pools"
+          icon={<IpGlobal16Icon />}
+          summary="IP pools are collections of external IP addresses that let you control which silos can use which IPs."
+          links={[docLinks.systemIpPools]}
+        />
```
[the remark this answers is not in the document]
charliepark: I think this is a little wordy, but works …
Screenshot 2024-04-30 at 10 57 22 AM

app/pages/SiloUtilizationPage.tsx
```diff
@@ -72,6 +74,12 @@ export function SiloUtilizationPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Metrics24Icon />}>Capacity &amp; Utilization</PageTitle>
+        <DocsPopover
+          heading="Utilization"
+          icon={<Metrics16Icon />}
+          summary="System metrics allow you to monitor the allocation and utilization of your resources across silos, including CPU, memory, and storage."
```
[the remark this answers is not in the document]
charliepark: Not to be completely hamstrung by the docs, but they don't yet mention quotas. That might be something to update here once we update the docs?

app/pages/project/vpcs/VpcsPage.tsx
```diff
@@ -120,6 +122,12 @@ export function VpcsPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Networking24Icon />}>VPCs</PageTitle>
+        <DocsPopover
+          heading="VPCs"
+          icon={<Networking16Icon />}
+          summary="VPCs are private networks that serve to isolate sets of instances from each other. Instances within a VPC can talk to each other over private IP addresses (if firewall rules allow it) but communication between VPCs must happen through external IPs."
```
david-crespo, line 128: Was able to cut a line by cutting a few words. Also "using private IP addresses" feels slightly more correct than "over private IP addresses"

> VPCs are private networks that isolate sets of instances from each other. Instances within a VPC can talk to each other using private IP addresses (if firewall rules allow it) but traffic between VPCs must go through external IPs.

image

app/pages/SiloUtilizationPage.tsx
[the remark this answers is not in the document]
charliepark: Or, I dunno, this might work: `System metrics let you monitor utilization of CPU, memory, and storage, in contrast to the total quota available to you.`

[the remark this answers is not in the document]
david-crespo: Yeah, something like that. I wanted to say "against quotas" but it's hard to say it because there is the quotas for each individual silo and then the sum of all the quotas. And then of course we intend to also add a true (virtual) total capacity.

app/pages/system/networking/IpPoolPage.tsx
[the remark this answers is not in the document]
david-crespo: Yeah, I think it's better to have a bit more info despite the length, and we can make it a bit shorter. I went back and forth thinking about saying different things on IP pools list and IP pool detail, but I think this would be fine and works for both.

> IP pools are collections of external IPs you can assign to silos. When a pool is linked to a silo, users in that silo can allocate IPs from the pool for their instances.

Screenshot 2024-04-30 at 2 23 43 PM

I think the default line is a bit much for the popover, but what if we stuck it in a tooltip on the default column? That column header is ridiculous anyway, it demands explanation.

app/pages/SiloUtilizationPage.tsx
```diff
@@ -72,6 +74,12 @@ export function SiloUtilizationPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Metrics24Icon />}>Capacity &amp; Utilization</PageTitle>
+        <DocsPopover
+          heading="Utilization"
+          icon={<Metrics16Icon />}
+          summary="System metrics let you monitor utilization of CPU, memory, and storage, in contrast to the total quota available to you."
```
david-crespo, line 80: How about 

> System metrics let you monitor utilization of CPU, memory, and storage against silo-level quotas.

david-crespo: Sorry, I meant that for system utilization, and for here I would leave out the bit about quotas:

> System metrics let you monitor utilization of CPU, memory, and storage

app/pages/system/networking/IpPoolPage.tsx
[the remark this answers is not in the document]
charliepark: Added that tooltip to the table.

david-crespo (approved):
THINK WE GOT IT

PR #2207, opened 2024-04-26
Update copy regarding disk attach / detach requirements

charliepark:
Currently, we have some copy on the Storage tab of the Instance page, where we tell the user that certain conditions have to be met (the instance has to be stopped) before the user can take specific actions (detaching the disk). There's some other copy that — while technically accurate (probably) — aren't things that the user really cares about in that moment.

image

This PR updates that view, so that the thing the user needs to know — you need to stop the instance in order to detach the disk, or add a new disk — is more clearly linked to the next step they need to take — stopping the instance.
Screenshot 2024-04-26 at 3 48 24 PM

I'm open to copy suggestions, as always, though I think what I have here is pretty clear.

### DIFF
--- a/app/pages/project/instances/instance/tabs/StorageTab.tsx
+++ b/app/pages/project/instances/instance/tabs/StorageTab.tsx
@@ -176,8 +176,8 @@ export function StorageTab() {
             onClick={() => setShowDiskCreate(true)}
             disabledReason={
               <>
-                Instance must be <span className="text-default">stopped</span> to create a
-                disk
+                Instance must be <span className="text-default">stopped</span> to create and
+                attach a disk
               </>
             }
             disabled={!instanceCan.attachDisk(instance)}
--- a/app/pages/project/instances/instance/tabs/StorageTab.tsx
+++ b/app/pages/project/instances/instance/tabs/StorageTab.tsx
@@ -62,9 +62,6 @@ const staticCols = [
   colHelper.accessor('timeCreated', Columns.timeCreated),
 ]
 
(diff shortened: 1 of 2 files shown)


app/pages/project/instances/instance/tabs/StorageTab.tsx
```diff
@@ -174,7 +174,12 @@ export function StorageTab() {
           <Button
             size="sm"
             onClick={() => setShowDiskCreate(true)}
-            disabledReason={<>Instance must be {attachableStates} to create a disk</>}
+            disabledReason={
+              <>
+                Instance must be <span className="text-default">stopped</span> to create a
+                disk
+              </>
+            }
```
david-crespo, line 182: could it be worth saying “create and attach” here? it’s the attach that’s the problem, right?

app/pages/project/instances/instance/tabs/StorageTab.tsx
```diff
@@ -128,7 +125,10 @@ export function StorageTab() {
       {
         label: 'Detach',
         disabled: !instanceCan.detachDisk(instance) && (
-          <>Instance must be in state {detachableStates} before disk can be detached</>
+          <>
+            Instance must be <span className="text-default">stopped</span> before disk can
+            be detached
```
david-crespo, line 130: might as well do what the other ones, fudge it grammatically to save a few words: “instance must be stopped to detach disks” or maybe even “instance must be stopped” by itself? the button is disabled and it says “detach”

david-crespo (approved):
Looks good, not particularly attached to my comments!

PR #2214, opened 2024-05-01
Switching from `@react-spring/web` to `framer-motion`

benjaminleonard:
These libraries are very similar, but in places I prefer the `framer-motion` API. One such case is that you can provide it an apple style `duration` and `bounce` and still get a spring animation but without needing to contort your brain with friction, mass etc.

`framer-motion` does have a much bigger bundle but you can keep it down by excluding the features we don't need. The [layout features](https://www.framer.com/motion/layout-animations/) in particular are very powerful but we don't do any big native style transitions to make its inclusion justifiable. 

```
`framer-motion`
dist/assets/app-ODvCp2jt.css                                140.78 kB │ gzip:  21.78 kB
dist/assets/app-DSIP2Da4.js                               1,124.33 kB │ gzip: 349.94 kB │ map: 5,436.95 kB

`react-spring`
dist/assets/app-zFe_5Sh-.css                                138.96 kB │ gzip:  21.43 kB
dist/assets/app-CT1MsEJt.js                               1,100.80 kB │ gzip: 344.06 kB │ map: 5,243.37 kB
```

WIP as I might find a few other spots where we might include animation.


david-crespo:
It's pretty out of date but I think this is a fine thing to do. I'm ok with a 6kb bundle bump if it lets us get nicer animations in the future. It looks like the bundle size reduction [tips](https://motion.dev/docs/react-reduce-bundle-size) might have changed and and are now easier to follow. If you still want this @benjaminleonard, could you redo it on latest main?

benjaminleonard:
Closing and re-doing in #2646

PR #2217, opened 2024-05-02
Soft validating image upload

benjaminleonard:
This PR contains some small validations that help nudge the user if they're uploading an image unlikely to boot. 

**We're checking for:**
1. The absence of `CD001` at `0x8001`(bootable CD), or a UEFI header (`EFI PART`) at offsets 512, 2048, 4096 are a strong indicator it's not bootable
2. File extensions that imply a compressed or unsuitable image, e.g. qcow2, vmdk or ones that end in gz or 7z
3. That the selected block size is the correct one for bootable CDs or the one that matches the location of the UEFI header

**Feedback wanted especially on whether:**
The UX validates softly enough so that we're not adding friction on users who have an image that would boot but doesn't meet what we expect, and for users that may be uploading an image that does not need to boot at all.

Note that this does not prevent form submission, it's merely a notice.

CleanShot 2024-05-02 at 12 57 46


david-crespo:
Weird flash was indeed fixed by our shuffling.

benjaminleonard:
I think this is about ready to go – and fairly low risk since it does not prevent the user from uploading an image even if it doesn't think its bootable. @iliana have you got a link to a bootable CD image I could tst?

app/forms/image-upload.tsx
```diff
@@ -640,3 +655,145 @@ export function CreateImageSideModalForm() {
...
+async function getEfiPartOffset(file: File) {
+  const offsets = [512, 2048, 4096]
+  for (const offset of offsets) {
+    const isMatch = await readAndMatch(file, offset, 8, 'EFI PART')
+    if (isMatch) return offset
+  }
+  return -1
+}
+
+const compressedExts = ['.gz', '.7z', '.qcow2', '.vmdk']
```
david-crespo, line 790: I think this all looks good, my main concern was getting the contents of a list like this right. @iliana ?

iliana:
> have you got a link to a bootable CD image I could tst?

well the ISO I downloaded to learn that we don't need a GPT was the Windows 10 retail ISO, so you could grab that off of Microsoft's site. although it's 6+ GiB :( but all the other CD images I'm aware of are hybrid ISO+GPT

I have been meaning to craft some small testing images for Propolis testing, might be worth having those for playwright tests here?

david-crespo:
Might be simpler to pass in more elaborate fake data here for `buffer`:

https://github.com/oxidecomputer/console/blob/9b20b7c9e6e8a9b3a45b10d73f4887e80782e567/test/e2e/utils.ts#L178-L189

app/forms/image-upload.tsx
```diff
@@ -640,3 +655,135 @@ export function CreateImageSideModalForm() {
...
+
+const validateImage = async (file: File) => {
+  const lowerFileName = file.name.toLowerCase()
+  return {
+    efiPartOffset: await getEfiPartOffset(file),
+    isBootableCd: (await readAtOffset(file, 0x8001, 5)) === 'CD001',
```
iliana, line 786: So this doesn't quite implement the decision tree of "if no `EFI PART` found, look for `CD001`. This is likely problematic in terms of what we tell customers for bootable hybrid GPT+ISO, which a lot of Linux distro ISOs are so that you can `dd` them directly to a flash drive (512-byte blocks!) and have them still boot. In this particular case, I think it's better to prefer the GPT and tell the user they should use 512-byte blocks.

My expectation is that one of these would return `{efiPartOffset: 512, isBootableCd: true, ...}` — would that always show a "this image might not be bootable" message of one kind or another?

iliana:
> it might be better to use test images that were not designed specifically for this

except for the fact that those are giant because they're operating systems :(

app/forms/image-upload.tsx
```diff
@@ -640,3 +655,135 @@ export function CreateImageSideModalForm() {
...
+}: {
+  blockSize: number
+  efiPartOffset: number
+  isBootableCd: boolean
+}) {
+  if (efiPartOffset === -1 && !isBootableCd) return null
+
+  // TODO: make sure this logic is correct
+
+  if (blockSize === efiPartOffset) return null
```
iliana, line 672: Oh, I just read this line. That should cover my concern.

david-crespo: It would probably help to move this logic over into `validateImage` so you can see everything up front. When you say "follow the GPT suggestion", you mean 4096 is correct and we should avoid warning if possible? How can we detect hybrid GPT+MBR?

iliana: oh sorry I meant hybrid GPT+ISO above. (hybrid GPT+MBR is a whole 'nother rabbit hole to fall down that we should ignore for this PR, and probably forever)

iliana: What I mean is if a user uploads a hybrid GPT+ISO, and the GPT is at offset 512, then the suggested block device size should be 512. If the user has 2048 selected, we maybe don't bother saying otherwise, but if the user has 4096 selected, that definitely won't boot and we should suggest 512.

benjaminleonard:
@iliana you had mentioned in chat you might write a separate library to check if an image is bootable, want to hold this PR for that?

iliana:
Tested my commit with:
- Alpine nocloud qcow2: `BootableNotice` displays as expected
- Alpine nocloud raw: Only 512 is correct, other options display "EFI PART" message
- Alpine ISO (hybrid GPT+ISO): Only 512 is correct, other options display "EFI PART" message
- A random Windows Server ISO I partially downloaded from archive.org: Only 2048 is correct, other options display "Bootable CDs" message

david-crespo:
Awesome, thank you!

david-crespo (approved):

benjaminleonard:
Lovely stuff! Thanks for humouring me @iliana

Issue #1099, opened 2022-07-28
Floating IPs in the instance create form


david-crespo:
Split off from #1097

Blocked on API implementation https://github.com/oxidecomputer/omicron/issues/1334

PR #2225, opened 2024-05-07
Add external IP selection to instance creation form

### LINKED ISSUE #1097: Ephemeral IP checkbox on instance create form
Beginning with https://github.com/oxidecomputer/omicron/pull/1458, the control plane now has the concept of an external IP address for guests. This is a publicly-visible address that allows inbound connections, such as SSH, from outside the rack into the guest. (That's distinct from the IP addresses that provide outbound connectivity, which don't appear in the API.)

These addresses are associated with the primary network interface, and are shown in the console as part of https://github.com/oxidecomputer/console/pull/1070. We also need a way to request such an address in the web console. The parameters for describing this address are currently defined [here](https://github.com/oxidecomputer/omicron/blob/232806389f022cb5db0e9e4911aa6abb5fe90f00/nexus/types/src/external_api/params.rs#L409). We currently support ephemeral addresses, optionally specifying the name of an IP Pool from which to draw the address. (Ephemeral addresses only live as long as the instance, and can't move around between different objects.) It's not clear yet how to specify the pool, since that's currently a top-level / operator / global resource (discussed in RFD 288), so it might be better to leave that off for now.

Though they're not yet implemented, Floating IP addresses will also be supported here. (Implementation tracked in https://github.com/oxidecomputer/omicron/issues/1334.) Those are separate, named API resources scoped to a VPC, which can be created independently from instances. In that case, the name of the Floating IP will be specified in the API request (it'll be looked up relative to the VPC of the primary network interface, most likely).

### LINKED ISSUE #1098: Specify IP pool for ephemeral IP on instance create
Split off from #1097

> It's not clear yet how to specify the pool, since that's currently a top-level / operator / global resource (discussed in RFD 288), so it might be better to leave that off for now.

charliepark:
Fixes #1097
Fixes #1098

Currently, when the user creates an instance, we just create an ephemeral IP and attach it to the instance. With this PR, we give them a checkbox (default = checked) to let them opt out/in of the ephemeral IP creation step, and we give them a dropdown that lets them pick which IP pool to draw the ephemeral IP from.

We will also add the option to select a floating IP (#1099), but that'll be separate from this PR.

I'm not sure about the layout / copy I have in this PR, so would welcome design feedback from @paryhin. I was able to simplify [the designs in Figma](https://www.figma.com/file/zIBsgP7gAkhapyNvVOqsn6/Floating-IPs?type=design&node-id=427%3A22653&mode=design&t=SaUnqJRcykwBNbPD-1) a bit (we can consolidate some of the options there), but I think Eugene will have good suggestions on what I have here.

Screenshot 2024-05-06 at 5 58 37 PM

### DIFF
--- a/app/forms/instance-create.tsx
+++ b/app/forms/instance-create.tsx
@@ -18,7 +18,6 @@ import {
   INSTANCE_MAX_CPU,
   INSTANCE_MAX_RAM_GiB,
   useApiMutation,
-  useApiQuery,
   useApiQueryClient,
   usePrefetchedApiQuery,
   type InstanceCreate,
@@ -148,6 +147,7 @@ CreateInstanceForm.loader = async ({ params }: LoaderFunctionArgs) => {
       query: { project, limit: DISK_FETCH_LIMIT },
     }),
     apiQueryClient.prefetchQuery('currentUserSshKeyList', {}),
+    apiQueryClient.prefetchQuery('projectIpPoolList', { query: { limit: 1000 } }),
   ])
   return null
 }
@@ -191,6 +191,14 @@ export function CreateInstanceForm() {
   const { data: sshKeys } = usePrefetchedApiQuery('currentUserSshKeyList', {})
   const allKeys = useMemo(() => sshKeys.items.map((key) => key.id), [sshKeys])
 
+  const { data: allPools } = usePrefetchedApiQuery('projectIpPoolList', {
+    query: { limit: 1000 },
+  })
+  const defaultPool = useMemo(
+    () => (allPools ? allPools.items.find((p) => p.isDefault)?.name : undefined),
+    [allPools]
+  )
+
   const defaultSource =
     siloImages.length > 0 ? 'siloImage' : projectImages.length > 0 ? 'projectImage' : 'disk'
 
@@ -202,6 +210,7 @@ export function CreateInstanceForm() {
     diskSource: disks?.[0]?.value || '',
     sshPublicKeys: allKeys,
     bootDiskSize: nearest10(defaultImage?.size / GiB),
+    externalIps: [{ type: 'ephemeral', pool: defaultPool }],
   }
 
   const form = useForm({ defaultValues })
@@ -518,7 +527,11 @@ export function CreateInstanceForm() {
       <SshKeysField control={control} isSubmitting={isSubmitting} />
       <FormDivider />
       <Form.Heading id="advanced">Advanced</Form.Heading>
-      <AdvancedAccordion control={control} isSubmitting={isSubmitting} />
+      <AdvancedAccordion
+        control={control}
+        isSubmitting={isSubmitting}
+        allPools={allPools.items}
+      />
       <Form.Actions>
         <Form.Submit loading={createInstance.isPending}>Create instance</Form.Submit>
         <Form.Cancel onClick={() => navigate(pb.instances({ project }))} />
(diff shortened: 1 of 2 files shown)


charliepark:
Ben had some suggestions, which I incorporated in the recent PRs. Here's the latest …
Screenshot 2024-05-07 at 11 35 43 AM
Main changes from Ben …
- added microcopy and link to docs
- made select/option dropdown conditionally render, based on whether box is checked or not (parallel with radio group above, with selective rendering of "network interface" second-level options

I know Eugene has a few thoughts as well.

app/forms/instance-create.tsx
```diff
@@ -555,6 +581,71 @@ const AdvancedAccordion = ({
           control={control}
           disabled={isSubmitting}
         />
+
+        <div className="max-w-lg space-y-2">
+          <h2 className="text-sans-md">
+            Ephemeral IP{' '}
+            <TextInputHint id="ephemeral-ip-description" className="mb-2">
+              Ephemeral IPs are non-permanent, randomly-assigned addresses, dynamically
+              allocated from a pool of IPs when the instance is created.{' '}
```
david-crespo, line 590: I think this works about as well without the adjectives:

> Ephemeral IPs are dynamically allocated from a pool of IPs when the instance is created.

paryhin:
CleanShot 2024-05-07 at 23 26 25@2x

1. We should keep the title for the IP pool input, as currently it is not clear what you are selecting.
2.  We should show "default" pool badge in the input in its rest state.
3. If we want to have this type of long description with links, we may consider the following behavior: 
CleanShot 2024-05-07 at 23 23 18

charliepark:
I'm not against having some kind of [\<details>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) component to reveal the text content, but it might need a different visual treatment, if all of our other info icons have a tooltip?

charliepark:
The latest: Screenshot 2024-05-07 at 7 18 11 PM

david-crespo:
Looking good. How about “Allocate” instead of “Automatically assign”? Automatic is redundant, everything the form does is automatic. And “assign” isn’t really a term we use, it would be attach. “Allocate and attach” would be most precise but is kind of long.

david-crespo:
I agree with Eugene that the fact that the listbox is for picking a pool should be indicated somehow. If the name of the pool wasn’t “ip-pool-1” the problem would be more obvious.

charliepark:
Ah, right; I'd meant to get that in there.

charliepark:
I can see what Eugene was getting at with the microcopy being on the screen. With the various titles and labels, it was getting crowded. I've put it into a tooltip for now, though that loses the link to find out more.
Screenshot 2024-05-07 at 9 39 00 PM

paryhin:
Yeah, switching from a tooltip is probably a separate PR, we can keep it as it is for now. If we want the link, we may put it in the contextual docs popover.  Everything else looks good to me.

david-crespo:
I think it needs more space and more hierarchy to the text. Here is a sort of maximalist approach: 1rem under heading, checkbox, and listbox, 2.5rem above heading, heading one size bigger, checkbox label text-secondary. The size bump is a problem because it doesn't match the other ones. It makes me wonder if all the input labels in forms should be one size up, or at least whether we should have some type pattern that makes it easier to imply grouping.

image

app/forms/instance-create.tsx
```diff
@@ -526,14 +546,28 @@ export function CreateInstanceForm() {
...
+  const assignEphemeralIp = !!ephemeralIp
+  const selectedPool = ephemeralIp && 'pool' in ephemeralIp ? ephemeralIp.pool : undefined
+  const defaultPool = siloPools.find((pool) => pool.isDefault)?.name
+
+  const poolLabel = (pool: { name: string; isDefault: boolean }) => (
+    <div className="flex items-center gap-2">
+      {pool.name}
+      {pool.isDefault && <Badge>default</Badge>}
+    </div>
+  )
```
david-crespo, line 570: this is a pure function, easy to either move out of render, or honestly just inline it at the call site

david-crespo (approved):
Looks great. Happy to have design follow up on the spacing and hierarchy stuff.

PR #2227, opened 2024-05-07
Use installed copy of openapi-gen-ts and prettier

david-crespo:
I realized while messing with package versions that while using `npx @oxide/openapi-gen-ts@0.2.0` is cute and self-contained, hard-coding the version means you would never get alerted to vulns with `npm audit` or new versions with `npm outdated`. So it seems better to hook into the local `npm` setup.


package.json
```diff
@@ -23,6 +23,8 @@
     "lint": "eslint --ext .js,.ts,.tsx,.json .",
     "lint-fast": "eslint --cache --fix --ext .js,.ts,.tsx,.json .",
     "fmt": "prettier --cache --write . && npm run lint -- --fix",
+    "openapi-gen-ts": "openapi-gen-ts",
+    "prettier": "prettier",
```
david-crespo, line 27: This feels silly but the only other way I know to use the installed versions in the script is to hit `node_modules/.bin/prettier` and `node_modules/.bin/openapi-gen-ts` directly. I guess we could do that?

PR #2237, opened 2024-05-10
Add docs popover button to instance create

### LINKED ISSUE #2221: Add docs popover button to instance create
Whoops! We forgot.

image

charliepark:
Fixes #2221 

The full-page form has a custom header, so we'll need to pass the DocsPopover component in as an optional prop.

Currently, `instance-create` is the only place where we call `FullPageForm`, so there aren't other places we need to add this in for now.

Screenshot 2024-05-10 at 3 12 39 PM

### DIFF
--- a/app/components/form/FullPageForm.tsx
+++ b/app/components/form/FullPageForm.tsx
@@ -44,6 +44,7 @@ interface FullPageFormProps<TFieldValues extends FieldValues> {
    * constrain the `name` prop to paths in the values object.
    */
   children: ReactNode
+  docsPopover?: ReactNode
 }
 
 const PageActionsContainer = classed.div`flex h-20 items-center gutter`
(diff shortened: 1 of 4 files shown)


app/components/form/FullPageForm.tsx
```diff
@@ -87,6 +89,7 @@ export function FullPageForm<TFieldValues extends FieldValues>({
     <>
       <PageHeader>
         <PageTitle icon={icon}>{title}</PageTitle>
+        {docsPopover}
       </PageHeader>
```
david-crespo, line 93: This is definitely satisfactory, but I am always looking for excuses to unbundle these giant components. And this one is easy. `title` and `icon` and `docsPopover` are only used in this header. So how about moving this whole `PageHeader` inside `app/forms/instance-create.tsx` and then we can eliminate all three props here.

david-crespo (approved):
woohoo. love deleting props

PR #2238, opened 2024-05-10
update inventory sidebar link

### LINKED ISSUE #2173: Inventory link in sidebar not highlighted when on physical disks tab
Should be an easy fix.

https://console-l2rqcn9lt-oxidecomputer.vercel.app/system/inventory/disks

image

charliepark:
Fixes #2173

Currently, the sidebar link directs users to `/inventory/sleds`, which means that the ReactRouter magical `isActive` prop doesn't know that it should be highlighted for either `/inventory/sleds` OR `/inventory/disks`. By changing the route to just `/inventory`, it matches for both, and so it highlights the sidebar item regardless of which tab they're on.
Screenshot 2024-05-10 at 4 01 09 PM
Screenshot 2024-05-10 at 4 01 18 PM

### DIFF
--- a/app/layouts/SystemLayout.tsx
+++ b/app/layouts/SystemLayout.tsx
@@ -108,7 +108,7 @@ export function SystemLayout() {
           <NavLinkItem to={pb.systemUtilization()}>
             <Metrics16Icon /> Utilization
           </NavLinkItem>
-          <NavLinkItem to={pb.sledInventory()}>
+          <NavLinkItem to={pb.inventory()}>
             <Servers16Icon /> Inventory
           </NavLinkItem>
           <NavLinkItem to={pb.ipPools()}>


david-crespo:
There are some loader and back button shenanigans that can happen when you link to the parent and rely on the replace nav. I’ll show you Monday how to test it out. If it doesn’t work well we might have to get more clever with the active style logic to match on `/inventory` while keeping the link to `/inventory/sleds`.

https://github.com/oxidecomputer/console/blob/efceb17dbaca35edcc27c2c755699cd16c901b95/app/routes.tsx#L151-L160

david-crespo (approved):
Exorcised my worry in #2253. Could not find any issues with this approach. We may want to update some other route tab pages in followup PRs if they have the same problem.

PR #2242, opened 2024-05-13
Update OMICRON_VERSION

charliepark:
Bump OMICRON_VERSION to latest sha (d2ed4524f17ae025fb2e55956d7c82d377872a3e)

PR #2245, opened 2024-05-15
Add IP Pool to table of floating IPs

### LINKED ISSUE #2243: Show IP pool on floating IPs list
Followup to #2242. Now that we have `ip_pool_id` on the floating IP response, we can show the IP pool name in the column. We need to resolve the ID into the name. We should use the `projectIpPoolList` endpoint in the loader to `setQueryData` for `projectIpPoolView`, then we can just `useApiQuery` for the cell as normal but it'll be instant. Like this:

https://github.com/oxidecomputer/console/blob/b603d2dd6152bb1998e1097d55d05fcdec7d49ac/app/pages/project/disks/DisksPage.tsx#L57-L70

charliepark:
Fixes #2243 

In #2242 we added the `ip_pool_id` to the response object for each Floating IP. Using the `ipPoolView` endpoint, we can pull the name of each pool, meaning we can add the IP Pool's name to the table of Floating IPs.

Hovering over the IP Pool name gives a tooltip with the description of the IP Pool, if it exists.

Screenshot 2024-05-15 at 9 44 05 AM

As there is not currently an IP Pool view at the project level (just at the system level), it doesn't make sense to set the IP Pool cell as a link (like we do with the `attached to instance` column), so we're just using a `<Badge>` component.

I made a slight tweak to the Tooltip component's prop types, as I had some confusion over why a `string` passed in as a child of the `Tooltip` wasn't working.

### DIFF
--- a/app/pages/project/floating-ips/FloatingIpsPage.tsx
+++ b/app/pages/project/floating-ips/FloatingIpsPage.tsx
@@ -13,6 +13,7 @@ import { Outlet, useNavigate, type LoaderFunctionArgs } from 'react-router-dom'
 import {
   apiQueryClient,
   useApiMutation,
+  useApiQuery,
   useApiQueryClient,
   usePrefetchedApiQuery,
   type FloatingIp,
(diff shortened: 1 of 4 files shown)


david-crespo (approved):
Excellent. This will be useful.

One possible tweak: change `IP` to `Address` or `IP address`? I feel like we've rehashed this question a million times in different contexts. In this case it's the presence of IP and IP pool next to each other that looks a little silly. I could go either way.

image

benjaminleonard:
There's a couple things we should continue to think about here – the first is when we use badges vs regular text fields. Badge might be reasonable here because it's a group of a thing, but I'm not 100% sold on it. The other side effect is that we should probably use `!normal-case` when we include something that has its own casing like the name of an IP pool.

Either way, not super important.

benjaminleonard:
That'll teach me for going through PRs chronologically ... https://github.com/oxidecomputer/console/pull/2247

PR #2246, opened 2024-05-15
Handle serial console `close` event properly

### LINKED ISSUE #2169: Serial console shows "Connected" when instance is not running
I have an instance that is stopped. When I navigate to its serial console in the console, it shows a blank terminal (as one would expect), but also "Connected" in the lower right. No errors display.

When I try to do the same thing in the CLI, I get:

```
$ oxide instance serial console --project iliana --instance axiomatic

Connection closed: Error
Server error: Invalid Request: instance is stopped and has no active serial console server
```

david-crespo:
Closes #2169

This is what it looks like when you try to connect to a stopped instance.

image

The error state is quite ugly, but I'm going to merge this and get design notes later because a) I think this is rarely seen, and b) overall this is an important fix.

image


david-crespo:
This is almost there, except @benjaminleonard @paryhin I need something to show here when the connection has closed. (Right now we show this for any state other than `open`.) It's not really an error state (the instance is stopped), it should be neutral. We can say the instance is in X state.

image

benjaminleonard:
I suppose we don't have any serial console history in any other state than connected?

We'd probably want to include something that takes the user to a screen where they can start the instance, or let them start the instance directly here. But that might get a bit complicated if we need to handle the instance not coming up.

Do we want any docs here? We don't have any guides around the serial console, but we could include something around troubleshooting images ... or our requirements (https://docs.oxide.computer/guides/architecture/os-hypervisor-storage#_guest_workload_support).

david-crespo:
That’s interesting! We do have a serial console history endpoint, I wonder if it stores it outside the instance and can therefore return something even when the instance is stopped. I kinda doubt it, but I’ll check.

david-crespo:
The history endpoint does not work for a stopped instance: it returns an error saying the instance is stopped and has no active serial console server.

On the other points:

* Yes to docs. We don't have one but we should add one (ASAP so we can make sure the link is at least there in the next release even if the content is not all the way where we want it)
* I think it's ok to just link back to the instance (probably include the instance state somewhere on the page) and not get too elaborate with controls. I guess I could imagine putting the same instance actions menu button we have on instance detail on this page, but I don't see why you'd want to do any of those things while looking at the serial console, plus when you hit start we'd have to then poll status and only try to connect when it'll work. All in all, that kind of liveness is something we should probably do eventually but I'd rather not do it now.

david-crespo:
Doing another one with one more change in it.

david-crespo:
Aaaagh whoops

PR #2247, opened 2024-05-15
Remove Badge from IP Pool cell

charliepark:
We just added this column, and I had stuck a Badge component into the new IP Pool column. Crespo pointed out that we typically use badge elements for things like states, where the instance has one of a set of predefined options.

This PR removes the badge, leaving a regular string in its place.

But!

This does create a bit of a dilemma with the tooltip, which houses the `description` of the IP pool. @paryhin, would love your thoughts on the best path forward for this. How should we show the pool's description? I'm open to any approach that makes sense to you.

Screenshot 2024-05-15 at 3 41 08 PM

(Note that we do not have a "view IP pool" option at the project level, just at the system level, which is why this isn't a link to a "view" page.)


david-crespo:
Charlie and I tried both making the whole cell hover and using a different cursor and neither felt quite right (plus the cell hover has a weird bug we couldn't figure out). So I can live with leaving it as-is. 

https://github.com/oxidecomputer/console/assets/3612203/44df89a9-2b09-48a3-9dac-a852cc282dc2

https://github.com/oxidecomputer/console/assets/3612203/911e6853-b664-4693-b8ff-36626c495207

Issue #1913, opened 2024-01-26
Consolidate routing tabs as url params vs dedicated pages


benjaminleonard:
Sometimes for tabs we use params:

`/projects/mock-project/vpcs/mock-vpc?tab=firewall-rules`

and sometimes pages:

`/projects/mock-project/instances/db1/storage`

We should either land on a single approach or document why we use which approach in which context.

david-crespo:
I think we should probably consolidate on route tabs, which have the advantage of letting you nest further routes under them. This would let us, for example, give routes to the firewall rules create and edit forms.

```
$ rg '<QueryParamTabs'
app/pages/system/silos/SiloPage.tsx
76:      <QueryParamTabs className="full-width" defaultValue="idps">

app/pages/system/networking/IpPoolPage.tsx
79:      <QueryParamTabs className="full-width" defaultValue="ranges">

app/pages/system/UtilizationPage.tsx
57:      <QueryParamTabs defaultValue="summary" className="full-width">

app/pages/project/vpcs/VpcPage/VpcPage.tsx
68:      <QueryParamTabs className="full-width" defaultValue="firewall-rules">
```

```
$ rg '<RouteTabs'
app/pages/system/inventory/sled/SledPage.tsx
72:      <RouteTabs fullWidth>

app/pages/system/inventory/InventoryPage.tsx
33:      <RouteTabs fullWidth>

app/pages/project/instances/instance/InstancePage.tsx
198:      <RouteTabs fullWidth>
```

PR #2249, opened 2024-05-17
Switch VPC tab bar to route tabs

benjaminleonard:
Prerequisite to a "New firewall rule from existing" feature so we can pass the initial values in through the loader. 

Also part of #1913


benjaminleonard:
@david-crespo I'm getting "react-router-dom.js?v=0ae1b1b1:215 Matched leaf route at location "/projects/mock-project/vpcs/mock-vpc/firewall-rules" does not have an element or Component. This means it will render an <Outlet /> with a null value by default resulting in an "empty" page." – unsure on how to fix this? I might have `Outlet` in the wrong place.

david-crespo:
Yeah I can definitely figure that out.

david-crespo:
Missing element for leaf route was simple (https://github.com/oxidecomputer/console/pull/2249/commits/373fedaf6f551f1485b527f1371cd164e27b2655): you were missing an `element`. It doesn't have one because it just uses the parent, so you have to say `element={null}`. We do this in a bunch of other spots:

https://github.com/oxidecomputer/console/blob/a228b75ba35952b68c0b8b0892c452d4fc29467a/app/routes.tsx#L182-L195

david-crespo:
Github actions. Why do you do this to me. This is an error that can only happen on main. The change that adds this check is not in this PR.

image

app/forms/firewall-rules-edit.tsx
```diff
@@ -5,36 +5,53 @@
...
   const vpcSelector = useVpcSelector()
   const queryClient = useApiQueryClient()
 
+  const { data } = usePrefetchedApiQuery('vpcFirewallRulesView', {
+    query: { project, vpc },
+  })
+
+  const originalRule = data.rules.find((r) => r.name === rule)
+
+  invariant(originalRule, 'Firewall rule must exist')
```
david-crespo, line 50: This isn't quite right because this is generic server error and we want to 404 here. I will make the loader 404 if the rule isn't there.

Issue #2220, opened 2024-05-03
Creating a new firewall rule with the same name overwrites original


benjaminleonard:
Tested on both the mock API and colo.

A side effect that we work with the entire firewall rules object rather than individual rules in the API. We should prevent the user from creating a new rule with a duplicate name.

This does strike me as another downside to the current API, requiring extra logic on the client like this.

david-crespo:
https://github.com/oxidecomputer/console/blob/a06d8521a1e069a9d2c1a387077fa8c3830c0d4e/app/forms/firewall-rules-create.tsx#L558-L562

PR #2250, opened 2024-05-17
Clone firewall rule

### LINKED ISSUE #2063: Firewall rules: Provide capability to easily set up new rules by copying an existing one

benjaminleonard:
Fixes #2063 inspired by #2067.

Much cleaner compared to that approach since we use `defaultValues` and don't need to set anything on the form directly after its been initialised.

CleanShot 2024-05-17 at 16 34 10

Cherry picked the route param commits from #2249


zephraph:
Clever! (👋)

benjaminleonard:
Come back!

david-crespo:
Love this and it is wonderfully clean. I have some tweaks to make and I'll add an e2e test too. My only issue is "New similar rule" — tolerable for sure but maybe we can do better. I see the difficulty here because you want to convey both of these things at once:

1. It's using the existing rule as a starting point, but you can edit it
2. It only opens the form, it does not actually create anything right off

So 2 rules out the usual words like "copy" or "duplicate". I wonder if "clone" has more of a templatey connotation. `Clone and edit`? `Copy and edit`? Could we have a tooltip on a non-disabled button saying "Create a new rule using this one as a template"? Is that disgusting?

app/util/str.ts
```diff
@@ -65,3 +65,18 @@ export const validateIp = (ip: string) => {
...
+  if (match) {
+    const base = match[1]
+    const num = parseInt(match[2], 10)
+    name = `${base}-${num + 1}`
+  } else {
+    name = `${name}-1`
+  }
+
+  return name
+}
```
charliepark, line 82: nice

charliepark:
One edge-case issue that shouldn't hold this up, as it will be hopefully fixed by another issue (#2220) shortly:
- I'm copying a rule, `firewall-rule`; it's successful
- I now have `firewall-rule` and `firewall-rule-1`
- I copy `firewall-rule` again, without making any other modifications; I save it
- Instead of having `firewall-rule-2`, I now have `firewall-rule-1`, and because of #2220, it has overwritten the first `firewall-rule-1`

Once #2220 is fixed, this will just create an error at the form submission step and the duplicate `firewall-rule-1` won't get created. I think we shouldn't be overly concerned with, say, passing in all firewall rule names to the function that creates the incremented firewall rule name; just wanted to note that edge-case issue.

charliepark (approved):

david-crespo:
We could do a random 3 digit number or something to combat that.

david-crespo:
Ugh these conflicts are not cool!

david-crespo:
@charliepark and I talked the button copy question to death and landed on `Clone` by itself. We decided the user uncertainty the first time about whether it creates it directly is tolerable because in the worst case, it creates an identical rule and you can just delete it. And of course we don't actually do that, we just open the form. `Copy` doesn't work because it suggests something more like copy/paste and we have the `Copy ID` button right above that actually does mean that.

david-crespo:
We decided to go with adding `-copy` to the copied rule's name instead of a number. We will fix https://github.com/oxidecomputer/console/issues/2220 for v10, so it is fine.

PR #2252, opened 2024-05-20
Add floating ips to instance create

### LINKED ISSUE #1099: Floating IPs in the instance create form

### LINKED ISSUE #1979: Assign existing floating IP on instance create form
Probably done alongside https://github.com/oxidecomputer/console/issues/1097 because both are about advanced networking configuration on instance create. Spun off from #1871.

charliepark:
Fixes #1979 

This PR adds the ability to attach a floating IP (or several) to an instance during the creation process.

add_floating_ip

There are a few slight differences from the design in Figma …
Screenshot 2024-05-20 at 11 07 02 AM
… namely the dropping of the checkbox, though this brings it into alignment with the "attach disks" flow. Essentially, the button triggers the modal, which allows the user to attach floating IPs. If they remove all of the floating IPs, none will be attached.


app/forms/instance-create.tsx
```diff
@@ -669,6 +747,131 @@ const AdvancedAccordion = ({
...
+                      }
+                    />
+                  )}
+                  <form>
+                    <Listbox
+                      name="floatingIp"
+                      items={availableFloatingIps.map((i) => ({
+                        value: i.name,
+                        label: (
+                          <>
```
david-crespo, line 831: fragment unnecessary because there's only one div in here. I guess if it works with just the fragment, might as well do that

charliepark:
Added an empty state for when the project has no floating IPs:
Screenshot 2024-05-22 at 9 03 48 AM

Then, when floating IPs exist and are available to be attached:
Screenshot 2024-05-22 at 9 04 08 AM

Then, when they've been committed to the instance (and will be attached when the instance is created):
Screenshot 2024-05-22 at 9 04 25 AM

benjaminleonard:
I like that as a pattern. We might want to explain why no floating IPs are found though; either none exist or all the ones that do are already attached to an instance right?

david-crespo:
We could add/are maybe planning to add a button that lets you create a floating IP right there, and that would cover it. The biggest issue for me is making clear which operations actually create stuff (e.g., floating IP, SSH key) and which are queuing things up to be created as part of instance create (e.g., disks). I wonder if the visual treatment could convey that somehow, or we can just put it in an info box at the top of the forms.

app/forms/instance-create.tsx
```diff
@@ -669,6 +748,142 @@ const AdvancedAccordion = ({
...
+              onDismiss={removeFloatingIpPlaceholder}
+              title="Attach floating IP"
+            >
+              <Modal.Body>
+                <Modal.Section>
+                  <Message
+                    variant="info"
+                    content={
+                      <>
+                        {'This instance will be reachable at '}
```
david-crespo, line 825: something like this
```suggestion
                        This instance will be reachable at{' '}
```

app/forms/instance-create.tsx
```diff
@@ -669,6 +748,142 @@ const AdvancedAccordion = ({
...
+                      >
+                        <Error16Icon title={`remove ${item.name}`} />
+                      </button>
+                    </MiniTable.Cell>
+                  </MiniTable.Row>
+                ))}
+              </MiniTable.Body>
+            </MiniTable.Table>
+          )}
+          {floatingIpList.items.length === 0 ? (
```
david-crespo, line 791: Should this be available IPs? You still can't do anything if floating IPs existing but are taken.

app/forms/instance-create.tsx
```diff
@@ -669,6 +748,142 @@ const AdvancedAccordion = ({
...
+                    variant="info"
+                    content={
+                      <>
+                        This instance will be reachable at{' '}
+                        {selectedFloatingIp ? (
+                          <HL>
+                            {
+                              attachableFloatingIps.find(
+                                (ip) => ip.name === selectedFloatingIp
+                              )?.ip
```
david-crespo, line 831: Indentation is a bit out of hand here. Might help to pull this out. Try a few things like that.

app/forms/instance-create.tsx
charliepark: The idea here is that if the list of floating IPs coming in from the API query has 0 items, we show them the empty message; if there are any items in that list at all (attached or not), we show them the `Attach floating IP` button. That button then is either disabled (no available floating IPs) or enabled (at least one available floating IP). But the disabled message on the button would indicate that no floating IPs are available. Let me know if that didn't make sense, or if there's an alternate flow you would prefer to see there?

david-crespo: I think what felt funny to me is that within this form, I'm not sure the distinction between "there are no floating IPs at all" and "there are floating IPs but they're all taken" is meaningful to the user. We are treating it as meaningful by representing those two states differently. Here are all the states:

| State | UI representation |
| --- | --- |
|  no floating IPs at all | empty state like the tables |
| floating IPs exist but are attached to other instances | disabled attach button |
| floating IPs exist, none available, 1+ attached to this instance | mini table + disabled attach button |
| floating IPs available, 1+ attached to this instance | mini table + enabled attach button |
| floating IPs available, none attached | enabled attach button |

I'm wondering if we should give the first two rows the same representation. This can't matter that much because the user is never going to see both versions next to each other and wonder why they're different, but I'm finding it helpful to think through it.

app/forms/instance-create.tsx
```diff
@@ -669,6 +758,124 @@ const AdvancedAccordion = ({
             />
           )}
         </div>
+
+        <div className="flex flex-1 flex-col gap-4">
+          <h2 className="text-sans-md">
+            Floating IPs{' '}
+            <TipIcon>
+              Floating IPs are static IP addresses that can be attached to instances
```
david-crespo, line 766: Might as well use the line we suffered for in the docs popover:

https://github.com/oxidecomputer/console/blob/9c519ce7ad81ae86a79bcf86f5293d779242f22f/app/pages/project/floating-ips/FloatingIpsPage.tsx?plain=1#L224

app/forms/instance-create.tsx
charliepark: If we consolidated them, which of the two treatments feels better to you? Just the disabled button (with no empty state)? That would certainly be less "involved" on an already pretty-lengthy page. Though the empty state does call your eye to it, and for something as foundational as a floating IP, I could see it making sense to have it called out like that.

david-crespo: I agree there is something nice about the full empty state almost inviting the user to create one. @benjaminleonard since you weighed in before, what do you think?

app/ui/lib/ListboxLabels.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <span>Created on {toLocaleDateString(snapshot.timeCreated)}</span>
+        <DiskNameFromId disk={snapshot.diskId} />
+        <Slash />
+        <span>
+          {formattedSize.value} {formattedSize.unit}
+        </span>
+      </div>
+    </>
+  )
+}
```
david-crespo, line 118: As far as I can tell, each of these is used in one place. Do you think we'll use them in more places?

app/forms/instance-create.tsx
```diff
@@ -573,24 +588,97 @@ export function CreateInstanceForm() {
...
+    .filter((x) => !!x) as Array<FloatingIp>
+
+  const addFloatingIpPlaceholder = () => {
+    setSelectedFloatingIp(undefined)
+    const newExternalIps = [
+      ...(externalIps.field.value || []),
+      { type: 'floating', floatingIp: '' },
+    ]
+    externalIps.field.onChange(newExternalIps)
+  }
```
david-crespo, line 640: I think using a dedicated boolean state var to tell whether the modal should be open would be a lot easier to understand and cuts some lines too:

```
const [floatingIpModalOpen, setFloatingIpModalOpen] = useState(false)
```

Opening and closing the modal is just a matter of setting that true or false, and then you don't have this weird half floating ip in there to clean up.

app/forms/instance-create.tsx
```diff
@@ -573,24 +588,97 @@ export function CreateInstanceForm() {
...
+
+  const attachFloatingIp = (name: string) => {
+    const newExternalIps = externalIps.field.value?.map((ip) => {
+      if (ip.type === 'floating' && ip.floatingIp === '') {
+        return { type: 'floating', floatingIp: name }
+      }
+      return ip
+    })
+    externalIps.field.onChange(newExternalIps)
+  }
```
david-crespo, line 658: This also becomes simpler without the placeholder concept: you just add `{ type: 'floating', floatingIp: name }` to the list.

app/forms/instance-create.tsx
```diff
@@ -669,6 +752,110 @@ const AdvancedAccordion = ({
...
+                disabled={availableFloatingIps.length === 0}
+                disabledReason="No floating IPs available"
+                onClick={addFloatingIpPlaceholder}
+              >
+                Attach floating IP
+              </Button>
+            </div>
+          )}
+
+          {!!attachedFloatingIps.length && (
```
david-crespo, line 818: not 100% sure on this, but I think because of the `isOpen` prop, you don't have to conditionally render the modal. you can just render it all the time and rely on isOpen

app/forms/instance-create.tsx
```diff
@@ -573,24 +588,97 @@ export function CreateInstanceForm() {
...
+  const detachFloatingIp = (name: string) => {
+    const newExternalIps = externalIps.field.value?.filter(
+      (ip) => ip.type === 'ephemeral' || ip.floatingIp !== name
+    )
+    externalIps.field.onChange(newExternalIps)
+  }
+
+  const isEmptyFloatingIpPlaceholderPresent = attachedFloatingIps.some(
+    (ip) => ip.floatingIp === ''
+  )
```
david-crespo, line 669: this can disappear if `floatingIpModalOpen` is just a state bool

app/forms/instance-create.tsx
charliepark: Good call; saved 21 lines / 10% of the PR.

app/forms/instance-create.tsx
```diff
@@ -573,24 +589,95 @@ export function CreateInstanceForm() {
...
+      ...(externalIps.field.value || []),
+      { type: 'floating', floatingIp: selectedFloatingIp?.name },
+    ])
+    closeFloatingIpModal()
+  }
+
+  const detachFloatingIp = (name: string) => {
+    externalIps.field.onChange(
+      externalIps.field.value?.filter(
+        (ip) => ip.type === 'ephemeral' || ip.floatingIp !== name
```
david-crespo, line 668: I find this version easier to understand, but I thought of a non-subjective argument for it too: if a third type is added, the current version will break and my suggestion will not.

```suggestion
        (ip) => !(ip.type === 'floating' && ip.floatingIp === name)
```

charliepark: Nice.

app/forms/instance-create.tsx
```diff
@@ -573,24 +589,95 @@ export function CreateInstanceForm() {
...
+
+  const closeFloatingIpModal = () => {
+    setFloatingIpModalOpen(false)
+    setSelectedFloatingIp(undefined)
+  }
+
+  const attachFloatingIp = () => {
+    externalIps.field.onChange([
+      ...(externalIps.field.value || []),
+      { type: 'floating', floatingIp: selectedFloatingIp?.name },
```
david-crespo, line 660: I was surprised this typechecks because `floatingIp` is a required string, but the value here could be undefined (a possibility we can eliminate by wrapping the inside of the function in `if (selectedFloatingIp) { }`).

image

It looks like `onChange` is not intelligently typed even though `field` is:

image


image

That's kind of shitty! It' s not generic either, so it doesn't seem like there's anything we can do about it:

https://github.com/react-hook-form/react-hook-form/blob/927535bf87412b2baa91faefb4c7ded327126c2b/src/types/controller.ts#L27

mock-api/msw/handlers.ts
```diff
@@ -486,6 +486,33 @@ export const handlers = makeHandlers({
...
+        const floatingIp = lookup.floatingIp({
+          project: project.id,
+          floatingIp: ip.floating_ip,
+        })
+        if (floatingIp.instance_id) {
+          throw 'floating IP cannot be attached to one instance while still attached to another'
+        }
+        floatingIp.instance_id = instanceId
+      } else if (ip.type === 'ephemeral') {
+        const poolId = db.ipPools.find((pool) => pool.name === ip.pool)?.id
```
david-crespo, line 500: use `lookup.ipPool`. the result is non-nullable because it throws the 404 if it doesn't find it

mock-api/msw/handlers.ts
```diff
@@ -486,6 +486,33 @@ export const handlers = makeHandlers({
...
+        const range = db.ipPoolRanges.find((range) => range.ip_pool_id === poolId)?.range
```
david-crespo, line 501: If the pool has no ranges you should 400. I think that's what the API does (it may 404 instead of 400, not sure).

I think it would make sense to wrap these three lines in a util function `getIpFromPool` that takes the pool `NameOrId` just like the lookup function. The functionality won't be right but we can use the function everywhere currently allocate an IP (don't make that change now unless it's like two spots) and then we can get the logic right later. One interesting question is whether this function should 400 if there are no ranges or it should just return `null` and the calling code decides what kind of error that is. Different call sites might want to handle it differently? I don't know. There's also the problem of distinguishing between not getting an IP because the pool doesn't exist, not getting an IP because there are no ranges, and not getting an IP because the pool is exhausted. Returning null from the function wouldn't be sufficient for that. I'm also not 100% sure what the 4xx status code from the API is in each of those cases.

app/forms/instance-create.tsx
```diff
@@ -573,24 +589,97 @@ export function CreateInstanceForm() {
...
+    [floatingIpList]
+  )
+
+  // To find available floating IPs, we remove the ones that are already committed to this instance
+  const availableFloatingIps = attachableFloatingIps.filter(
+    (ip) => !attachedFloatingIps.find((attachedIp) => attachedIp.floatingIp === ip.name)
+  )
+  const attachedFloatingIpsData = attachedFloatingIps
+    .map((ip) => attachableFloatingIps.find((fip) => fip.name === ip.floatingIp))
+    .filter((x) => !!x) as Array<FloatingIp>
```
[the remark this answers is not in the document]
charliepark: wild that that works, where the earlier version needed casting; thank you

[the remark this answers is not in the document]
david-crespo: `isTruthy` is just a guard like the `isFloating` thing we wrote. it's interesting, it just asserts that whatever it is, it is not falsy

david-crespo:
I just accidentally found an incredible and strange behavior that is probably not caused by anything specific to this PR, and it does it in both FF and Chrome:

1. Go to instance create [preview deploy](https://console-git-addfloatingipstoinstancecreate-oxidecomputer.vercel.app/projects/mock-project/instances-new)
2. Do NOT enter an instance name. scroll down and attach the one available floating IP
3. Hit submit. You get an error because name is empty and it takes you to the name field.
4. Type in any name
5. Press enter

Instead of submitting the form, you just removed the floating IP because the delete button was focused. WTF! Probably an easy fix, but how exciting.

app/forms/instance-create.tsx
```diff
@@ -669,6 +757,106 @@ const AdvancedAccordion = ({
...
+                    key={item.name}
+                  >
+                    <MiniTable.Cell>{item.name}</MiniTable.Cell>
+                    <MiniTable.Cell>{item.ip}</MiniTable.Cell>
+                    <MiniTable.Cell>
+                      <button
+                        onClick={() => {
+                          detachFloatingIp(item.name)
+                        }}
+                      >
```
david-crespo, line 792: OMG. Found the culprit. The default type of `button` is `submit`, believe it or not. Which is ridiculous. Adding `type="button"` fixes it. I'm going to see if there's a lint rule we can turn on for this, requiring a type on all buttons.

```suggestion
                      <button type="button" onClick={() => detachFloatingIp(item.name)}>

```

david-crespo (approved):
I think this is ready to go. Only thing I can think of to add is a couple of lines in the e2e test to test that remove is hooked up.

charliepark:
Re-running that failing test, as I don't think it's related, but we'll need to look into it if it keeps flaking.

david-crespo:
That flake was weird — it looked a lot like one I got on the button type branch, where it wasn’t a flake. I noticed another case today where state seemed to be leaking between GH actions runs — a lint rule failed on a branch that should not have that rule on.

PR #2256, opened 2024-05-22
Dim plus sign on disabled button

### LINKED ISSUE #2255: Plus in create button needs to fade when disabled
Should be a one-liner.

image

charliepark:
Fixes #2255

With this PR, we're adding a ternary to the CSS display logic; if the button is disabled, we use the `-disabled` color; otherwise, we use the `-secondary` color.

In this screenshot, the top button is enabled; the bottom is disabled.
Screenshot 2024-05-21 at 5 22 03 PM

### DIFF
--- a/app/ui/lib/CreateButton.tsx
+++ b/app/ui/lib/CreateButton.tsx
@@ -13,10 +13,12 @@ import { AddRoundel12Icon } from '@oxide/design-system/icons/react'
 
 import { Button, buttonStyle, type ButtonProps } from '~/ui/lib/Button'
 
-export const CreateButton = ({ children, disabled, ...props }: ButtonProps) => (
-  <Button size="sm" className="shrink-0" disabled={disabled} {...props}>
+export const CreateButton = ({ children, ...props }: ButtonProps) => (
+  <Button size="sm" className="shrink-0" {...props}>
     <AddRoundel12Icon
-      className={cn(disabled ? 'text-accent-disabled' : 'text-accent-secondary', 'mr-2')}
+      // dim the icon color from the default (text accent) because it looks a
+      // lot brighter than text, but default disabled color is fine
+      className={cn('mr-2', { 'text-accent-secondary': !props.disabled })}
     />
     {children}
   </Button>
--- a/app/ui/lib/CreateButton.tsx
+++ b/app/ui/lib/CreateButton.tsx
@@ -6,15 +6,18 @@
  * Copyright Oxide Computer Company
  */
 
+import cn from 'classnames'
 import { Link, type LinkProps } from 'react-router-dom'
 
 import { AddRoundel12Icon } from '@oxide/design-system/icons/react'
 
 import { Button, buttonStyle, type ButtonProps } from '~/ui/lib/Button'
 
-export const CreateButton = ({ children, ...props }: ButtonProps) => (
-  <Button size="sm" className="shrink-0" {...props}>
-    <AddRoundel12Icon className="mr-2 text-accent-secondary" />
+export const CreateButton = ({ children, disabled, ...props }: ButtonProps) => (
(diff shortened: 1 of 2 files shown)


app/ui/lib/CreateButton.tsx
```diff
@@ -6,15 +6,18 @@
...
 
 import { AddRoundel12Icon } from '@oxide/design-system/icons/react'
 
 import { Button, buttonStyle, type ButtonProps } from '~/ui/lib/Button'
 
-export const CreateButton = ({ children, ...props }: ButtonProps) => (
-  <Button size="sm" className="shrink-0" {...props}>
-    <AddRoundel12Icon className="mr-2 text-accent-secondary" />
+export const CreateButton = ({ children, disabled, ...props }: ButtonProps) => (
+  <Button size="sm" className="shrink-0" disabled={disabled} {...props}>
```
charliepark, line 17: Because we're now pulling the `disabled` prop out of the `...props` spread on line 16, we need to add it in as an explicit prop in the `Button` component on 17.

test/e2e/instance-networking.e2e.ts
```diff
@@ -35,9 +35,13 @@ test('Instance networking tab — NIC table', async ({ page }) => {
...
   )
 
+  // Ensure that the 'Add network interface' button is disabled when the instance is running
+  await expect(page.getByRole('button', { name: 'Add network interface' })).toBeDisabled()
+
   // Have to stop instance to edit NICs
   await stopInstance(page)
 
+  await expect(page.getByRole('button', { name: 'Add network interface' })).toBeEnabled()
   await page.click('role=button[name="Add network interface"]')
```
david-crespo, line 45: Line 45 is also the same selector, just done as a legacy string. might as pull out `const addNicButton` to use in all three spots.

app/ui/lib/CreateButton.tsx
david-crespo: Another way would be to not pull it out and use `props.disabled` on line 19. either way is fine

david-crespo:
I went to go remind myself why we have to set this color at all instead of relying on the default text color of the button — the problem was that the default text accent looks too bright next to the text when used to fill a nearly solid icon (Eugene has this in the design, we didn't make it up ourselves). 


| overridden                         | default                         |
|----------------------------------|----------------------------------|
| Screenshot 2024-05-22 at 10 43 24 AM  | Screenshot 2024-05-22 at 10 43 37 AM  |


However, the disabled text color for the button is `text-accent-disabled` anyway, so we can get away with only overriding the color when it's not disabled.

```diff
diff --git a/app/ui/lib/CreateButton.tsx b/app/ui/lib/CreateButton.tsx
index 989efeef..7f6e9431 100644
--- a/app/ui/lib/CreateButton.tsx
+++ b/app/ui/lib/CreateButton.tsx
@@ -13,10 +13,12 @@ import { AddRoundel12Icon } from '@oxide/design-system/icons/react'
 
 import { Button, buttonStyle, type ButtonProps } from '~/ui/lib/Button'
 
-export const CreateButton = ({ children, disabled, ...props }: ButtonProps) => (
-  <Button size="sm" className="shrink-0" disabled={disabled} {...props}>
+export const CreateButton = ({ children, ...props }: ButtonProps) => (
+  <Button size="sm" className="shrink-0" {...props}>
     <AddRoundel12Icon
-      className={cn(disabled ? 'text-accent-disabled' : 'text-accent-secondary', 'mr-2')}
+      // dim the icon color from the default (text accent) because it looks a
+      // lot brighter than text, but default disabled color is fine
+      className={cn('mr-2', { 'text-accent-secondary': !props.disabled })}
     />
     {children}
   </Button>
```

david-crespo (approved):

PR #2257, opened 2024-05-23
Filter out already-attached disks from instance create and edit views

charliepark:
Not ready for review yet; just saving this as a draft PR so I can write down some notes and come back to it.

An enhancement I haven't made here yet is that we can disable the "attach existing disk" button, but to do that we'll need to move the API query higher in the stack, and then pass in the detached-and-available disks to the list component. If the detached-and-available count is 0, we'll disable the button. I'll take a look at that in a little bit.


david-crespo:
Note you could avoid moving the API call up by passing in the list of attached disks to the side modal and letting it handle removing them from the list. Not sure which is better!

PR #2261, opened 2024-05-24
Turn on react/button-has-type lint rule and fix all

### LINKED ISSUE #2260: create RemoveButton component
We discovered recently that we have some `<button>` elements that don't pass in a `type="button"` attribute/value. We *do* add in `type="button"` as a default prop in the `<Button>` component, but there are times where we won't want to add on the style of a `<Button>`.

We have a few places in the MiniTable rows where we have code like this:

```
<MiniTable.Cell>
    <button onClick={() => onChange(items.filter((i) => i.name !== item.name))}>
        <Error16Icon title={`remove ${item.name}`} />
    </button>
</MiniTable.Cell>
```

Adding a `RemoveButton` component that receives an `onClick` and `title` prop, but that adds `type="button"` automatically so we don't have to think about it.

david-crespo:
Discovered because of a delightful bug in https://github.com/oxidecomputer/console/pull/2252#issuecomment-2130334790. This actually fixes an existing bug on the instance create form where if you press enter to submit the form while in the name field, for example, it will remove an attached disk.

Closes #2260 too.

### Bug demo

When the disk disappears, that's because I've pressed enter with a random text input focused.

https://github.com/oxidecomputer/console/assets/3612203/0d8416af-858f-45c7-9314-ba161889d792

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -86,6 +86,7 @@ module.exports = {
     radix: 'error',
     'react-hooks/exhaustive-deps': 'error',
     'react-hooks/rules-of-hooks': 'error',
+    'react/button-has-type': 'error',
     'react/jsx-boolean-value': 'error',
     'react/display-name': 'off',
     'react/react-in-jsx-scope': 'off',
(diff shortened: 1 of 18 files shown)


charliepark (approved):

charliepark:
a fun find

benjaminleonard:
Such a useful rule. Wish there was an easy way to share the lint rules we use here to the others without doing it manually.

PR #2262, opened 2024-05-24
Filter out committed disks

### LINKED ISSUE #2254: already-attached disks are listed in disks-to-attach modal in instance create form
In the instance-create form, we give the user the ability to attach existing disks. We should make sure that any disks that have been already selected for attachment aren't available to be selected again. Currently, they aren't filtered out, so can be selected multiple times.
Screenshot 2024-05-21 at 10 41 47 AM

charliepark:
Fixes #2254 

This PR filters out the already-"attached" disks from the "attach an additional disk" form that shows up during instance creation.

This is a cleaner version of what I was trying in #2257. We pass the committed disks to the form and filter them there.

### DIFF
--- a/app/forms/disk-attach.tsx
+++ b/app/forms/disk-attach.tsx
@@ -17,7 +17,7 @@ type AttachDiskProps = {
   /** If defined, this overrides the usual mutation */
   onSubmit: (diskAttach: { name: string }) => void
   onDismiss: () => void
-  disksToFilter?: string[]
+  diskNamesToExclude?: string[]
   loading?: boolean
   submitError?: ApiError | null
 }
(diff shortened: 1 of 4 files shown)


app/forms/disk-attach.tsx
```diff
@@ -28,6 +29,7 @@ type AttachDiskProps = {
 export function AttachDiskSideModalForm({
   onSubmit,
   onDismiss,
+  disksToFilter,
```
david-crespo, line 32: I would give this a default value of `[]` so you don't have to do the `?` on 44 and rename it to `disksToExclude` to avoid the ambiguity of filter in vs. filter out. Maybe even go as far as `diskNamesToExclude` as a hint to future callers.

david-crespo (approved):
looks great and the test is great, just the rename

PR #2267, opened 2024-06-03
Combobox

### LINKED ISSUE #1942: Need combobox component + field
Let's give Headless's combobox a try I guess.

https://headlessui.com/react/combobox

charliepark:
This uses the Headless combobox component ([here](https://headlessui.com/react/combobox)), making a few tweaks to align with our current dropdowns.

Fixes #1942 

Screenshot 2024-06-03 at 3 23 50 PM

Also added a few tests to check for combobox interactivity.


david-crespo:
image

I made some tweaks:

- Clean up props
- Use `match-sorter` for fuzzy filtering (turn off sorting, though, to keep original order — that feels less jarring to me, but could go either way)
- Tweak borders on options — there was a double border between the options due to each option having borders on all sides
- Tweak `onChange` so that when you clear the field and blur it stays empty

so this is pretty close now. The remaining things I would like to do are:

- [ ] More intensive testing on typing and clearing and blurring and filtering
  - Not sure whether we need to do this through Playwright — on one hand it feels like it should be a more unit-y test, on the other we don't really do those anymore because they are annoying to write
- [ ] Try to make `virtual` prop work so it doesn't chug when there are 1000 items
  - It's currently usable in Firefox but unusable in Chrome, and most people will probably be using Chrome on computers slower than an M1 Mac
- [ ] Convert a few other `ListboxField` to `ComboboxField` and see how they feel

charliepark:
In migrating a few more Listboxes over to Comboboxes, I discovered a few props that were missing that I've added, like for a disabled state.

It also appears that the ComboboxOption, which currently has a value: string and a label: string, isn't as flexible as the existing ListboxOption, on two accounts. First, Listboxes allow for rendered HTML elements, meaning we can use richer designs for the dropdowns. Second, all of our existing implementations have an item where the label and the value are identical. We'll have occasions — especially with those rendered HTML elements — where the selected value (a string) will be different from the label (a string, or a ReactNode).

This PR has been going long enough that I want to treat those other situations separately.

david-crespo (approved):
Very nice. Let's do the followup work in followups and get this on dogfood ASAP so people can try it.

- More tests
- `virtual`
- Enable fancier option rendering
- Think about when to use listbox vs. combobox — I figure listbox is simpler for things where we expect there to be a small number of times, but we need a clearer standard

PR #2269, opened 2024-06-04
Use Remeda (lodash-like lib)

david-crespo:
This is kind of a palate cleanser, nothing important. I saw this lovely library https://remedajs.com/, maintained by @tkdodo, so I tried it and it's very nice. We have some little utilities around arrays and objects and Remeda has drop-in replacements for them. The advantage of using Remeda is

1. We can delete our tests for these utils because we know remeda is legit
2. Anything else we need is just there
3. Many of remeda's functions are curried and optimized for lazy evaluation as part of a pipeline, which means a `filter().map().filter()` chain doesn't have to loop through the list three times.

This PR makes our bundle only 300 bytes bigger after gzip, and that will probably go away after we get rid of the util functions completely instead of wrapping Remeda.

### DIFF
--- a/app/api/roles.ts
+++ b/app/api/roles.ts
@@ -12,8 +12,7 @@
  * it belongs in the API proper.
  */
 import { useMemo } from 'react'
-
-import { lowestBy, sortBy } from '~/util/array'
+import * as R from 'remeda'
 
 import type { FleetRole, IdentityType, ProjectRole, SiloRole } from './__generated__/Api'
 import { usePrefetchedApiQuery } from './client'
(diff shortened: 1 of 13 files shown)


app/api/roles.ts
```diff
@@ -12,8 +12,7 @@
  * it belongs in the API proper.
  */
 import { useMemo } from 'react'
-
-import { lowestBy, sortBy } from '~/util/array'
+import * as R from 'remeda'
```
david-crespo, line 15: It turns out this tree-shakes perfectly well, so we don't need to bother with any manual `import { sortBy } from 'remeda'`.

PR #2284, opened 2024-06-21
Enable "reserved names" on Name field as a new, optional, validation step

### LINKED ISSUE #2220: Creating a new firewall rule with the same name overwrites original

charliepark:
Fixes #2220 

This PR adds a validation to the NameField element, where an array of reserved names can be passed in. If the form has this optional prop, and if the user tries to pass in a name to the form that matches a string in the array, it will fail the validation check.

This does not affect the _update_ form for firewall rules.

Screenshot 2024-06-21 at 2 19 16 PM

### DIFF
--- a/app/components/form/fields/NameField.spec.tsx
+++ b/app/components/form/fields/NameField.spec.tsx
@@ -10,7 +10,7 @@ import { describe, expect, it } from 'vitest'
 import { validateName } from './NameField'
 
 describe('validateName', () => {
-  const validate = (name: string) => validateName(name, 'Name', true)
+  const validate = (name: string) => validateName(name, 'Name', true, ['example'])
 
   it('returns undefined for valid names', () => {
     expect(validate('abc')).toBeUndefined()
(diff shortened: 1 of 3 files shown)


david-crespo:
Documenting what we talked about IRL: the firewall rule update endpoint is unusual in that it will not complain if we reuse a name because you need to be able to overwrite an existing rule with a new one. If we changed the API to be more like our other endpoints, where you add and remove one by one, it would be different, and we could rely on the API to tell us when a name is already in use.

I do not expect to need this reserved names concept for other endpoints because they will 400 if the name is in use, and there is no risk of overwriting an existing resource by trying to create a new one with the same name. So I think we should fix https://github.com/oxidecomputer/console/issues/2220 by doing something more narrowly scoped to the firewall rules form logic, like fixing this TODO:

https://github.com/oxidecomputer/console/blob/a06d8521a1e069a9d2c1a387077fa8c3830c0d4e/app/forms/firewall-rules-create.tsx#L558-L562

PR #2285, opened 2024-06-21
Add detach button for ephemeral ip

### LINKED ISSUE #2205: Instance networking: add detach button for ephemeral IP
https://docs.oxide.computer/api/instance_ephemeral_ip_detach

charliepark:
Fixes #2205 

This adds a `Detach` button for ephemeral IPs, with a confirmation:
Screenshot 2024-06-21 at 5 35 23 PM

Now that we can detach ephemeral IPs, we also need an empty state for that table, so this PR also adds that:
Screenshot 2024-06-21 at 5 32 12 PM

I believe we don't currently have a way to attach an ephemeral IP to an instance in the UI, but I suspect that's something we'll tackle separately.

### DIFF
--- a/app/pages/project/instances/instance/tabs/NetworkingTab.tsx
+++ b/app/pages/project/instances/instance/tabs/NetworkingTab.tsx
@@ -255,6 +255,16 @@ export function NetworkingTab() {
     }),
   ]
 
+  const ephemeralIpDetach = useApiMutation('instanceEphemeralIpDetach', {
+    onSuccess() {
+      queryClient.invalidateQueries('instanceExternalIpList')
+      addToast({ content: 'Your ephemeral IP has been detached' })
+    },
+    onError: (err) => {
+      addToast({ title: 'Error', content: err.message, variant: 'error' })
+    },
+  })
+
   const floatingIpDetach = useApiMutation('floatingIpDetach', {
     onSuccess() {
       queryClient.invalidateQueries('floatingIpList')
@@ -300,10 +310,35 @@ export function NetworkingTab() {
               }),
           },
         ]
+      } else {
+        return [
+          copyAction,
+          {
+            label: 'Detach',
+            onActivate: () =>
+              confirmAction({
+                actionType: 'danger',
+                doAction: () =>
+                  ephemeralIpDetach.mutateAsync({
+                    path: { instance: instanceName },
+                    query: { project },
+                  }),
+                modalTitle: 'Detach Ephemeral IP',
+                modalContent: (
+                  <p>
+                    Are you sure you want to detach ephemeral IP <HL>{externalIp.ip}</HL>{' '}
+                    from <HL>{instanceName}</HL>? The instance will no longer be reachable
+                    at <HL>{externalIp.ip}</HL>.
+                  </p>
+                ),
+                errorTitle: 'Error detaching ephemeral IP',
+              }),
+          },
+        ]
       }
       return [copyAction]
     },
-    [floatingIpDetach, instanceName, project]
+    [ephemeralIpDetach, floatingIpDetach, instanceName, project]
   )
 
   const ipTableInstance = useReactTable({
--- a/app/pages/project/instances/instance/tabs/NetworkingTab.tsx
+++ b/app/pages/project/instances/instance/tabs/NetworkingTab.tsx
@@ -19,7 +19,7 @@ import {
   type ExternalIp,
   type InstanceNetworkInterface,
 } from '@oxide/api'
-import { Networking24Icon } from '@oxide/design-system/icons/react'
+import { IpGlobal24Icon, Networking24Icon } from '@oxide/design-system/icons/react'
 
 import { HL } from '~/components/HL'
 import { CreateNetworkInterfaceForm } from '~/forms/network-interface-create'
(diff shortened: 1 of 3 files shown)


david-crespo:
Looks good! I don’t think it should say the IP twice. Agree about doing attach in a separate PR. It should be really simple — all you can do is pick the pool.

charliepark:
Screenshot 2024-06-22 at 9 45 53 AM

Screenshot 2024-06-22 at 9 45 59 AM

paryhin:
Let's make sure the title and button text match. We have two options:

- Keep the title as is and change the button to "Detach"
- Change the title to "Confirm detach..." and keep the "Confirm" button

charliepark:
Screenshot 2024-06-24 at 1 09 34 PM

david-crespo:
Nitpick, but I have a slight preference for "Confirm detach floating IP" even though it's grammatically dubious. I think of it as an abbreviation for "Confirm (you want to) detach (the) floating IP". It fits a pattern we follow elsewhere:

image

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -306,7 +306,7 @@ export function NetworkingTab() {
             confirmAction({
               actionType: 'danger',
               doAction,
-              modalTitle: `Confirm detachment of ${externalIp.kind} IP`,
+              modalTitle: `Confirm detach of ${externalIp.kind} IP`,
```
david-crespo, line 309: don't need `of` either!

charliepark: oh boo I thought I searched for it

david-crespo:
I vaguely remember that for the detach floating IP menu button, we went with non-destructive styling because the floating IP continues to exist after being detached. In this case, the ephemeral IP is no longer allocated and there's no way to guarantee you can get it back, so it is destructive. I wonder if there's a way to communicate that here. I think making this one red and the floating IP one not would be kind of weird, and it also wouldn't be sufficient to explain what's going on. I can't think of a great wording, but maybe a second sentence in this popup would help, like "Ephemeral IPs are deallocated when they are detached." Deallocated is not a user-meaningful term, really. It would be nice if we had a way to convey that that are Gone in some stronger sense, but not _that_ strong, because the IP might well still be available in the pool.

image

mock-api/msw/handlers.ts
```diff
@@ -561,6 +561,13 @@ export const handlers = makeHandlers({
     disk.state = { state: 'detached' }
     return disk
   },
+  instanceEphemeralIpDetach({ path, query }) {
+    const instance = lookup.instance({ ...path, ...query })
+    const ip = db.ephemeralIps.find((eip) => eip.instance_id === instance.id)
```
david-crespo, line 566: I think an instance can have more than one ephemeral IP, which means you need to match on both the instance ID and the `ip`. This currently works because there's only one.

charliepark: It's possible the docs are incorrect, but https://docs.oxide.computer/guides/managing-instances#_attach_or_detach_external_ip_addresses currently says "An instance may have at most one ephemeral IP, and at most 32 total external IPs." Happy to make this bit of code more robust just in case they end up in a strange state with multiple ephemeral IPs, or in case >1 is legit.

david-crespo: Nope, that's correct. Max = 1. I do think you should still match on the IP, though.

https://github.com/oxidecomputer/omicron/blob/895f2800ad96b51c538ae8011df3654d511563c5/nexus/src/app/mod.rs#L104

charliepark: We talked through this and checked the API, and determined that it was best to match the API logic, to simply detach the first ephemeral IP found (which should be the only ephemeral IP present).

test/e2e/instance-networking.e2e.ts
```diff
@@ -122,4 +122,9 @@ test('Instance networking tab — External IPs', async ({ page }) => {
 
   // And that button shouldbe enabled again
   await expect(page.getByRole('button', { name: 'Attach floating IP' })).toBeEnabled()
+
+  // Detach the ephemeral IP
+  await clickRowAction(page, 'ephemeral', 'Detach')
+  await page.getByRole('button', { name: 'Confirm' }).click()
+  await expect(externalIpTable.getByRole('cell', { name: 'ephemeral' })).toBeHidden()
```
david-crespo, line 129: add an assert before clicking the row action that this same selector is visible. The `clickRowAction` implicitly tests for that, but I like to use the same selector for before-and-after checks.

mock-api/msw/handlers.ts
```diff
@@ -567,7 +567,7 @@ export const handlers = makeHandlers({
     // https://github.com/oxidecomputer/omicron/blob/d52aad0/nexus/db-queries/src/db/datastore/external_ip.rs#L782-L794
     // https://github.com/oxidecomputer/omicron/blob/d52aad0/nexus/src/app/sagas/instance_ip_detach.rs#L79-L82
     const ip = db.ephemeralIps.find((eip) => eip.instance_id === instance.id)
-    if (!ip) throw notFoundErr('ephemeral IP')
+    if (!ip) throw notFoundErr('Could not find an ephemeral IP')
```
david-crespo, line 570: You had it right the first time. See this line:

https://github.com/oxidecomputer/console/blob/c0f9f5e78f8bf73cfce6306eb73b127129230d42/mock-api/msw/db.ts#L25

david-crespo (approved):
great!

mock-api/msw/handlers.ts
charliepark: Ah, womp; ty

PR #2286, opened 2024-06-24
More firewall rule form help copy

### LINKED ISSUE #2263: Add help text on firewall rule form explaining what a host filter does
The host is the source of the traffic, so a host filter naming an instance limits where traffic can come from, not what instance it can go to.

charliepark:
Fixes #2263 

Thanks @FelixMcFelix for the consultation. The host filter message is the main addition, but there are also small tweaks to the other messages.

Screenshot 2024-06-26 at 4 22 52 PM

Screenshot 2024-06-26 at 4 23 01 PM

Screenshot 2024-06-26 at 4 23 05 PM


david-crespo:
I find the wording a little unsatisfying, let's mull it over

FelixMcFelix (approved):
I think these tooltips look great! They feel more concise while highlighting some edge cases, I like them.

PR #2290, opened 2024-06-26
Add button to attach an ephemeral IP

### LINKED ISSUE #2288: Instance networking: attach ephemeral IP button
We're adding ephemeral IP detach in #2285. An attach ephemeral IP button would make sense next to the floating IP button. There is a max of 1 ephemeral IP per instance ([nexus source](https://github.com/oxidecomputer/omicron/blob/895f2800ad96b51c538ae8011df3654d511563c5/nexus/src/app/mod.rs#L104)), so we should disable the button whenever there is already an ephemeral IP.

image

charliepark:
Fixes #2288 — enables a means of attaching an ephemeral IP to the instance when the instance …
  A) does not have an ephemeral IP,
and
  B) has a network interface
Screenshot 2024-06-27 at 12 21 38 PM
Screenshot 2024-06-27 at 12 21 46 PM
Screenshot 2024-06-27 at 12 21 54 PM

Normally, we'd set up a button to be visible-but-disabled when it can't be interacted with. Because the normal state for instances is that they wouldn't need to attach an ephemeral IP (there's a max count of 1 for ephemeral IPs), and it would be a big disabled button cluttering up the UI, we decided to hide the button when they can't attach an ephemeral IP.
Screenshot 2024-06-27 at 12 27 48 PM


app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -337,9 +339,16 @@ export function NetworkingTab() {
     getCoreRowModel: getCoreRowModel(),
   })
 
-  const disabledReason =
-    eips.items.length >= 32
-      ? 'IP address limit of 32 reached for this instance'
+  const ephemeralDisabledReason =
+    eips.items.filter((ip) => ip.kind === 'ephemeral').length >= 1
+      ? 'Ephemeral IP already attached'
+      : rows.length === 0
```
david-crespo, line 345: rename `rows` to `nics`. I'm guessing the name is a holdover from when this page only had nics

app/pages/project/floating-ips/AttachEphemeralIpModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <Modal.Footer
+        actionText="Attach"
+        disabled={!pool}
+        onAction={() =>
+          instanceEphemeralIpAttach.mutate({
+            path: { instance: instance.name },
+            query: { project },
+            body: { pool },
+          })
+        }
```
david-crespo, line 87: Just a note, no changes needed: I was wondering if we could get this submit working through the normal `onSubmit` mechanism on the `form`. But because of how `Modal.Footer` works:

https://github.com/oxidecomputer/console/blob/34c648b525da61f7c7a6ecf1c036d3c3947cccd6/app/ui/lib/Modal.tsx#L142-L150

we don't have the direct control over the button, which would let us set `form={formRef} type="submit"` on it, which would let us the put this functionality in the `onSubmit` on the form.

david-crespo (approved):
let's party

PR #2296, opened 2024-06-28
Expand combobox flexibility

### LINKED ISSUE #2065: Firewall Rules: Text -> Combobox for VPC, VPC Subnet, and Instance fields
When creating a firewall rule, user has the options of specifying the source (aka "Host filters") and target using VPC, VPC subnet, or a specific instance as the scope. The name of these objects is currently captured in a free-form textbox which may be prone to typos and is just harder to use.

We already have the target/host name field label changed dynamically based on the type field. We probably want to go one step further to make the name field dynamically display the candidates of that target/host type, similar to what we do for Create Disk:
image

Note: "Picklist" is mentioned in a general sense here. It's likely that a picklist of instances may not be practical since there could be many of them. Perhaps we need a different mechanism for listing/specifying an instance name. For IP and IP subnet, textbox may still be the most appropriate input format.

charliepark:
Fixes #2065 

This is an evolution of the combobox.

**Currently,** a combobox provides a dropdown list of items with a text input that allows the user to filter the items in the list.
Screenshot 2024-08-26 at 5 02 54 PM
If the user enters in a non-matching string, they just get the message that there isn't a match.
Screenshot 2024-08-26 at 5 03 09 PM

There are times, though, where the user needs to be able to enter in arbitrary text into the combobox, with the intent being "if this matches item(s) in the list, show me the matching item(s), but if not, give me the option to create a new item. We see this, for example, in the VPC Firewall Rules form, where the user _might_ be selecting from a set of existing VPCs (or subnets, or instances), or _might_ need to create a reference to a not-yet-existing VPC (etc.).

Here, for example, we see that `mock-` matches an option …
Screenshot 2024-08-26 at 5 08 42 PM
… but when we extend the search string a bit and it doesn't match any options, we see that it doesn't show the "no items match" text, and instead acts like a normal text input box …
Screenshot 2024-08-26 at 5 08 59 PM
… and the user can then add that new VPC to the form:
Screenshot 2024-08-26 at 5 09 10 PM

This behavioral change (enable addition of new option) is handled by the `allowArbitraryValues` prop.

This change in Combobox behavior is then applied in this PR to the firewall rules form. In a previous PR we pulled the common code from the firewall-rules-create and firewall-rules-edit forms into a new file, firewall-rules-common. This PR updates that code, moving from the existing text fields (where the user would have to know the instance name, or VPC name, or VPC subnet name), so they can select from a dropdown. This usecase is also where we see the need for the user to add a new item that isn't currently in the list of dropdown items (via that prop mentioned above, `allowArbitraryValues`).


app/ui/lib/Combobox.tsx
```diff
@@ -31,11 +31,17 @@ export type ComboboxBaseProps = {
   description?: React.ReactNode
   isDisabled?: boolean
   isLoading?: boolean
-  items: ComboboxItem[]
+  // without items, the combobox acts as a regular text input
+  items?: ComboboxItem[]
```
david-crespo, line 35: Interesting, this hadn't occurred to me. I think you could still require it and just treat empty array the same as undefined. Because I don't see a place where we'd use this without there being a relevant list (which might happen to be empty).

charliepark:
This is coming along, but man it feels like there's a lot of code duplication in this main file.

Also, we'll eventually need to do something similar re: comboboxes in the router routes forms, and possibly other places. Would love to come up with an extracted component that we could just drop in place.

app/forms/firewall-rules-common.tsx
```diff
@@ -6,115 +6,328 @@
...
+  project: string
+  instances: Array<Instance>
+  vpcs: Array<Vpc>
+  nameTaken: (name: string) => boolean
+  error: ApiError | null
+}
+
+export const CommonFields = ({
+  control,
+  project,
```
david-crespo, line 293: I'd just `const { project } = useProjectSelector()` in here and avoid this prop.

david-crespo (commented):
This is awesome. The interaction works well and feels good. I'm thinking we should change the placeholder or add help text to make clear arbitrary values are allowed. Not sure what the wording is but I'm sure there's plenty of prior art out there in the world.

image

I noticed what may be a bug, it's kind of interesting. Maybe these fields shouldn't be marked `required` at all?

https://github.com/user-attachments/assets/5402527a-b423-4777-bfa4-d25e899c288b

Still going through the harder parts of the code, but here are a few very straightforward suggestions.

app/forms/firewall-rules-edit.tsx
```diff
@@ -32,26 +33,38 @@ import { valuesToRuleUpdate, type FirewallRuleValues } from './firewall-rules-ut
...
+  const originalRule = firewallRules.rules.find((r) => r.name === rule)
   if (!originalRule) throw trigger404
 
+  await Promise.all([
+    apiQueryClient.prefetchQuery('instanceList', {
+      query: { project, limit: PAGE_SIZE },
+    }),
+    apiQueryClient.prefetchQuery('vpcList', { query: { project, limit: PAGE_SIZE } }),
+  ])
+
```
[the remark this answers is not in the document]
charliepark: Resolving, as this was simplified by calling these in the firewall-rules-common file

charliepark:
> Maybe these fields shouldn't be marked `required` at all? They're required for the subform-that-isn't-really-a-form, but the `required` prop applies to the overall form.

Good point. This is a little squirelly, but we could add, to `FieldLabel`, a prop for `showOptionalTag`, defaulting to `true`. Then, in cases where we don't want it to slap the `(optional)` text on the form field label, we pass in `false` for that value.
Screenshot 2024-09-03 at 6 04 17 PM

app/forms/firewall-rules-common.tsx
```diff
@@ -281,22 +284,22 @@ const ProtocolField = ({ control, protocol }: ProtocolFieldProps) => (
...
+  const { project } = useProjectSelector()
+  // prefetchedApiQueries below are prefetched in firewall-rules-create and -edit
+  const { data: instancesData } = usePrefetchedApiQuery('instanceList', {
+    query: { project, limit: PAGE_SIZE },
+  })
+  const instances = instancesData?.items ?? []
+  const { data: vpcData } = usePrefetchedApiQuery('vpcList', {
+    query: { project, limit: PAGE_SIZE },
+  })
+  const vpcs = vpcData?.items ?? []
```
david-crespo, line 301: neither of these can be undefined because of the prefetch. you should be able to do

```ts
const { data: { items: vpcs } } = ...
```

app/forms/disk-attach.tsx
```diff
@@ -57,6 +57,7 @@ export function AttachDiskSideModalForm({
       <ComboboxField
         label="Disk name"
         placeholder="Select a disk"
+        description=""
```
david-crespo, line 60: ?

charliepark: Ah, sorry; was from an earlier experiment. Not necessary because of better defaults in ComboboxField. Will get this and 2 other spots.

app/components/form/fields/useItemsList.ts
```diff
@@ -42,3 +42,14 @@ export const useCustomRouterItems = () => {
 
   return { isLoading, items: routerItems }
 }
+
+// Get a list of subnets for a given VPC; If no project/VPC is provided, use the VPC from the VpcSelector
+export const useVpcSubnets = ({ project, vpc }: { project?: string; vpc?: string }) => {
```
david-crespo, line 47: Might as well make these args required since they're passed in at both call sites

charliepark:
Working through a few validations on the form, specifically around the `required` target and filters.

app/forms/firewall-rules-common.tsx
```diff
@@ -6,115 +6,324 @@
...
+// a list of all items, return the items that are available
+const availableItems = (
+  committedItems: Array<VpcFirewallRuleTarget | VpcFirewallRuleHostFilter>,
+  // Items is conditional because VPC Subnet fetching isn't 100% guaranteed
+  items?: Array<Vpc | VpcSubnet | Instance>
+) => {
+  if (!items) return []
+  return items
+    .map((item) => item.name)
+    .filter((name) => !committedItems.map((ci) => ci.value).includes(name))
```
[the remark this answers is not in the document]
charliepark: Well, that is juicy. Good catch.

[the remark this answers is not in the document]
charliepark: (For posterity: we determined that the VPC subnets available to select here should be restricted to the current VPC, rather than allowing the user to pick any VPC and then any subnet. I'm also adjusting the filter to consider both the type and value, to make sure we don't prune the dropdown list too aggressively.)

app/ui/lib/FieldLabel.tsx
```diff
@@ -17,6 +17,7 @@ interface FieldLabelProps {
   tip?: string
   optional?: boolean
   className?: string
+  hideOptionalTag?: boolean
```
david-crespo, line 20: The prop can be removed from this component thanks to 5e515cbf4e28a33386f3492485356ce71584aa59. It's unused because `optional` covers it.

app/forms/firewall-rules-common.tsx
```diff
@@ -26,95 +37,277 @@ import * as MiniTable from '~/ui/lib/MiniTable'
...
+    type: 'vpc',
+    value: '',
+    // only becomes relevant when the type is 'VPC subnet'; ignored otherwise
+    subnetVpc: vpc,
+  }
+  // prefetchedApiQueries below are prefetched in firewall-rules-create and -edit
+  const {
+    data: { items: instances },
+  } = usePrefetchedApiQuery('instanceList', {
+    query: { project, limit: PAGE_SIZE },
```
david-crespo, line 273: One last thing I thought of: I think instance and VPC list calls should either have no `limit`, which uses a [default value of 100](https://github.com/oxidecomputer/dropshot/blob/d3218d30634dc6419a00a75d56ed98891d0f3c03/dropshot/src/server.rs#L138), or maybe they should have an explicit page size of like 1000, which we use in a few spots when we mean "approximately everything": https://github.com/search?q=repo%3Aoxidecomputer%2Fconsole%20%22limit%3A%201000%22&type=code

Make sure to change both the prefetches and these hook calls if you change it.

david-crespo (approved):
LET'S JAM

app/components/form/fields/ComboboxField.tsx
```diff
@@ -38,22 +37,37 @@ export function ComboboxField<
...
+    : allowArbitraryValues
+      ? 'Select an option or enter a custom value'
+      : 'Select an option',
   ...props
 }: ComboboxFieldProps<TFieldValues, TName>) {
   const { field, fieldState } = useController({ name, control, rules: { required } })
   return (
     <div className="max-w-lg">
       <Combobox
-        isDisabled={disabled}
```
david-crespo, line 48: Is this right? Seems like the mismatch between `disabled` and `isDisabled` may require this to be passed.

david-crespo: Ah, I think I see what you did: you changed the prop on `ComboboxField` to `isDisabled` and let it get passed through as part of the props spread.

https://github.com/oxidecomputer/console/commit/045c58a90f66c3b5eab8d2d504272536b307a7ef

charliepark: Yeah, I realized that there was an existing `isDisabled` prop and it made sense to just use that.

PR #2302, opened 2024-07-01
Fix `disabled:` Tailwind variant

### LINKED ISSUE #2268: Fix `disabled` `addVariant` call in Tailwind config
@charliepark and I just noticed that this line is not behaving as designed. The comma-separated string thing doesn't seem to be valid for Tailwind, so it is just taking the first one, failing to style disabled inputs. We haven't looked into the scope of this, but I expect fixing it will require at least one or two fixes where we were relying on broken behavior, so we want to do this in a separate PR from #2267 where we noticed this.

https://github.com/oxidecomputer/console/blob/ab2939bfa5f38a7d10d3d074e3ebf98ec41842c7/tailwind.config.js#L74

To get the logical OR behavior we want, we need to do this:

```ts
addVariant('disabled', ['&.visually-disabled', '&:disabled'])
```

(To get logical AND behavior you would put them in the same selector, like `&.visually-disabled:disabled`.)

I confirmed what the behavior is in the Tailwind playground:

https://play.tailwindcss.com/BZ5FK84yVE?file=config

image

david-crespo:
Closes #2268. I also cleaned up a couple of other variants.

I wasn't able to find any spots where we were relying on the bad setup, but we should give it a bit more testing.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -71,7 +71,7 @@ module.exports = {
       addVariant('children', '& > *')
       addVariant('between', '& > * + *')
       addVariant('selected', '.is-selected &')
-      addVariant('disabled', '&.visually-disabled, &:disabled')
+      addVariant('disabled', ['&.visually-disabled', '&:disabled'])
       addUtilities(
         Array.from({ length: 12 }, (_, i) => i)
           .map((i) => ({
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -66,7 +66,6 @@ module.exports = {
   plugins: [
     plugin(({ addVariant, addUtilities, variants }) => {
       addVariant('children', '& > *')
-      addVariant('between', '& > * + *')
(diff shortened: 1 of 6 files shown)


tailwind.config.js
```diff
@@ -65,13 +65,9 @@ module.exports = {
   },
   plugins: [
     plugin(({ addVariant, addUtilities, variants }) => {
-      // imitation of the twin.macro svg: variant. svg:text-green-500 puts green
-      // on an SVG that's an immediate child of the element
-      addVariant('svg', '& > svg')
```
david-crespo, line 70: I think when we added this, you couldn't do it inline as an arbitrary value. Now you can.

charliepark (approved):
Flipped through this locally, and didn't come across anything looking broken. Looks good.

david-crespo:
Thanks, will merge after we cut v9.

PR #2303, opened 2024-07-01
Improve combobox implementations in preparation for v9

charliepark:
This PR makes a few general improvements to the combobox component and a few forms that use either comboboxes or listboxes.

- All placeholders should now have a consistent phrasing — `Select a disk`, `Select a pool`, etc., rather than `Select pool`, etc.
- On the instance create form, we had been pre-filling the dropdown on the Boot Disk section with the first item in each dropdown. This made things easier for users, but removed some intentionality. We have removed the default filling of that field, meaning the user will now need to select the appropriate disk for their new instance.
- Updated tests.

### DIFF
--- a/app/components/form/fields/SubnetListbox.tsx
+++ b/app/components/form/fields/SubnetListbox.tsx
@@ -18,7 +18,6 @@ type SubnetListboxProps<
   TName extends FieldPath<TFieldValues>,
 > = Omit<ListboxFieldProps<TFieldValues, TName>, 'items'> & {
   vpcNameField: FieldPath<TFieldValues>
-  noItemsPlaceholder?: string
 }
 
 /**
@@ -31,12 +30,7 @@ type SubnetListboxProps<
 export function SubnetListbox<
   TFieldValues extends FieldValues,
   TName extends FieldPath<TFieldValues>,
->({
-  vpcNameField,
-  control,
-  noItemsPlaceholder,
-  ...fieldProps
-}: SubnetListboxProps<TFieldValues, TName>) {
+>({ vpcNameField, control, ...fieldProps }: SubnetListboxProps<TFieldValues, TName>) {
   const projectSelector = useProjectSelector()
 
   const [vpcName] = useWatch({ control, name: [vpcNameField] })
@@ -59,7 +53,7 @@ export function SubnetListbox<
       disabled={!vpcExists}
       control={control}
       placeholder="Select a subnet"
-      noItemsPlaceholder={noItemsPlaceholder}
+      noItemsPlaceholder={vpcName ? 'No subnets found' : 'Select a VPC to see subnets'}
     />
   )
 }
--- a/app/components/form/fields/SubnetListbox.tsx
+++ b/app/components/form/fields/SubnetListbox.tsx
@@ -18,6 +18,7 @@ type SubnetListboxProps<
   TName extends FieldPath<TFieldValues>,
 > = Omit<ListboxFieldProps<TFieldValues, TName>, 'items'> & {
   vpcNameField: FieldPath<TFieldValues>
+  noItemsPlaceholder?: string
 }
 
 /**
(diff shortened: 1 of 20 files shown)


app/forms/instance-create.tsx
```diff
@@ -774,21 +771,25 @@ const AdvancedAccordion = ({
...
+                      <MiniTable.Cell>{item.ip}</MiniTable.Cell>
+                      <MiniTable.RemoveCell
+                        onClick={() => detachFloatingIp(item.name)}
+                        label={`remove floating IP ${item.name}`}
+                      />
+                    </MiniTable.Row>
+                  ) : (
+                    <></>
+                  )
+                )}
```
david-crespo, line 792: This change is unnecessary because the type of `attachedFloatingIpsData` is `FloatingIp[]` — they can't be falsy. I suspect your VS Code had a blip where it was using TS 5.4, for which this filter fails to narrow the type properly.

```ts
const attachedFloatingIpsData = attachedFloatingIps
  .map((ip) => attachableFloatingIps.find((fip) => fip.name === ip.floatingIp))
  .filter((ip) => !!ip)
```

test/e2e/instance-create.e2e.ts
```diff
@@ -7,7 +7,32 @@
...
+  await page.getByRole('tab', { name: 'Silo images' }).click()
+  await page.getByRole('button', { name: 'Select an image' }).click()
+  await page.getByRole('option', { name }).click()
+}
+
+const selectAProjectImage = async (page: Page, index: number) => {
+  await page.getByRole('tab', { name: 'Project images' }).click()
+  await page.getByRole('button', { name: 'Image' }).click()
+  await page.getByRole('option', { name: images[index].name }).click()
+}
```
david-crespo, line 29: How about changing this one to take a name like the others? That would make the call sites for this function more legible, less abstract. I think the fact that we were using indexes before was kinda wrong.

app/components/form/fields/SubnetListbox.tsx
```diff
@@ -47,11 +47,13 @@ export function SubnetListbox<
...
 
   return (
-    <ComboboxField
+    <ListboxField
       {...fieldProps}
       items={subnets.map(({ name }) => ({ value: name, label: name }))}
       disabled={!vpcExists}
       control={control}
+      placeholder="Select a subnet"
+      noItemsPlaceholder="Select a VPC to see subnets"
```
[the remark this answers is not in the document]
charliepark: Solid. Have updated it.
Screenshot 2024-07-02 at 12 09 23 PM
Screenshot 2024-07-02 at 12 09 31 PM
Screenshot 2024-07-02 at 12 09 43 PM

test/e2e/network-interface-create.e2e.ts
```diff
@@ -20,7 +20,7 @@ test('can create a NIC with a specified IP address', async ({ page }) => {
 
   // fill out the form
   await page.getByLabel('Name').fill('nic-1')
-  await page.getByRole('button', { name: 'VPC' }).click()
+  await page.getByRole('button', { name: 'VPC' }).first().click()
```
david-crespo, line 23: Nitpick, but it's better to handle these cases by being more specific in the selector. 

```suggestion
  await page.getByLabel('VPC', { exact: true }).click()
```
Unfortunately `getByRole('button', { name: 'VPC', exact: true })` didn't work because that's not the exact name, and neither did `getByPlaceholder` (presumably because it's a button, not an input, so it's not really a placeholder).

image

david-crespo (approved):
Looks good. The comments are just suggestions.

app/forms/network-interface-create.tsx
```diff
@@ -65,19 +65,23 @@ export function CreateNetworkInterfaceForm({
...
       />
       <SubnetListbox
         name="subnetName"
         label="Subnet"
         vpcNameField="vpcName"
         required
         control={form.control}
+        noItemsPlaceholder={
+          form.watch('vpcName') ? 'No subnets found' : 'Select a VPC to see subnets'
+        }
```
david-crespo, line 84: I'd rather do this inside `SubnetListbox`. You can look at `vpcName` in there.

charliepark:
… or … rather, missed a few last tests; updating

PR #2304, opened 2024-07-02
Update empty state text

### LINKED ISSUE #2289: Reword "you need to ... to be able to ..." empty table message
We use this message all over ([code search](https://github.com/search?q=repo%3Aoxidecomputer%2Fconsole%20%22you%20need%20to%22&type=code)), but the tone is a bit off. "Create an instance to see it here" conveys the same thing with a lot less weirdness, so that's my default choice, but we can probably do better. Maybe we don't need this bit of text at all.

image

charliepark:
Fixes #2289 

It was quick to update the copy here, though if we decide we want to remove the text completely, that's fine, too. I think for now it makes sense to keep the text in place.

I added a `text-balance` CSS class to balance multi-line text, though I'm holding lightly to that if we aren't into it:

**Without `text-balance`:**
Screenshot 2024-07-02 at 2 24 25 PM
**With `text-balance`:**
Screenshot 2024-07-02 at 2 25 29 PM

### DIFF
--- a/app/components/form/fields/SshKeysField.tsx
+++ b/app/components/form/fields/SshKeysField.tsx
@@ -131,7 +131,7 @@ export function SshKeysField({
           <EmptyMessage
             icon={<Key16Icon />}
             title="No SSH keys"
-            body="You need to add a key to be able to see it here"
+            body="Add a key to see it here"
             buttonText="Add SSH Key"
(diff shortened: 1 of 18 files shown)


david-crespo:
Oh yeah, love the `text-balance`! I don't think I knew about that.

david-crespo (approved):
Love it

PR #2313, opened 2024-07-03
Add copy-to-clipboard to External IPs table on networking tab

### LINKED ISSUE #2307: Add quick copy button to External IPs table
In a runthrough on console with Will, he noted the inconsistency on the instance networking tab, where the external IPs don't have the click-to-copy button (though the copy button is there in the IP list in the header).

We need the quick actions button in that row because of the detach action, so we can now add that copy button directly to the cell.

charliepark:
This adds a "copy to clipboard" component to the IP cell on the external IPs and network interfaces tables on the instance's networking tab.

Screenshot 2024-07-03 at 2 36 09 PM

As this ends up being practically identical to the IP rendered in the header, I extracted the copy-able IP address into a new `CopyableIp` component. (Open to alternate names!) Because we might have occasions where the IP does not need to be linked (like on the Network Interfaces row), the linking is optional.

Fixes #2307


david-crespo (approved):
I like it. I thought it might feel cluttered, but the buttons totally blend in.

PR #2315, opened 2024-07-03
DRAFT: Add form navigation guard

### LINKED ISSUE #2186: Use `useBlocker` to confirm closing side modal form
Similar to #996 — it's kind of easy to accidentally click out of a side modal form and lose everything you've put in.

charliepark:
This is a WIP that I'll come back to next week.

The idea is that we take the `useBlocker` hook that we're currently using in FullPageForm and we put it in its own component that we can then use on both the FullPageForm and the SideModalForm. It seems to be working right now, but I'll want to try more edge cases, see what obvious things I'm missing, and otherwise kick the tires a bit more.

Fixes #2186


app/components/form/FormNavGuard.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useEffect } from 'react'
+import { type FieldValues, type UseFormReturn } from 'react-hook-form'
+import { useBlocker } from 'react-router-dom'
+
+import { Modal } from '~/ui/lib/Modal'
+
+export function FormNavGuard<TFieldValues extends FieldValues>({
+  form,
+}: {
+  form: UseFormReturn<TFieldValues>
```
david-crespo, line 17: This doesn't need to be generic since we're not actually using the form values. I think you can do `form: UseFormReturn` with no param and it'll work.

charliepark:
One or two test issues I'll have to work out as well

app/components/form/FormNavGuard.tsx
charliepark: Both of the callsites for this have `form` as a `UseFormReturn<TFieldValues>`. I think it'd be easier to just pass it in as-is and set the type within the new component to also have the generic? (TS complains at the callsites if I strip it from the new component's type) Or is there an aspect to generics that I'm missing?

david-crespo: Yep, I was wrong. I thought that leaving it off like this 

```diff
-export function FormNavGuard<TFieldValues extends FieldValues>({
-  form,
-}: {
-  form: UseFormReturn<TFieldValues>
-}) {
+export function FormNavGuard({ form }: { form: UseFormReturn }) {
   const { isDirty, isSubmitting, isSubmitSuccessful } = form.formState
   // Confirms with the user if they want to navigate away if the form is
   // dirty. Does not intercept everything e.g. refreshes or closing the tab
```

would be more like `UseFormReturn<any>` (which does work btw because it doesn't care about types anymore after), but it's actually `UseFormReturn<FieldValues>` (the default value for that generic param) and surprisingly, that is incompatible with our `TFieldValues extends FieldValues`. I think you're right, the best thing is to leave it as-is.

charliepark:
Test seems flaky; running playwright locally in both chrome and firefox seems to be fine; merged in main and trying again.

david-crespo:
That just flaked on main, too: https://github.com/oxidecomputer/console/actions/runs/9876960315/job/27277408168

charliepark:
This is in a pretty good spot, though when there isn't a route change (for example, the Create New Disk form in the Instance Create form … in fact, that might be the only one), we don't get the benefit of useBlocker preventing the dismissal of the side modal. One possible approach would be to use the onDismiss function for the SideModalForm, and to check if the form is dirty before firing the onDismiss. When doing that, if it *is* dirty, open the confirmation modal. I'll need to revisit this, though, as I'm a bit foggy.

charliepark:
If we do decide to add a guard for the side modal form, #2328 is a much simpler implementation, so I don't think this particular branch / implementation makes sense to consider further.

PR #2328, opened 2024-07-17
Simpler nav guard for SideModalForm

### LINKED ISSUE #2186: Use `useBlocker` to confirm closing side modal form

charliepark:
This adds a modal for the SideModalForm that protects it from getting accidentally closed when the form is dirty.

I had experimented with [a more involved approach, using useBlocker](https://github.com/oxidecomputer/console/pull/2315), but as Crespo and I were talking through it, he noted that side modal forms don't always require a route change, so this is possibly both more reliable and simpler.

The deeper question: is the benefit of adding this bit of friction worth it? I'm leaning towards "yes", as the friction is pretty minimal and the potential for data loss is … low-to-moderate? … but am open to alternate takes.

Fixes #2186


david-crespo:
This is great. Modal on top of modal is a little funny, but I can live with it, the QoL improvement is huge.

Text is a little tight when there's only one line. Like a McDonald's hamburger, super thin patty. @benjaminleonard ?

image

david-crespo:
Before/after on overlay fix. Note how in the before, the background gets double-dark when the confirm modal opens.

https://github.com/user-attachments/assets/22572e5b-baa7-4d60-be99-5cb82866610e

PR #2329, opened 2024-07-17
Remove truncation on topbar dropdown

### LINKED ISSUE #2327: Truncation tooltip on long project name in top bar picker is obscured
image

charliepark:
This removes the truncation in the topbar dropdown, just in case that's the direction we want to go with it. If not, no worries.

Fixes #2327 if we go with it.

Screenshot 2024-07-17 at 11 45 03 AM


david-crespo:
Sad on mobile, but already sad without this change, so I don't think it matters. A problem for #2087 to deal with. 

image

PR #2337, opened 2024-07-23
Add columns on SledsTab for sled Policy and State

### LINKED ISSUE #2333: Add sled state and policy to sleds table
Just like we did with physical disks.

charliepark:
This PR adds columns for the Policy and State on the Sleds tab, as we have them on the Disks tab.

The columns on the right side are new:
Screenshot 2024-07-23 at 3 47 44 PM

Same as before:
Screenshot 2024-07-23 at 3 47 55 PM

Fixes #2333

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -45,6 +45,7 @@ module.exports = {
     '@typescript-eslint/no-empty-interface': 'off',
     '@typescript-eslint/ban-ts-comment': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
+    '@typescript-eslint/no-duplicate-type-constituents': 'off',
     '@typescript-eslint/no-unused-vars': [
       'error',
       { argsIgnorePattern: '^_', varsIgnorePattern: '^_' },
--- a/app/components/StatusBadge.tsx
+++ b/app/components/StatusBadge.tsx
@@ -5,7 +5,13 @@
  *
  * Copyright Oxide Computer Company
  */
-import type { DiskState, InstanceState, SnapshotState } from '@oxide/api'
+import type {
+  DiskState,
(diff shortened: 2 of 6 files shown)


app/components/StatusBadge.tsx
```diff
@@ -69,3 +75,17 @@ export const SnapshotStatusBadge = (props: {
...
+}
+
+const STATE_BADGE_COLORS: Record<PhysicalDiskState | SledState, BadgeColor> = {
+  active: 'default',
+  decommissioned: 'neutral',
+}
+
+export const StateBadge = ({ state }: { state: PhysicalDiskState | SledState }) => (
+  <Badge color={STATE_BADGE_COLORS[state]}>{state}</Badge>
+)
```
charliepark, line 91: This one could be a little simpler if we skipped the STATE_BADGE_COLOR piece of it and just did what we do with PolicyKindBadge where we accept explicit strings.

david-crespo:
It would be good to add a couple more sleds to the mock data so that they show the possible combinations of states and policies, like disks do.

.eslintrc.cjs
```diff
@@ -45,6 +45,7 @@ module.exports = {
     '@typescript-eslint/no-empty-interface': 'off',
     '@typescript-eslint/ban-ts-comment': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
+    '@typescript-eslint/no-duplicate-type-constituents': 'off',
```
david-crespo, line 48: For future us, we did this because eslint's fix on save was eliminating a redundant arm from this type union (which we're no longer using) which we wanted to keep.

```
type T = SledPolicy['kind'] | PhysicalDiskPolicy['kind']
```

PR #2339, opened 2024-07-25
Add SSH info to Instance Connect tab

### LINKED ISSUE #2271: Instance connect tab should talk about SSH
Extremely obvious in hindsight. I suspect most people clicking the connect tab are not looking for the serial console, and are instead looking for SSH or VNC. We should say something about SSH, like

> If you specified SSH keys when you created the instance, you can connect to it through an external IP [copy dynamic based on whether it has external IPs]:
>
> ```
> ssh username@[actual IP]
> ```
> 
> Blah blah blah, username depends on the OS, etc.

and ideally link to a doc. The [quick start appendix](https://docs.oxide.computer/guides/quickstart#_appendix) has a section about creating SSH keys — maybe that should be its own guide.

Re: VNC, we don't do a lot of "coming soon" in the console, so we should probably just link to a doc. (We don't have a doc yet.)

image

charliepark:
Fixes #2271

This adds an informational box to the Connect tab, talking about SSH and how the user can connect with their instance.

If they've set up any external IPs, it gives them most of what they need, using one of their floating IPs (and falling back to an ephemeral IP if no floating IPs).

image

If they haven't set up any external IPs, it directs them to the networking tab to set one up.

image

Learn more links go to the new Remote Access guide: https://docs.oxide.computer/guides/remote-access


charliepark:
We could maybe add one of the docs popup buttons to the top right of the SSH box to house the docs link, but I don't like how that'd add a whole new piece to that component.
Screenshot 2024-07-25 at 12 08 21 PM

david-crespo:
I'm fine with the docs link right in the text.

> Read our [SSH guide](#) to learn more.

or whatever

app/pages/project/instances/instance/tabs/ConnectTab.tsx
```diff
@@ -6,33 +6,102 @@
...
+  const sshCopy = externalIp ? (
+    <>
+      <p>
+        If you specified SSH keys when you created this instance, you can connect to it
+        through an external IP: <InlineCode>ssh [username]@{externalIp}</InlineCode>
+      </p>
+      <p>
+        The <InlineCode>[username]</InlineCode> in the SSH command will depend on your
+        instance’s boot disk’s OS, but might be <InlineCode>debian</InlineCode>,{' '}
+        <InlineCode>ubuntu</InlineCode>, <InlineCode>arch</InlineCode>, etc.
```
charliepark, line 54: I haven't actually confirmed that `arch` is correct, or what we want to use here; open to alternates.

paryhin:
This styling is slightly confusing, as the SSH information section looks like a modal with no buttons. For the information banners, we should stick to consistently using the blue style like this: 

CleanShot 2024-08-05 at 13 25 22@2x

david-crespo:
What about the serial console one? It also is not a modal. I was thinking these two should be styled the same because they are two ways of connecting to the instances. We might have a third one for VNC soon.

paryhin:
These are different components: the first one is a card (e.g., [Card Component](https://ui.shadcn.com/docs/components/card)), and the second one is an alert / banner /  (e.g., [Alert Component](https://ui.shadcn.com/docs/components/alert)). They serve different purposes and have different properties. Cards are used for grouping related content and actions, while alerts are used to call out user attention and provide contextual information.  For example, we may actually put the banner / alert in the card itself if we need a better visual grouping.

david-crespo:
Current state:

image

david-crespo:
This is close, but needs:

- [ ] Copy once-over
- [ ] Style check on inline code bit, it feels not quite right
- [ ] A decent docs link for serial console

charliepark:
Current: Screenshot 2024-08-21 at 4 56 32 PM

Issue #1714, opened 2023-08-07
Focus a default button instead of a close button in dialogs


iliana:
![image](https://github.com/oxidecomputer/console/assets/52814/72db67de-e716-4e19-9343-be4b39c58776)

This is after hitting the instance menu -> Delete; the focus is around the close button but it would make more sense to focus a default button.

This is a particularly useful example where the default focus should absolutely not be the dangerous red "Confirm" button.

PR #2340, opened 2024-07-25
Focus confirm button instead of cancel in modals

david-crespo:
We fixed this once in #1714, though maybe it was only for side modals. In any case, here I swap the order in the HTML so the confirm button autofocuses and use `flex-row-reverse` to get them to display visually in the right order. This has the downside that tabbing through goes in a funny order, but that seems less bad on net than the status quo.

### Before

image

### After

image


charliepark (approved):
nice

david-crespo:
Reverted in 6615cb6b5b2470cdc0ae6d66392115c413e0adc8

I missed this _very_ clear line in #1714:

> This is a particularly useful example where the default focus should absolutely not be the dangerous red "Confirm" button.

PR #2343, opened 2024-07-26
Use real links in dropdown menus

### LINKED ISSUE #2342: User settings link is not a link
In the user dropdown (upper right corner), the "Settings" option is not a real link. A user is unable to middle-mouse click or right-click > open in new tab to open the settings page in a new tab without losing their place on the current page.

david-crespo:
Closes #2342 

Looks like this works, but I need to make sure there's nothing weird about it. I had noticed this problem before, but for some reason when I saw #2342 I immediately had a vision of how to do it.

```[tasklist]
- [x] Add `DropdownMenu.LinkItem`
- [x] Use it for the settings link per #2342
- [x] Make sure it's not messed up in any way
- [ ] ~~Use it for other menu items that work as links!~~
```

### DIFF
--- a/app/ui/styles/components/menu-button.css
+++ b/app/ui/styles/components/menu-button.css
@@ -10,7 +10,7 @@
   @apply z-30 min-w-36 rounded border p-0 bg-raise border-secondary;
 
   & .DropdownMenuItem {
-    @apply block cursor-pointer select-none border-b py-2 pl-3 pr-6 text-left text-sans-md text-secondary border-secondary last-of-type:border-b-0;
+    @apply block cursor-pointer select-none border-b py-2 pl-3 pr-6 text-left text-sans-md text-secondary border-secondary last:border-b-0;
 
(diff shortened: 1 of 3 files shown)


app/ui/styles/components/menu-button.css
```diff
@@ -10,7 +10,7 @@
   @apply z-30 min-w-36 rounded border p-0 bg-raise border-secondary;
 
   & .DropdownMenuItem {
-    @apply block cursor-pointer select-none border-b py-2 pl-3 pr-6 text-left text-sans-md text-secondary border-secondary last-of-type:border-b-0;
+    @apply block cursor-pointer select-none border-b py-2 pl-3 pr-6 text-left text-sans-md text-secondary border-secondary last:border-b-0;
```
david-crespo, line 13: `last-of-type` was getting thrown off by the mix of `Item` and `LinkItem`.

PR #2349, opened 2024-07-30
tooling: make mock API default firewall rules more accurate

david-crespo:
* Remove `allow-rdp` rule, which was removed in the real API a long time ago
* Set up the default rules in new VPCs that get created
* Update VPC create form test to actually submit the form


charliepark (approved):
👍

PR #2351, opened 2024-07-30
Other dropdown links

david-crespo:
I went and looked at other spots where we could convert `navigate` calls in dropdown menu item callbacks to real `Link`s. I don't really like it, though, so I am probably posting this to throw away.


david-crespo:
Nothing has happened since I posted this to make me like it more, so bye

PR #2355, opened 2024-08-05
Upgrade to RR 6.26 and use new `replace` helper instead of `<Navigate>`

### LINKED ISSUE #2352: Use new `replace` instead of `<Navigate>` for route replace navs
We have a few of these and they've never felt quite right. Now we can do them slightly earlier in the render cycle with `loader={() => replace('/path')}` thanks to https://github.com/remix-run/react-router/pull/11811.

https://github.com/oxidecomputer/console/blob/07b6c151ccba7fb31957e247faa499d1a26b266a/app/routes.tsx#L215-L218

david-crespo:
Closes #2352 

Seems to work! I like it! The nice thing about this is we can stop doing that thing where we run the loader for the target page on the starting page. That was to prevent a situation where you're sitting on a blank page while the loader for the target page runs. At least, I couldn't trigger the blank screen.

### DIFF
--- a/app/routes.tsx
+++ b/app/routes.tsx
@@ -5,7 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
-import { createRoutesFromElements, Navigate, Route } from 'react-router-dom'
+import { createRoutesFromElements, replace, Route } from 'react-router-dom'
 
 import { RouterDataErrorBoundary } from './components/ErrorBoundary'
 import { NotFound } from './components/ErrorPage'
@@ -160,7 +160,7 @@ export const routes = createRoutesFromElements(
           loader={InventoryPage.loader}
           handle={{ crumb: 'Inventory' }}
         >
-          <Route index element={<Navigate to="sleds" replace />} loader={SledsTab.loader} />
+          <Route index element={null} loader={() => replace('sleds')} />
           <Route path="sleds" element={<SledsTab />} loader={SledsTab.loader} />
           <Route path="disks" element={<DisksTab />} loader={DisksTab.loader} />
         </Route>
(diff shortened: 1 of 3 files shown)


app/routes.tsx
```diff
@@ -160,7 +160,7 @@ export const routes = createRoutesFromElements(
           loader={InventoryPage.loader}
           handle={{ crumb: 'Inventory' }}
         >
-          <Route index element={<Navigate to="sleds" replace />} loader={SledsTab.loader} />
+          <Route index element={null} loader={() => replace('sleds')} />
```
david-crespo, line 163: The nulls are because you get warnings if you don't have an element to render, even if the loader returns a redirect.

david-crespo:
Well, I was able to find one weird behavior. I changed this line so the path is `/projects/<project name>` so the `replace` thing happens when you click link to that page.

https://github.com/oxidecomputer/console/blob/07b6c151ccba7fb31957e247faa499d1a26b266a/app/util/path-builder.ts#L38

When I click the link, the redirect works as expected and there is no blank screen. However, when I click the back button, I get _nothing_, which means that instead of leaving `/projects` (A) there and replacing `/projects/mock-project` (B1) with `/projects/mock-project/instances` (B2), it is replacing A with B1 and then B1 with B2. That seems like an RR bug to me, so I'll look into it on the RR side.

https://github.com/user-attachments/assets/f76f998f-1297-4584-976c-4939677ea9ac

david-crespo:
Reported the possible bug in https://github.com/remix-run/react-router/issues/11873 with a stackblitz repro.

david-crespo:
`replace` doesn't work like I expected. https://github.com/remix-run/react-router/issues/11873#issuecomment-2278268934 suggests we could use `replace` on initial pageload and then `redirect` on subsequent navs, but I'm not sure how to do that yet and there's no urgency here. We can figure it out later.

PR #2358, opened 2024-08-07
Error on firewall rule name conflicts

### LINKED ISSUE #2220: Creating a new firewall rule with the same name overwrites original

david-crespo:
Closes #2220 

Still needs a little work, maybe copy polish and more importantly putting an error indicator at form level so there's some feedback if you hit submit while the name field is off screen.

image

### DIFF
--- a/app/forms/firewall-rules-create.tsx
+++ b/app/forms/firewall-rules-create.tsx
@@ -589,15 +589,25 @@ export function CreateFirewallRuleForm() {
       title="Add firewall rule"
       onDismiss={onDismiss}
       onSubmit={(values) => {
-        // TODO: this silently overwrites existing rules with the current name.
-        // we should probably at least warn and confirm, if not reject as invalid
-        const otherRules = existingRules
-          .filter((r) => r.name !== values.name)
-          .map(firewallRuleGetToPut)
+        // error if name is already in use
+        if (data.rules.find((r) => r.name === values.name)) {
+          // TODO: there should also be a form-level error so if the name is off
+          // screen, it doesn't look like the submit button isn't working. Maybe
+          // instead of setting a root error, it would be more robust to show a
+          // top level error if any errors are found in the form. We might want to
+          // expand the submitError prop on SideModalForm for this
+
+          form.setError('name', {
+            // TODO: might be worth mentioning that the names are unique per VPC as opposed to globally
+            message: 'Name taken. To update an existing rule, edit it directly. ',
+          })
+          return
+        }
+
         updateRules.mutate({
           query: vpcSelector,
           body: {
-            rules: [...otherRules, valuesToRuleUpdate(values)],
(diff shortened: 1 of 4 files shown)


app/forms/firewall-rules-create.tsx
```diff
@@ -589,15 +589,25 @@ export function CreateFirewallRuleForm() {
...
+          // top level error if any errors are found in the form. We might want to
+          // expand the submitError prop on SideModalForm for this
+
+          form.setError('name', {
+            // TODO: might be worth mentioning that the names are unique per VPC as opposed to globally
+            message: 'Name taken. To update an existing rule, edit it directly. ',
+          })
+          return
+        }
+
```
david-crespo, line 606: Doing this here works in a basic sense, but it would probably be more strictly correct and more flexible if we could do it as a `validate` prop on the `NameField` in question.

https://github.com/oxidecomputer/console/blob/07b6c151ccba7fb31957e247faa499d1a26b266a/app/forms/firewall-rules-create.tsx#L202

Right now, `NameField` specifically disallows additional validation, but it would be easy to extend this to allow a function to be passed in, which we would call after `validateName`.

https://github.com/oxidecomputer/console/blob/07b6c151ccba7fb31957e247faa499d1a26b266a/app/components/form/fields/NameField.tsx#L14-L25

The nice thing about hooking into RHF's validation situation is that we get certain behaviors for free, for example:

* After an initial error, forms revalidate their inputs on every change event. So if you put in `allow-ssh` and get a name taken error, changing it to `allow-ssh2` would immediately clear the error. In the current implementation, that error is not cleared until submit.
* If we wanted to change initial validation behavior for the form as a whole (for example changing it from the default `mode: 'onSubmit'` to `mode: 'onBlur'` so the user would see right away that they can't use that name) that only affects this bit of validation if it hooked up to the input

david-crespo:
It works, and it automatically scrolls to the error on validation failure, so we don't actually need a form-level error to alert you. That's great.

https://github.com/user-attachments/assets/47ad432e-12e6-44ea-b658-9021f7a984a5

At the end of the video, on successful edit, I did notice you can get a brief `Something went wrong` page flash due to an invariant failure. The fact that it's intermittent suggests a race condition around React state updates. I reproduced it once or twice, but then I had trouble.

image

david-crespo:
I wasn't able to repro that issue anymore today, but in 7044287c62e44923374a8660406770cfe1a50877 I made a change that should give some protection: on successful update, dismiss the modal before refetching the rules.

charliepark (approved):
Nice!

PR #2359, opened 2024-08-08
Add router tab and routes pages

### LINKED ISSUE #2295: VPC routers and routes
We can easily have a top-level routers tab on the VPC detail listing the routers, but routes are children of routers, so I think we'd have to have a click on router take you to a router detail page listing the routes.

Routers don't have much going on, data-wise. Routes have a target and destination, and those look a lot like firewall rule targets, so we know how to display those.

https://github.com/oxidecomputer/console/blob/db99a4db4eb7f17cad953b9f4d7d804f3e1bf650/app/api/__generated__/Api.ts#L3700-L3719

https://github.com/oxidecomputer/console/blob/db99a4db4eb7f17cad953b9f4d7d804f3e1bf650/app/api/__generated__/Api.ts#L2525-L2570

charliepark:
Closes #2295 

This PR adds a new tab to the VPC page, showing that VPC's routers.
Screenshot 2024-08-14 at 1 30 09 PM

Clicking through to the router then shows the routes belonging to that router.
Screenshot 2024-08-14 at 1 30 19 PM
Screenshot 2024-08-14 at 1 30 33 PM

Certain restrictions apply, such as not being able to add routes to system routers …
Screenshot 2024-08-14 at 1 31 57 PM

… or not allowing certain edit/deletes to happen for system routers …
Screenshot 2024-08-14 at 1 32 34 PM

… though for custom routers, things are fine …
Screenshot 2024-08-14 at 1 33 03 PM


app/forms/vpc-router-create.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export function CreateRouterSideModalForm() {
+  const queryClient = useApiQueryClient()
+  const vpcSelector = useVpcSelector()
+  const navigate = useNavigate()
+
+  const onDismiss = (navigate: NavigateFunction) => {
+    navigate(pb.vpcRouters(vpcSelector))
+  }
```
david-crespo, line 31: I think all the `onDismiss` can do without the `navigate` arg and close over the `navigate` that's already in scope in the component.

```suggestion
  const onDismiss = () => navigate(pb.vpcRouters(vpcSelector))
```

app/forms/vpc-router-route-edit.tsx
```diff
@@ -0,0 +1,136 @@
...
+]
+
+const targetTypes: ListboxItem[] = [
+  { value: 'ip', label: 'IP' },
+  { value: 'vpc', label: 'VPC' },
+  { value: 'subnet', label: 'subnet' },
+  { value: 'instance', label: 'instance' },
+  { value: 'internet_gateway', label: 'Internet gateway' },
+  { value: 'drop', label: 'Drop' },
+]
```
david-crespo, line 46: Very slightly safer:

```ts
type DestinationItem = { value: RouteDestination['type']; label: string }

const destinationTypes: DestinationItem[] = [
  // ...
]

type TargetItem = { value: RouteTarget['type']; label: string }

const targetTypes: TargetItem[] = [
  // ...
]
```

app/forms/vpc-router-route-create.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <NameField name="name" control={form.control} />
+      <DescriptionField name="description" control={form.control} />
+      <ListboxField
+        name="destination.type"
+        items={routerRouteDestinationTypes}
+        label="Destination type"
+        control={form.control}
+        placeholder="Select a destination type"
+        required
+      />
```
david-crespo, line 71: Here's a way to share all the common props without having to deal with `control`: https://github.com/oxidecomputer/console/commit/54c40c431921c521dcf1ab55032d7e9742cb9d36. Seems to work.

app/pages/project/vpcs/VpcPage/tabs/VpcRoutersTab.tsx
```diff
@@ -0,0 +1,117 @@
...
+        cell: (info) => (
+          <LinkCell to={pb.vpcRouter({ ...vpcSelector, router: info.getValue() })}>
+            {info.getValue()}
+          </LinkCell>
+        ),
+      }),
+      colHelper.accessor('description', {
+        header: 'Description',
+        cell: (info) => info.getValue() || <EmptyCell />,
+      }),
```
david-crespo, line 64: There's a common description column that does truncation for you.

https://github.com/oxidecomputer/console/blob/07b6c151ccba7fb31957e247faa499d1a26b266a/app/pages/system/silos/SiloIdpsTab.tsx#L46

mock-api/msw/db.ts
```diff
@@ -138,6 +138,39 @@ export const lookup = {
...
+    )
+    if (!route) throw notFoundErr(`route '${id}'`)
+
+    return route
+  },
+  vpcRouterRouteList({ project, router, vpc }: PP.VpcRouter): Json<Api.RouterRoute>[] {
+    if (!project) throw notFoundErr('no project specified')
+    if (!router) throw notFoundErr('no router specified')
+    if (!vpc) throw notFoundErr('no VPC specified')
+    return db.vpcRouterRoutes
```
david-crespo, line 172: This should use the router lookup and then filter the db table with the router ID

mock-api/vpc.ts
```diff
@@ -45,6 +45,98 @@ export const vpc2: Json<Vpc> = {
...
 export const vpcs: Json<Vpc[]> = [vpc, vpc2]
 
+export const routerRoutes: Json<Array<RouterRoute>> = [
+  {
+    id: '51e50342-790f-4efb-8518-10bf01279514',
+    name: 'default1',
+    description: "VPC Subnet route for 'default'",
+    time_created: '2024-07-11T17:46:21.161086Z',
+    time_modified: '2024-07-11T17:46:21.161086Z',
+    vpc_router_id: 'fc59fb4d-baad-44a8-b152-9a3c27ae8aa1',
```
david-crespo, line 55: You could make this and spread it into each of these to take some of the noise out

```ts
const routeBase = {
  time_created: '2024-07-11T17:46:21.161086Z',
  time_modified: '2024-07-11T17:46:21.161086Z',
  vpc_router_id: vpcRouter.id,
  kind: 'default',
}
```

mock-api/msw/handlers.ts
```diff
@@ -1053,6 +1053,84 @@ export const handlers = makeHandlers({
...
+        .filter((r) => r.vpc_router_id === vpcRouter.id),
+    }
+  },
+  vpcRouterRouteCreate({ body, query }) {
+    const vpcRouter = lookup.vpcRouter(query)
+    const newRoute: Json<Api.RouterRoute> = {
+      id: uuid(),
+      vpc_router_id: vpcRouter.id,
+      // TODO: look into proper setting of `kind`
+      kind: 'default',
```
david-crespo, line 1111: All user-created routes are `kind: custom`.

https://github.com/oxidecomputer/omicron/blob/5fb9029a0f03a18969b41c5c7f96a35f35abfc2f/nexus/src/external_api/http_entrypoints.rs#L5737

app/forms/vpc-router-route-create.tsx
charliepark: Oh, awesome. Thank you.

charliepark:
Ah, per https://rfd.shared.oxide.computer/rfd/0021#vpc_route_types, not every route type is modifiable, so I'll need to revisit the logic on the action buttons on those rows.

david-crespo:
That RFD is very old. You should always confirm in the API code. Unfortunately it's kind of a weird one: looks like `default` type routes can also be updated, but only their target. Ew. But I guess it's not that hard to hide the fields in the form?

https://github.com/oxidecomputer/omicron/blob/5fb9029a0f03a18969b41c5c7f96a35f35abfc2f/nexus/src/app/vpc_router.rs#L252-L272

app/pages/project/vpcs/VpcPage/tabs/VpcRoutersTab.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const staticColumns = useMemo(
+    () => [
+      colHelper.accessor('name', {
+        cell: makeLinkCell((router) => pb.vpcRouter({ ...vpcSelector, router })),
+      }),
+      colHelper.accessor('description', Columns.description),
+    ],
+    [vpcSelector]
+  )
```
david-crespo, line 60: Table looks sparse. Might as well add a created time column like in the subnets table. That's our usual table padding.

Screenshot 2024-08-13 at 12 07 26 PM

app/forms/vpc-router-route-edit.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const queryClient = useApiQueryClient()
+  const routeSelector = useVpcRouterRouteSelector()
+  const { project, vpc, router: routerName, route: routeName } = routeSelector
+  const navigate = useNavigate()
+  const { data: route } = usePrefetchedApiQuery('vpcRouterRouteView', {
+    path: { route: routeName },
+    query: { project, vpc, router: routerName },
+  })
+
+  const defaultValues: RouterRouteUpdate = { ...route }
```
david-crespo, line 57: So this is kind of interesting. To begin with, I don't think we need to copy this, so you could just do

```ts
const defaultValues: RouterRouteUpdate = route
```

However, there's something else kind of interesting here. Thanks to TypeScript's most problematic characteristic, the `defaultValues` object has a lot more stuff on it at runtime than the type system is aware of. So when you use `body` directly in `onSubmit` without pulling out specific values, you actually get all of them. Somewhat surprisingly, the API does not actually care that there's extra stuff there. I tried it on colo and got a 200. And on reflection, I think if this was a problem, we would have seen it before.

Screenshot 2024-08-13 at 12 06 53 PM

The best thing to do is probably what subnet edit does:

https://github.com/oxidecomputer/console/blob/3a6f815ad50bf2c7fafb55dcdd31ad0232adbfeb/app/forms/subnet-edit.tsx#L53

Though honestly the use of `satisfies` there is probably suboptimal since it doesn't give the resulting thing a type of `VpcSubnetUpdate`. This is probably better:

```ts
const defaultValues: VpcSubnetUpdate = R.pick(subnet, ['name', 'description'])
```

Also based on a quick search (`rg body **/*edit*`) it looks like we usually use destructuring in `onSubmit` to stick only the fields we want in the body.

app/forms/vpc-router-route/shared.tsx
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+import type { RouteDestination, RouteTarget } from '~/api'
+
+// VPCs can not be specified as a destination in custom routers
+// https://github.com/oxidecomputer/omicron/blob/4f27433d1bca57eb02073a4ea1cd14557f70b8c7/nexus/src/app/vpc_router.rs#L363
+const destTypes: Record<Exclude<RouteDestination['type'], 'vpc'>, string> = {
+  ip: 'IP',
+  ip_net: 'IP net',
+  subnet: 'subnet',
```
david-crespo, line 16: These labels should be capitalized since they appear in the listbox. Also `IP net` should probably be `IP network` like we're doing in the badge in `RouterRoutePage`.

Screenshot 2024-08-13 at 12 05 34 PM

app/pages/project/vpcs/RouterRoutePage.tsx
```diff
@@ -0,0 +1,212 @@
...
+  ])
+  return null
+}
+
+const RouterRouteTypeValueBadge = ({ type, value }: { type: string; value?: string }) => {
+  const typeString = type
+    .replace('_', ' ')
+    .replace('ip net', 'ip network')
+    .replace('internet gateway', 'gateway')
+    .replace('subnet', 'VPC subnet')
```
david-crespo, line 61: This should be done as a record like `destTypes` and `targetTypes` to make sure we're always catching all possible cases. Slightly awkward question of whether we should use exactly the same abbreviations in both spots. I am leaning no, since "Internet gateway" looks ridiculous in the badge but it probably looks fine in the listbox on the form. So I guess I'd say make a separate mapping in this file for this purpose.

david-crespo (commented):
first pile o' comments

test/e2e/vpcs.e2e.ts
```diff
@@ -74,3 +74,99 @@ test('can create and delete subnet', async ({ page }) => {
...
+  await expect(page.getByRole('textbox', { name: 'Target value' })).toHaveValue('outbound')
+  await page.getByRole('button', { name: 'Target type' }).click()
+  await page.getByRole('option', { name: 'IP' }).click()
+  await page.getByRole('textbox', { name: 'Target value' }).fill('1.1.1.1')
+  await page.getByRole('button', { name: 'Create route' }).click()
+  await expect(routeRows).toHaveCount(2)
+  await expectRowVisible(table, { Name: 'new-route' })
+
+  // see the destination value of 0.0.0.0
+  await expectRowVisible(table, { Destination: 'ip0.0.0.0' })
```
david-crespo, line 155: this and the above can be combined into one since it's one row. same with 162-165. and you might as well assert the target value too

app/pages/project/vpcs/RouterRoutePage.tsx
```diff
@@ -0,0 +1,212 @@
...
+      <PageHeader>
+        <PageTitle icon={<Networking24Icon />}>{router}</PageTitle>
+        <div className="inline-flex gap-2">
+          <DocsPopover
+            heading="Routers"
+            icon={<Networking16Icon />}
+            summary="Routers summary copy TK"
+            links={[docLinks.routers]}
+          />
+          <MoreActionsMenu label="Instance actions" actions={actions} />
```
david-crespo, line 180:
```suggestion
          <MoreActionsMenu label="Router actions" actions={actions} />
```

app/pages/project/vpcs/RouterRoutePage.tsx
```diff
@@ -0,0 +1,212 @@
...
+    .replace('internet gateway', 'gateway')
+    .replace('subnet', 'VPC subnet')
+  return value ? (
+    <TypeValueCell key={`${type}|value`} type={typeString} value={value} />
+  ) : (
+    <Badge>{type}</Badge>
+  )
+}
+
+export function RouterRoutePage() {
```
david-crespo, line 69: This component (and file) should be called `RouterPage` because it's the router detail.

mock-api/msw/handlers.ts
```diff
@@ -1053,6 +1053,78 @@ export const handlers = makeHandlers({
...
+  vpcRouterRouteUpdate({ body, path, query }) {
+    const route = lookup.vpcRouterRoute({ ...path, ...query })
+    if (body.destination) {
+      route.destination = body.destination
+    }
+    if (body.target) {
+      route.target = body.target
+    }
+    return route
+  },
```
david-crespo, line 1122: This should handle name and description too. I looked through other update endpoints and I wish we had a less manual way of doing these, but it's fine.

mock-api/vpc.ts
```diff
@@ -45,6 +45,97 @@ export const vpc2: Json<Vpc> = {
...
+    name: 'default-v4',
+    description: 'The default route of a vpc',
+    kind: 'default',
+    target: {
+      type: 'internet_gateway',
+      value: 'outbound',
+    },
+    destination: {
+      type: 'ip_net',
+      value: '0.0.0.0/0',
```
david-crespo, line 104: This is a real nitpick, but this would be more realistic-looking as `192.168.1.0/24` and the v6 as `2001:db8:abcd:12::/64`.

david-crespo:
You forgot to `git add` the new one when you moved the file.

david-crespo (changes requested):
Another pile of comments, some of which are marked outdated because of `RouterRoutePage` -> `RouterPage` move

app/forms/vpc-router-route-create.tsx
```diff
@@ -0,0 +1,97 @@
...
+      queryClient.invalidateQueries('vpcRouterRouteList')
+      addToast({ content: 'Your route has been created' })
+      onDismiss()
+    },
+  })
+
+  const form = useForm({ defaultValues })
+  const targetType = form.watch('target.type')
+
+  const onChangeTargetType = (value: string | null | undefined) => {
```
david-crespo, line 49: The nice thing about inlining callbacks like this is you don't have to annotate the argument type

charliepark: Great point!

charliepark:
Updated screenshots in PR description with latest iterations

david-crespo (approved):
Whew! Thanks for all the polish. I think this is ready to go!

PR #2360, opened 2024-08-08
Poll instance state when instance is transitioning

### LINKED ISSUE #2095: Poll instance state
Very apt and evocative description of the problem from @iliana:

> something about the flow of starting an existing instance, waiting for it to go from starting -> running so that i can successfully look at the serial console, or waiting for it to go from stopping -> stopped so i can start it again or delete it is kind of frustrating; i find myself repeatedly refreshing the page at the browser level because i don't see a button to refresh the data in the console, and also because there seems to be a lag time between when an instance is actually stopped, and when the console knows.

I have experienced this myself. I think the fix here is simply to poll, and make very clear when we are polling. If we are not clear enough, then users will expect every page to be polling live or at least be unsure whether to expect it. 

A maybe less nerve-wracking fix would be to sprinkle more refresh buttons around, like we have on the instance list. We could also stand to have a better UI treatment when you click refresh, because when I click it and nothing changes, I feel like maybe nothing happened. In places we have space, we could add a little message next to the button saying when it was last refreshed, or maybe we could do that in a tooltip, which would not have any space requirements.

image

david-crespo:
Closes #2095

Works ok on instance detail:

https://github.com/user-attachments/assets/589c96a6-3f58-410b-9d8e-132371414150

Just a draft but we can make some progress here without making big choices. Currently I'm polling every second when the instance is starting, creating, stopping, or rebooting, and not polling otherwise, but that could trivially be changed to poll in both cases but poll faster in a transitioning state.

When I tried to do this on the instance list view, I noticed a very annoying issue thanks to the mock instance we have sitting ("stuck" in the sense that we have no logic in place to get it to change) in `starting` — every poll request/response causes the whole table to render, which means that any dropdown menu you have open closes on you. This is totally unusable, and is even worse when you consider that an instance may genuinely be stuck in one of these transitioning states. We need to figure out how to keep that menu open through the render. It shouldn't be too bad. It doesn't happen with the corner action menu on instance detail for the starting instance even though that should also be rendering every time, so that bodes well.

https://github.com/user-attachments/assets/0bdfc92a-93ec-4388-b2d6-6b9d949f9012

### DIFF
--- a/app/pages/project/instances/InstancesPage.tsx
+++ b/app/pages/project/instances/InstancesPage.tsx
@@ -9,10 +9,14 @@ import { createColumnHelper } from '@tanstack/react-table'
 import { useMemo } from 'react'
 import { useNavigate, type LoaderFunctionArgs } from 'react-router-dom'
 
-import { apiQueryClient, usePrefetchedApiQuery, type Instance } from '@oxide/api'
+import {
+  apiQueryClient,
+  useApiQueryClient,
+  usePrefetchedApiQuery,
+  type Instance,
+} from '@oxide/api'
 import { Instances16Icon, Instances24Icon } from '@oxide/design-system/icons/react'
 
-import { instanceTransitioning } from '~/api/util'
 import { DocsPopover } from '~/components/DocsPopover'
 import { RefreshButton } from '~/components/RefreshButton'
 import { getProjectSelector, useProjectSelector, useQuickActions } from '~/hooks'
@@ -26,7 +30,6 @@ import { CreateLink } from '~/ui/lib/CreateButton'
 import { EmptyMessage } from '~/ui/lib/EmptyMessage'
 import { PageHeader, PageTitle } from '~/ui/lib/PageHeader'
 import { TableActions } from '~/ui/lib/Table'
-import { useInterval } from '~/ui/lib/use-interval'
 import { docLinks } from '~/util/links'
 import { pb } from '~/util/path-builder'
 
@@ -52,10 +55,11 @@ InstancesPage.loader = async ({ params }: LoaderFunctionArgs) => {
   return null
 }
 
-const refetchInstances = () => apiQueryClient.invalidateQueries('instanceList')
-
 export function InstancesPage() {
   const { project } = useProjectSelector()
+
+  const queryClient = useApiQueryClient()
+  const refetchInstances = () => queryClient.invalidateQueries('instanceList')
 
   const makeActions = useMakeInstanceActions(
     { project },
@@ -64,13 +68,6 @@ export function InstancesPage() {
 
   const { data: instances } = usePrefetchedApiQuery('instanceList', {
     query: { project, limit: PAGE_SIZE },
-  })
-
-  // if any instance in the list is transitioning, poll
-  // TODO: figure out this logic. polling if any instance is transitioning might be excessive
-  useInterval({
-    fn: () => apiQueryClient.invalidateQueries('instanceList'),
-    delay: instances.items.some(instanceTransitioning) ? 1000 : null,
   })
 
   const navigate = useNavigate()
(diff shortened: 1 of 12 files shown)


app/pages/project/instances/instance/InstancePage.tsx
```diff
@@ -99,6 +101,11 @@ export function InstancePage() {
     query: { project: instanceSelector.project },
   })
 
+  useInterval({
+    fn: () => apiQueryClient.invalidateQueries('instanceView'),
+    delay: instanceTransitioning(instance) ? 1000 : null,
+  })
+
```
david-crespo, line 108: We're doing it this way instead of `refetchInterval` on the `useApiQuery` call because of this: if that call is what defines the instance, it can't also rely on the instance. We can probably hack around that by making two calls, passing the `instance` from the first to the second, but that is somewhat appalling. `useInterval` is more flexible probably. For example, it makes it easy to use the `refreshData` function as the callback if we want. I kind of want to avoid that because it makes 5 calls instead of one.

image

david-crespo:
Fixed it in 2a5fc12ca17b3125944969db5c20e0b27b51a062. 

https://github.com/user-attachments/assets/e9be7a39-3ce1-4d68-a51b-5dfbf385664d

TIL the `useMutation` result is not referentially stable. We can eliminate some unnecessary renders in a few spots.

image

app/pages/project/instances/InstancesPage.tsx
```diff
@@ -70,6 +66,13 @@ export function InstancesPage() {
     query: { project, limit: PAGE_SIZE },
   })
 
+  // if any instance in the list is transitioning, poll
+  // TODO: figure out this logic. polling if any instance is transitioning might be excessive
+  useInterval({
+    fn: () => apiQueryClient.invalidateQueries('instanceList'),
+    delay: instances.items.some(instanceTransitioning) ? 1000 : null,
+  })
+
```
david-crespo, line 75: Not sure if polling when any instance is transitioning is a good idea. Might make more sense to just always poll but on a slower interval like 10s.

david-crespo:
I realized it's extremely easy to poll on the serial console page and connect as soon as it's running. It seems to work. I want to test how it behaves when you stop the instance.

https://github.com/user-attachments/assets/3f74b83f-9601-4bac-92cc-4618e02bcb3f

app/pages/project/instances/actions.tsx
```diff
@@ -132,6 +138,13 @@ export const useMakeInstanceActions = (
...
     },
-    [projectSelector, deleteInstance, navigate, rebootInstance, startInstance, stopInstance]
+    [
+      project,
+      navigate,
+      deleteInstanceAsync,
+      rebootInstance,
+      startInstance,
+      stopInstanceAsync,
+    ]
```
david-crespo, line 148: The changes in this file are no longer strictly needed for this PR because I've taken the polling on the list page back out, but they are still very good to do.

app/pages/project/instances/instance/InstancePage.tsx
david-crespo: No longer the case! I figured out how to do look at the instance data in a `refetchInterval` callback. It was hidden in `query.state`.

david-crespo:
https://github.com/user-attachments/assets/d94594c0-fc36-45cc-a7fa-387008d0f998

app/ui/styles/components/spinner.css
```diff
@@ -46,6 +46,10 @@
   stroke: var(--content-default);
 }
 
+.spinner-secondary .path {
+  stroke: var(--content-secondary);
+}
+
```
david-crespo, line 52: Decided against using this, but this fixes the `variant="secondary"` spinner to make the brighter part gray instead of green.

image

PR #2364, opened 2024-08-12
instance create: focus image field on validation error

### LINKED ISSUE #2345: Need global feedback on instance form validation error
When you try to make an instance without selecting an image (for example) the form refuses to submit and there is an error on the image field, but if you're scrolled down past the image field, you don't see _anything_. The form just refuses to submit. 

image

david-crespo:
Closes #2345 

RHF automatically focuses the first field with an error, but we were not passing the `ref` for the image select listbox through to the button. And it must be the button specifically because the wrapping divs are not focusable unless they have a tabindex.

https://github.com/user-attachments/assets/cebc763c-fdfd-416d-bd4c-f0ec69501986

### DIFF
--- a/app/components/form/fields/ListboxField.tsx
+++ b/app/components/form/fields/ListboxField.tsx
@@ -79,7 +79,7 @@ export function ListboxField<
         name={name}
         hasError={fieldState.error !== undefined}
         isLoading={isLoading}
-        ref={field.ref}
+        buttonRef={field.ref}
       />
       <ErrorMessage error={fieldState.error} label={label} />
     </div>
--- a/app/components/form/fields/ListboxField.tsx
+++ b/app/components/form/fields/ListboxField.tsx
@@ -79,6 +79,7 @@ export function ListboxField<
         name={name}
         hasError={fieldState.error !== undefined}
         isLoading={isLoading}
+        ref={field.ref}
       />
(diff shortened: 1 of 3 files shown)


david-crespo:
WHAT ARE YOU EVEN TALKING ABOUT

image

david-crespo:
Ok, whew. I was able to repro locally after `rm -rf node_modules`. Still a bit of a wtf.

app/components/form/fields/ListboxField.tsx
```diff
@@ -79,6 +79,7 @@ export function ListboxField<
         name={name}
         hasError={fieldState.error !== undefined}
         isLoading={isLoading}
+        buttonRef={field.ref}
```
david-crespo, line 82: Something I noticed while looking at this is that even after adding `ref`, we are not giving all the props from `field` to `Listbox`. We are still missing `onBlur` and `disabled`, both of which could become more UX-relevant in the future. For example, if we change a form from the RHF default of validating on submit to validating on blur, the lack of `onBlur` here will be a problem. Similarly, if we do #1895, I think the lack of disabled here will be a problem. We can fix it here, but I'm guessing there are a bunch more cases like this.

charliepark (approved):

PR #2366, opened 2024-08-12
Focus name field on API name already exists error

david-crespo:
Learned about `shouldFocus: true` while reading the docs to understand #2364. Trivial improvement. It's even tested.

### Before

https://github.com/user-attachments/assets/19009f20-9de2-4a95-9d44-410a737a6f35

### After

https://github.com/user-attachments/assets/5df55e4b-23f6-498a-9f7d-12a44a11e7ca

### DIFF
--- a/app/components/form/SideModalForm.tsx
+++ b/app/components/form/SideModalForm.tsx
@@ -56,7 +56,7 @@ type SideModalFormProps<TFieldValues extends FieldValues> = {
  * slightly awkward but it also makes some sense. I do not believe there is
  * any way to distinguish between fresh pageload and back/forward.
  */
-export function useShouldAnimateModal() {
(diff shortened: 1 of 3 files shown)


app/components/form/SideModalForm.tsx
```diff
@@ -56,7 +56,7 @@ type SideModalFormProps<TFieldValues extends FieldValues> = {
  * slightly awkward but it also makes some sense. I do not believe there is
  * any way to distinguish between fresh pageload and back/forward.
  */
-export function useShouldAnimateModal() {
+function useShouldAnimateModal() {
```
david-crespo, line 59: this wasn't being used elsewhere, and it being exported messes up HMR

charliepark (approved):

PR #2370, opened 2024-08-14
Add docs popover for SSH Keys

### LINKED ISSUE #2211: Docs popover on SSH keys page
Not sure we have a good doc to link to, so this may be conditional on making such a doc.

charliepark:
Fixes #2211 

Screenshot 2024-08-14 at 2 29 16 PM

Am 100% open to revamping this copy.

### DIFF
--- a/app/pages/settings/SSHKeysPage.tsx
+++ b/app/pages/settings/SSHKeysPage.tsx
@@ -85,7 +85,7 @@ export function SSHKeysPage() {
           heading="SSH key management"
           icon={<Key16Icon />}
           summary="SSH Keys are used for authentication when connecting to your instances."
-          links={[docLinks.sshKeys]}
+          links={[docLinks.sshKeyCreate, docLinks.sshKeyManage]}
         />
       </PageHeader>
       <TableActions>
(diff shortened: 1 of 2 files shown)


charliepark:
Slight tweak:
Screenshot 2024-08-14 at 2 36 05 PM

app/pages/settings/SSHKeysPage.tsx
```diff
@@ -79,6 +81,12 @@ export function SSHKeysPage() {
     <>
       <PageHeader>
         <PageTitle icon={<Key24Icon />}>SSH Keys</PageTitle>
+        <DocsPopover
+          heading="SSH key management"
```
david-crespo, line 85: I'd just say "SSH keys"

app/util/links.ts
```diff
@@ -27,6 +27,9 @@ export const links = {
     'https://docs.oxide.computer/guides/architecture/service-processors#_server_sled',
   snapshotsDocs:
     'https://docs.oxide.computer/guides/managing-disks-and-snapshots#_snapshots',
+  sshKeysDocs: 'https://docs.oxide.computer/guides/user-settings#_ssh_keys',
+  sshKeyCreateDocs: 'https://docs.oxide.computer/guides/user-settings#create-ssh-key',
+  sshKeyManageDocs: 'https://docs.oxide.computer/guides/user-settings#manage-ssh-keys',
```
david-crespo, line 32: I don't think we need to link to two sections of the same page

david-crespo (approved):
Looks good. Let's keep an eye on the docs site in case we shuffle the pages around before v10. Angela and I talked about that a bit at sync IIRC.

charliepark:
Sounds good. Will also let her know about this PR.

PR #2382, opened 2024-08-19
Bump omicron for oxql api tweak

david-crespo:
No big deal.


app/api/window.ts
```diff
@@ -42,7 +42,7 @@ if (typeof window !== 'undefined') {
   window.oxql = {
     query: async (q: string) => {
       const result = await api.methods.timeseriesQuery({ body: { query: q } })
-      const data = handleResult(result)
+      const data = handleResult(result).tables
```
david-crespo, line 45: @bnaecker this is the only change we had to make since we're not using OxQL in app code. Though even if we were it would be very similar to this. And the typechecker pointed it out.

PR #2389, opened 2024-08-19
Instances table: remove hostname, normal row height, split CPU/RAM

### LINKED ISSUE #2357: Remove hostname column from instances table
Hostname is not useful. External IP is useful. Could use the `ListPlusCell` for multiple IPs.

david-crespo:
Closes #2357

We were giving `hostname` way too much prominence considering it's not really doing anything.

### Before

image

### After

image

### DIFF
--- a/app/components/TimeAgo.tsx
+++ b/app/components/TimeAgo.tsx
@@ -26,10 +26,8 @@ export const TimeAgo = ({
     </div>
   )
   return (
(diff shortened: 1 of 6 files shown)


benjaminleonard:
Looks good to me

david-crespo:
It's really surprising to me that changing this extremely key table did not cause any e2e tests to fail, so I'm going to add one.

PR #2391, opened 2024-08-20
Polling on instance list

### LINKED ISSUE #2377: Poll instance list
I ended up leaving this out of #2360 because I felt it was a little aggressive to always poll if any instance on the current page is in a transitional state. However, after you see auto-polling on the instance detail page, it feels like a big lack on the instance list, particularly when you start or stop an instance from that page. It changes to `stopping` and then sits there, which feels quite useless.

image

Because #2360 established a visual indicator that we are polling on instance detail (spinner next to the state badge), I feel more comfortable being clever in our logic around when to poll. Some ways it could work (not mutually exclusive):

* Poll for up to N seconds after an action if any instance in scope is transitioning
* Poll for up to N seconds after initial pageload if any instance in scope is transitioning
* Increase polling interval after a certain amount of time polling

See also #2376 for fancier polling logic on instance detail.

david-crespo:
Closes #2377 

- [x] Smart logic
- [x] e2e test it
- [x] Figure out how to indicate when we're polling (don't. use "Updated" message)

### Demo

When the page loads below, it polls every 3 seconds because there is an instance in a starting state. After 30 seconds, it would switch to polling every minute, but I don't wait for that in this clip. I stop the instance, which causes it to transition to `stopping`, and the `stopped` (because of the polling). At that point, there are no instances in transitional states, so polling slows to every minute (i.e., not at all in the short time in this clip). Then when I start the instance it polls again while it's in `starting` and then stops (slows) again after it's `running`.

https://github.com/user-attachments/assets/3328ec0f-ceef-4b93-bad1-77d5b1eb5d05

### DIFF
--- a/app/pages/project/instances/InstancesPage.tsx
+++ b/app/pages/project/instances/InstancesPage.tsx
@@ -93,7 +93,7 @@ export function InstancesPage() {
       // like starting or stopping. After that, it will keep polling, but more
       // slowly. For example, if you stop an instance, its state will change to
       // `stopping`, which will cause this logic to start polling the list until
-      // it lands in `stopped`, at which point it will stop polling because
+      // it lands in `stopped`, at which point it will poll only slowly because
       // `stopped` is not considered transitional.
       refetchInterval({ state: { data } }) {
         const prevTransitioning = transitioningInstances.current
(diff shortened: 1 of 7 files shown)


charliepark (approved):
Nice!

PR #2392, opened 2024-08-20
Add dropdown to instance breadcrumb

### LINKED ISSUE #2380: Instance page breadcrumb needs a dropdown selector
When on an instance page, you should be able to select other instances from that project.

image

charliepark:
Closes #2380

We'll want to revisit the overall breadcrumb designs (see #2380 for a few possible directions / links to more in Figma), but in the interim, we can add the dropdown to the instance item in the breadcrumb nav.

Screenshot 2024-08-20 at 3 00 22 PM

Screenshot 2024-08-20 at 3 01 49 PM

### DIFF
--- a/app/pages/project/instances/InstancesPage.tsx
+++ b/app/pages/project/instances/InstancesPage.tsx
@@ -10,12 +10,7 @@ import { filesize } from 'filesize'
 import { useMemo } from 'react'
 import { useNavigate, type LoaderFunctionArgs } from 'react-router-dom'
 
-import {
-  apiQueryClient,
-  useApiQueryClient,
-  usePrefetchedApiQuery,
-  type Instance,
-} from '@oxide/api'
+import { apiQueryClient, usePrefetchedApiQuery, type Instance } from '@oxide/api'
 import { Instances16Icon, Instances24Icon } from '@oxide/design-system/icons/react'
 
 import { DocsPopover } from '~/components/DocsPopover'
@@ -55,11 +50,10 @@ InstancesPage.loader = async ({ params }: LoaderFunctionArgs) => {
   return null
 }
 
+const refetchInstances = () => apiQueryClient.invalidateQueries('instanceList')
+
 export function InstancesPage() {
   const { project } = useProjectSelector()
-
-  const queryClient = useApiQueryClient()
-  const refetchInstances = () => queryClient.invalidateQueries('instanceList')
 
   const makeActions = useMakeInstanceActions(
     { project },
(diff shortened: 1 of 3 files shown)


david-crespo:
The CI failure was a flake, but both Chrome and Safari flaked on the same test, which has to do with instance states, so it most likely was caused by #2360. There's probably a timeout I can tweak that will fix it. I'll take a look.

david-crespo (approved):

david-crespo:
Ok, this is amazing: that CI test failure is real, it's dead reproducible locally. The only thing this PR changes is it adds another query watching the `instanceList` endpoint. This might be interesting to debug.

app/pages/project/instances/instance/InstancePage.tsx
```diff
@@ -94,7 +94,10 @@ export function InstancePage() {
   const makeActions = useMakeInstanceActions(instanceSelector, {
     onSuccess: refreshData,
     // go to project instances list since there's no more instance
-    onDelete: () => navigate(pb.instances(instanceSelector)),
+    onDelete: () => {
+      apiQueryClient.invalidateQueries('instanceList')
+      navigate(pb.instances(instanceSelector))
+    },
```
david-crespo, line 100: This was a legit bug, and I know why this change turned it up. The failure was that the expected disappearance of `you-fail` from the instance list was not happening.

What happens in this instance delete test is 

1. go to the instance detail page for `you-fail`
2. delete it
3. let it automatically land us on the list view
4. confirm the instance is gone
 
Before this PR, there was nothing on the instance detail page "observing" the `instanceList` query, so when we nav to instance list on delete success, that query will always be fetched fresh because it is "new" from RQ's point of view. But with this PR, there is now the breadcrumb that uses the list of instances, so the query is "active" on the instance detail page.

Here is the funny bit: when you do what the test does _by hand_, you inevitably sit on the `you-fail` page longer than the 2 seconds required for the query to become stale. That means it will automatically refetch when the loader fires for the instance list page on navigation. But the test is faster than you, so it lands on the instance detail and hits delete in under 2 seconds, so the query is _not_ stale by the time we land on the instance list and therefore does not refetch. I confirmed this theory by adding a sleep to the test so it sits on the page for 2 seconds before clicking delete. That makes the test pass! But of course the real fix is to specifically invalidate the instance list in the delete success callback.

app/pages/project/instances/InstancesPage.tsx
```diff
@@ -55,12 +50,11 @@ InstancesPage.loader = async ({ params }: LoaderFunctionArgs) => {
...
 }
 
+const refetchInstances = () => apiQueryClient.invalidateQueries('instanceList')
+
 export function InstancesPage() {
   const { project } = useProjectSelector()
 
-  const queryClient = useApiQueryClient()
-  const refetchInstances = () => queryClient.invalidateQueries('instanceList')
-
```
david-crespo, line 63: The changes in this file are irrelevant to the fix, but I did this while debugging and I like it better.

app/pages/project/instances/instance/InstancePage.tsx
charliepark: Oh, wow. Good sleuthing on this.

PR #2393, opened 2024-08-21
Choose custom router on subnet create/edit forms

### LINKED ISSUE #2379: Subnet: Show linked custom router
The optional `custom_router_id` attribute is a key piece of information for the VPC router feature. It'll be very useful to show the name of the router in the subnet table.

charliepark:
I'm still getting the forms to submit correctly, so this isn't where it needs to be for review, yet.

But here's the current state:
Screenshot 2024-08-21 at 9 53 36 AM

Fixes #2379


app/table/cells/RouterLinkCell.tsx
```diff
@@ -0,0 +1,36 @@
...
+    if (!value) return <EmptyCell />
+    // probably not possible but let’s be safe
+    if (isError) return <Badge color="neutral">Deleted</Badge>
+    if (!subnet) return <SkeletonCell /> // loading
+    return (
+      <LinkCell to={pb.vpcRouter({ project, vpc, router: subnet.name })}>
+        {subnet.name}
+      </LinkCell>
+    )
+  }, [value, project, vpc, isError, subnet])
```
david-crespo, line 35: No need for `useMemo` here. In the `useCustomRouterItems` case where I recommended it, the problem is that without memoization, the `items` array is new on every render, which means it will always cause a re-render of the component it's passed into. In this case, you're just rendering this cell and nothing is messed up by it not being referentially stable. The result is basically the same as if you didn't have the `useMemo`, so `useMemo` actually _adds_ overhead (however minimal it may be).

> [So when should I useMemo and useCallback?](https://kentcdodds.com/blog/usememo-and-usecallback#so-when-should-i-usememo-and-usecallback)
> 1. Referential equality
> 2. Computationally expensive calculations

app/table/cells/RouterLinkCell.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { EmptyCell, SkeletonCell } from './EmptyCell'
+import { LinkCell } from './LinkCell'
+
+export const RouterLinkCell = ({ value }: { value?: string }) => {
+  const { project, vpc } = useVpcSelector()
+  const { data: subnet, isError } = useApiQuery('vpcRouterView', {
+    path: { router: value || '' },
+    query: { project, vpc },
+  })
```
david-crespo, line 24: how to handle the non-existing router

https://github.com/oxidecomputer/console/blob/8dcddcef62b8d10dfcd3adb470439212b23b3d5e/app/table/cells/InstanceLinkCell.tsx#L19-L23

mock-api/msw/handlers.ts
```diff
@@ -1159,7 +1160,10 @@ export const handlers = makeHandlers({
     const newSubnet: Json<Api.VpcSubnet> = {
       id: uuid(),
       vpc_id: vpc.id,
-      ...body,
+      name: body.name,
+      description: body.description || '',
+      ipv4_block: body.ipv4_block || '',
+      custom_router_id: body.custom_router || '',
```
david-crespo, line 1166: I don't think any of these should fall back to `''` — description and ipv4_block are always strings already and the other one should stay `undefined` if it's `undefined`.

test/e2e/vpcs.e2e.ts
```diff
@@ -45,34 +45,95 @@ test('can create and delete subnet', async ({ page }) => {
...
-  await expect(rows.nth(0).getByText('mock-subnet')).toBeVisible()
-  await expect(rows.nth(0).getByText('10.1.1.1/24')).toBeVisible()
+  await dialog.getByRole('textbox', { name: 'Name' }).fill('mock-subnet-2')
+  await dialog.getByRole('textbox', { name: 'IPv4 block' }).fill('10.1.1.2/24')
 
-  await expect(rows.nth(1).getByText('mock-subnet-2')).toBeVisible()
-  await expect(rows.nth(1).getByText('10.1.1.2/24')).toBeVisible()
+  // little hack to catch a bug where we weren't handling empty input here properly
+  await dialog.getByRole('textbox', { name: 'IPv6 block' }).fill('abc')
+  await dialog.getByRole('textbox', { name: 'IPv6 block' }).clear()
```
david-crespo, line 69: Confirmed this bit breaks the test without the fix for the `ipv6Block` field.

image

david-crespo (approved):
I think this is solid. I like the string sentinel value pattern for a clearable listbox. It might make sense to build this logic into the listbox field itself somehow, so we don't have to manually wire up the conversion between form values and JSON values. On the other hand, it seems like the way we're doing it is kinda how you're supposed to do it — RHF doesn't seem to like it when you try to store the processed maybe-not-a-string value in the form state.

PR #2396, opened 2024-08-26
Mock API: error if parent selectors present when resource is fetched by ID

david-crespo:
Trying to avoid the bug fixed in https://github.com/oxidecomputer/console/pull/2394, where we accidentally include parent selectors. At first I thought we should figure out how to do this automatically, hence https://github.com/oxidecomputer/oxide.ts/issues/264, but as noted there, it's hard to do automatically, and I remembered we already have this whole manually-implemented mock DB lookup apparatus in the console. We have run into this category of bug once before, and I fixed it in precisely this way in https://github.com/oxidecomputer/console/pull/2203, but only for one resource, which is silly.

The mechanical stuff here was pretty easy. The annoying part was going through the real API and code fixing a couple of spots where we currently deviate.

### DIFF
--- a/mock-api/msw/db.ts
+++ b/mock-api/msw/db.ts
@@ -26,15 +26,31 @@ export const notFoundErr = (msg: string) => {
   return json({ error_code: 'ObjectNotFound', message } as const, { status: 404 })
 }
 
-export const badSelectorErr = (resource: string, parents: string[]) => {
-  const message = `when ${resource} is specified by ID, ${commaSeries(parents, 'and')} should not be specified`
-  return json({ error_code: 'InvalidRequest', message }, { status: 400 })
-}
-
 export const lookupById = <T extends { id: string }>(table: T[], id: string) => {
   const item = table.find((i) => i.id === id)
   if (!item) throw notFoundErr(`by id ${id}`)
   return item
+}
+
+/**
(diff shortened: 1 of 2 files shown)


mock-api/msw/db.ts
```diff
@@ -168,7 +205,10 @@ export const lookup = {
   vpcSubnet({ subnet: id, ...vpcSelector }: PP.VpcSubnet): Json<Api.VpcSubnet> {
     if (!id) throw notFoundErr('no subnet specified')
 
-    if (isUuid(id)) return lookupById(db.vpcSubnets, id)
+    if (isUuid(id)) {
+      ensureNoParentSelectors('subnet', vpcSelector)
+      return lookupById(db.vpcSubnets, id)
+    }
```
david-crespo, line 211: There's a small downside to this implementation for double-nested resources like subnet: if you were to try to select a subnet by `(subnet name, VPC ID, project name)`, the thing that would blow up would be the VPC lookup, and the error would be "when VPC is specified by ID, project should not be specified." When I put it that way, it's not that confusing. But if you only look at the error message and not what endpoint the error is coming from, you can get thrown off.

charliepark (approved):
Thank you for fixing all of these

PR #2404, opened 2024-08-30
Truncate truncated descriptions even more

david-crespo:
In the draft v10 release notes, description was cut off in my routes screenshot, but I was pretending I didn't notice. @benjaminleonard, however, could not so pretend. He said it was "bad." Touché.

rel-9-vpc-routes

So I thought, let's try using CSS's built in truncation on top of ours. The only reason we do that to begin with is so we can detect when we're truncating and add a tooltip on hover showing the full text. But combining the two works shockingly well. The reason it's a draft is I want to go and look at all the places we call `Truncate` and make sure they look good.

https://github.com/user-attachments/assets/bbdcd452-5009-452d-9819-ecf9614fdc20

### DIFF
--- a/mock-api/vpc.ts
+++ b/mock-api/vpc.ts
@@ -48,7 +48,7 @@ export const vpcs: Json<Vpc[]> = [vpc, vpc2]
 export const defaultRouter: Json<VpcRouter> = {
   id: 'fc59fb4d-baad-44a8-b152-9a3c27ae8aa1',
   name: 'mock-system-router',
-  description: 'a fake router',
+  description: 'Routes are automatically added to this router as VPC subnets are created',
   time_created: new Date(2024, 0, 1).toISOString(),
   time_modified: new Date(2024, 0, 2).toISOString(),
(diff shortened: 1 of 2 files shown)


mock-api/vpc.ts
```diff
@@ -48,7 +48,7 @@ export const vpcs: Json<Vpc[]> = [vpc, vpc2]
 export const defaultRouter: Json<VpcRouter> = {
   id: 'fc59fb4d-baad-44a8-b152-9a3c27ae8aa1',
   name: 'mock-system-router',
-  description: 'a fake router',
+  description: 'Routes are automatically added to this router as VPC subnets are created',
```
david-crespo, line 51: Match the real default description given to system routers

david-crespo:
I went through all the usages cases and couldn't find any issues. One potential weirdness I found was if we're using the `middle` truncate position (which we only use in the file input), if something squishes the container, it will have two ellipses (the middle one and the CSS one on the end). But in a side modal form that can't happen because the width is fixed.

image

Issue #2367, opened 2024-08-13
Don't pass `useMutation` result to dep arrays


david-crespo:
As I learned in https://github.com/oxidecomputer/console/pull/2360#issuecomment-2278739819, the object returned by `useMutation` is not referentially stable — it is remade on every render. But I thought it was, so we're using it in a lot of dependency arrays. We basically need to go through every `useApiMutation` call and check if the mutation is being included in a dependency array, and change it to depend directly on the relevant property instead, like the `mutate` function (which is stable). Unfortunately I can't think of a way to lint this, so we'll just have to try to do it as goodly as we can.

https://github.com/oxidecomputer/console/blob/3a6f815ad50bf2c7fafb55dcdd31ad0232adbfeb/app/pages/system/SiloImagesPage.tsx#L73-L95

david-crespo:
Thanks @charliepark!

PR #2405, opened 2024-08-30
Move IP Pools edit form to view page

### LINKED ISSUE #2353: Add more actions menu on IP pool detail that lets you edit
Right now you can only edit from the list view. Check other resources with a detail view to see if we need to make this change, like VPC and instance.

charliepark:
This PR moves the "Edit IP pool" side modal from the IP Pools list page to the individual IP Pool view page. In some ways, this makes more sense with our existing URL schema, as the edit URL is `http://localhost:4000/system/networking/ip-pools/:pool/edit`, so is conceptually already a child of the pool itself, rather than the list.

We start with an additional actions button in the header …
Screenshot 2024-08-30 at 11 07 07 AM
… which leads to the side modal on the view page …
Screenshot 2024-08-30 at 11 07 30 AM

The `delete` button is disabled for IP Pools with existing ranges …
Screenshot 2024-08-30 at 11 07 20 AM
… but is enabled for IP Pools that can, in fact, be deleted:
Screenshot 2024-08-30 at 11 07 56 AM

Closes #2353, and if we feel this is a good pattern to replicate, we can roll it out to other resources as well.

### DIFF
--- a/app/forms/ip-pool-edit.tsx
+++ b/app/forms/ip-pool-edit.tsx
@@ -40,10 +40,11 @@ export function EditIpPoolSideModalForm() {
     onSuccess(_pool) {
       if (pool.name !== _pool.name) {
         queryClient.invalidateQueries('ipPoolList')
+        // as the pool's name has changed, we need to navigate to an updated URL
         navigate(pb.ipPool({ pool: _pool.name }))
       } else {
         queryClient.invalidateQueries('ipPoolView')
-        navigate(pb.ipPool({ pool: pool.name }))
+        onDismiss()
       }
       addToast({ content: 'Your IP pool has been updated' })
     },
--- a/app/forms/ip-pool-edit.tsx
+++ b/app/forms/ip-pool-edit.tsx
@@ -35,6 +35,7 @@ export function EditIpPoolSideModalForm() {
   const { data: pool } = usePrefetchedApiQuery('ipPoolView', { path: poolSelector })
 
   const form = useForm({ defaultValues: pool })
+  const onDismiss = () => navigate(pb.ipPool({ pool: poolSelector.pool }))
 
   const editPool = useApiMutation('ipPoolUpdate', {
     onSuccess(_pool) {
@@ -43,13 +44,11 @@ export function EditIpPoolSideModalForm() {
         // as the pool's name has changed, we need to navigate to an updated URL
         navigate(pb.ipPool({ pool: _pool.name }))
       } else {
-        queryClient.invalidateQueries('ipPoolView')
         onDismiss()
       }
       addToast({ content: 'Your IP pool has been updated' })
     },
   })
-  const onDismiss = () => navigate(pb.ipPool({ pool: poolSelector.pool }))
 
   return (
     <SideModalForm
--- a/app/forms/ip-pool-edit.tsx
+++ b/app/forms/ip-pool-edit.tsx
@@ -39,11 +39,12 @@ export function EditIpPoolSideModalForm() {
 
   const editPool = useApiMutation('ipPoolUpdate', {
     onSuccess(_pool) {
+      queryClient.invalidateQueries('ipPoolList')
       if (pool.name !== _pool.name) {
-        queryClient.invalidateQueries('ipPoolList')
         // as the pool's name has changed, we need to navigate to an updated URL
         navigate(pb.ipPool({ pool: _pool.name }))
       } else {
+        queryClient.invalidateQueries('ipPoolView')
         onDismiss()
       }
       addToast({ content: 'Your IP pool has been updated' })
--- a/app/forms/ip-pool-edit.tsx
+++ b/app/forms/ip-pool-edit.tsx
@@ -30,22 +30,24 @@ EditIpPoolSideModalForm.loader = async ({ params }: LoaderFunctionArgs) => {
 export function EditIpPoolSideModalForm() {
   const queryClient = useApiQueryClient()
   const navigate = useNavigate()
-
   const poolSelector = useIpPoolSelector()
-
-  const onDismiss = () => navigate(pb.ipPools())
 
   const { data: pool } = usePrefetchedApiQuery('ipPoolView', { path: poolSelector })
 
+  const form = useForm({ defaultValues: pool })
+
   const editPool = useApiMutation('ipPoolUpdate', {
     onSuccess(_pool) {
-      queryClient.invalidateQueries('ipPoolList')
+      if (pool.name !== _pool.name) {
+        navigate(pb.ipPool({ pool: _pool.name }))
+      } else {
+        queryClient.invalidateQueries('ipPoolView')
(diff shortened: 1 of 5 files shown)


app/forms/ip-pool-edit.tsx
```diff
@@ -30,22 +30,25 @@ EditIpPoolSideModalForm.loader = async ({ params }: LoaderFunctionArgs) => {
...
   const editPool = useApiMutation('ipPoolUpdate', {
     onSuccess(_pool) {
-      queryClient.invalidateQueries('ipPoolList')
+      if (pool.name !== _pool.name) {
+        queryClient.invalidateQueries('ipPoolList')
+        navigate(pb.ipPool({ pool: _pool.name }))
+      } else {
+        queryClient.invalidateQueries('ipPoolView')
+        navigate(pb.ipPool({ pool: pool.name }))
+      }
```
david-crespo, line 47: would add a comment explaining why the difference. also the navigate line is the same in both cases, you can just do `navigate(pb.ipPool({ pool: _pool.name }))`

charliepark: I realized that we don't actually need the `navigate` function when the description has changed; just an onDismiss(). Have added a comment on the first `navigate` (when name has changed), though.

david-crespo: Why the difference in invalidates?

test/e2e/ip-pools.e2e.ts
```diff
@@ -133,6 +133,24 @@ test('IP pool delete', async ({ page }) => {
...
+  await page.getByRole('button', { name: 'IP pool actions' }).click()
+  await page.getByRole('menuitem', { name: 'Delete' }).click()
+  await expect(page.getByRole('dialog', { name: 'Confirm delete' })).toBeVisible()
+  await page.getByRole('button', { name: 'Confirm' }).click()
+
+  // get redirected back to the list after successful delete
+  await expect(page).toHaveURL('/system/networking/ip-pools')
+  await expect(page.getByRole('cell', { name: 'ip-pool-3' })).toBeHidden()
+})
+
```
david-crespo, line 153: wonderful

app/pages/system/networking/IpPoolPage.tsx
```diff
@@ -70,16 +73,54 @@ IpPoolPage.loader = async function ({ params }: LoaderFunctionArgs) {
 export function IpPoolPage() {
   const poolSelector = useIpPoolSelector()
   const { data: pool } = usePrefetchedApiQuery('ipPoolView', { path: poolSelector })
+  const { data: ranges } = usePrefetchedApiQuery('ipPoolRangeList', {
+    path: poolSelector,
+    query,
+  })
+  const navigate = useNavigate()
+  const deletePool = useApiMutation('ipPoolDelete', {
```
david-crespo, line 81: See https://github.com/oxidecomputer/console/issues/2367. Do

```ts
const { mutateAsync: deletePool } = useApiMutation(...)
```

and change `deletePool.mutateAsync()` to `deletePool()` accordingly. Unfortunately, simply changing `deletePool` to `deletePool.mutateAsync` in the dep array doesn't satisfy the linter.

david-crespo (approved):
Ah, I get it! I didn't realize we were already using the perfect route on the list view, which was weird. So we can keep the edit button on the list, and it goes to the same route, but that route now shows the detail view with the form. Very nice. I do like this pattern. I think we should do the same to the other pages in a separate PR.

Looks good, just one change suggested, and the `onSuccess` thing is still not clear enough on the way.

Something we can look at in a followup is making these real links where possible like I did in https://github.com/oxidecomputer/console/pull/2343 for the settings link. It's more accessible and lets you middle click the links to open in a new tab. The annoying part is that we're using helpers to construct the row actions and the more actions dropdowns, so those need to modified to allow a different kind of item that just has a URL instead of an `onActivate` callback. I tried that in https://github.com/oxidecomputer/console/pull/2351 and it didn't feel worth it for like 3 links, but now there will be more.

app/forms/ip-pool-edit.tsx
charliepark: We can maybe drop the View one; will walk through the various elements of it.

The `ipPoolList` invalidation is needed because of the presence of the IP pools in the header nav. Without that invalidation, even though the `navigate` takes us to the new page and the header is correct, the dropdown has old data (this shows `ip-pool-1` renamed to `ip-pool-updated`, _without_ the invalidation:
Screenshot 2024-08-30 at 2 26 47 PM

Adding `queryClient.invalidateQueries('ipPoolList')` corrects that:
Screenshot 2024-08-30 at 2 27 22 PM

The other invalidation would be useful if the page showed the description (the only form fields in the "Edit IP pool" form are name and description), but since the description is _not_ shown on the IP Pool page, we can actually just remove that invalidation. We'd need to add it back in if we decide down the road to show the description on this page.

david-crespo: If so, isn't the `ipPoolList` invalidation needed regardless of whether the name changed? And on the view one, I think we should always do the corresponding invalidation even if we aren't relying on directly yet. Leaving out an invalidation was how we got this head-scratcher https://github.com/oxidecomputer/console/pull/2392#discussion_r1724068821.

charliepark: Sure, getting the View invalidation back in makes sense, in case we add descriptions in there.

For List, my thought was that it was only needed when the name changed, as the description changing didn't impact that navigation dropdown, so it wouldn't be necessary to invalidate that; every time I manually checked the list page after editing the description it seemed to show the updated description state in the table of IP pools. But perhaps there's an aspect of the List invalidation I'm missing. Will move the it up a level.

app/forms/ip-pool-edit.tsx
```diff
@@ -30,23 +30,27 @@ EditIpPoolSideModalForm.loader = async ({ params }: LoaderFunctionArgs) => {
...
   const editPool = useApiMutation('ipPoolUpdate', {
     onSuccess(_pool) {
       queryClient.invalidateQueries('ipPoolList')
+      if (pool.name !== _pool.name) {
+        // as the pool's name has changed, we need to navigate to an updated URL
+        navigate(pb.ipPool({ pool: _pool.name }))
+      } else {
+        queryClient.invalidateQueries('ipPoolView')
+        onDismiss()
+      }
```
david-crespo, line 49: I think this is equivalent.

```suggestion
      queryClient.invalidateQueries('ipPoolList')
      if (pool.name === _pool.name) {
        queryClient.invalidateQueries('ipPoolView')
      }
      navigate(pb.ipPool({ pool: _pool.name }))
```

Is the reason for the conditional that you were getting that error flash on nav? Otherwise this would work, wouldn't it?

```ts
queryClient.invalidateQueries('ipPoolList')
queryClient.invalidateQueries('ipPoolView')
navigate(pb.ipPool({ pool: _pool.name }))
```

Or maybe navigate first I guess?

PR #2409, opened 2024-09-03
Tweak IP pool edit `onSuccess` logic

david-crespo:
Followup for https://github.com/oxidecomputer/console/pull/2405#discussion_r1739475703, which I posted seconds after merge. I made the conditional narrower and added a comment.

### DIFF
--- a/app/forms/ip-pool-edit.tsx
+++ b/app/forms/ip-pool-edit.tsx
@@ -35,19 +35,21 @@ export function EditIpPoolSideModalForm() {
   const { data: pool } = usePrefetchedApiQuery('ipPoolView', { path: poolSelector })
 
   const form = useForm({ defaultValues: pool })
-  const onDismiss = () => navigate(pb.ipPool({ pool: poolSelector.pool }))
 
   const editPool = useApiMutation('ipPoolUpdate', {
-    onSuccess(_pool) {
+    onSuccess(updatedPool) {
       queryClient.invalidateQueries('ipPoolList')
-      if (pool.name !== _pool.name) {
-        // as the pool's name has changed, we need to navigate to an updated URL
-        navigate(pb.ipPool({ pool: _pool.name }))
-      } else {
-        queryClient.invalidateQueries('ipPoolView')
-        onDismiss()
-      }
+      navigate(pb.ipPool({ pool: updatedPool.name }))
       addToast({ content: 'Your IP pool has been updated' })
+
+      // Only invalidate if we're staying on the same page. If the name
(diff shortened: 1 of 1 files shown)


app/forms/ip-pool-edit.tsx
```diff
@@ -35,19 +35,21 @@ export function EditIpPoolSideModalForm() {
...
+
+      // Only invalidate if we're staying on the same page. If the name
+      // _has_ changed, invalidating ipPoolView causes an error page to flash
+      // while the loader for the target page is running because the current
+      // page's pool gets cleared out while we're still on the page. If we're
+      // navigating to a different page, its query will fetch anew regardless.
+      if (pool.name === updatedPool.name) {
         queryClient.invalidateQueries('ipPoolView')
-        onDismiss()
       }
```
david-crespo, line 52: I found the if-else confusing because the `onDismiss` and `navigate` do exactly the same thing

app/forms/ip-pool-edit.tsx
```diff
@@ -35,19 +35,21 @@ export function EditIpPoolSideModalForm() {
...
-      } else {
+      navigate(pb.ipPool({ pool: updatedPool.name }))
+      addToast({ content: 'Your IP pool has been updated' })
+
+      // Only invalidate if we're staying on the same page. If the name
+      // _has_ changed, invalidating ipPoolView causes an error page to flash
+      // while the loader for the target page is running because the current
+      // page's pool gets cleared out while we're still on the page. If we're
+      // navigating to a different page, its query will fetch anew regardless.
+      if (pool.name === updatedPool.name) {
```
david-crespo, line 50: Another way of doing this that's worth considering would be to add the ability to `invalidateQueries` to invalidate more narrowly than an entire endpoint.

https://github.com/oxidecomputer/console/blob/c648c44cc3b8c7daea7d85414a463caa2d6b1149/app/api/hooks.ts#L243-L253

So in this case we'd be able to avoid the conditional by invalidating the fetch under the old name, but not the new name. Working on a quick PR showing that option. I thought it was kind of dicey but it's not.

app/forms/ip-pool-edit.tsx
charliepark: Ah, good point.

charliepark (approved):
Cleaner. Also agree that `updatedPool` is clearer.

app/forms/ip-pool-edit.tsx
charliepark: As I mentioned in the other PR, I prefer (w/o strong feelings on it) the way you have it in this PR. But open to either direction.

PR #2410, opened 2024-09-03
Add `params` arg to `invalidateQueries`

david-crespo:
This feels slightly worse somehow. I have liked not having the `params` arg on `invalidateQueries`.


charliepark:
Personally, the `if` at the callsite helps highlight the intent of the code a bit more clearly to me, and I do like it without the `params` prop. That being said, I don't have strong feelings on it.

PR #2411, opened 2024-09-03
Rename "remove" -> "delete" for IP pool ranges

david-crespo:
I'm not married to this, it's a random thing I had on my to-do list from a while ago. Technically the endpoint is `ip_pool_range_remove` and `.../remove`, so arguably we should match the API so people know what to look for in the docs on the off chance they're trying to do this by API or CLI. On the other hand, to me "remove" suggests that it still exists somehow but not associated with the pool, and this is not the case. I'm on the fence.

### Before

image

### After

image

### DIFF
--- a/app/pages/system/networking/IpPoolPage.tsx
+++ b/app/pages/system/networking/IpPoolPage.tsx
@@ -208,7 +208,7 @@ function IpRangesTable() {
   const makeRangeActions = useCallback(
     ({ range }: IpPoolRange): MenuAction[] => [
       {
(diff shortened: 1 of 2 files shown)


charliepark:
No strong feelings, but a slight preference on my end for remove, as, conceptually, how do you delete a range of numbers? (okay, that's IPv4-brain, but.) But ask me again tomorrow and I'll probably go the other way on it.

david-crespo:
Yeah, I was lukewarm on it anyway. Thanks!

PR #2412, opened 2024-09-04
More helpful empty projects table text

david-crespo:
Thought of this while listening to a podcast about UI design. The system works!

Screenshot 2024-09-04 at 4 09 52 PM

### DIFF
--- a/app/pages/ProjectsPage.tsx
+++ b/app/pages/ProjectsPage.tsx
@@ -37,7 +37,7 @@ const EmptyState = () => (
   <EmptyMessage
     icon={<Folder24Icon />}
     title="No projects"
-    body="Create a project to see it here"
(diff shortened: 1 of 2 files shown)


benjaminleonard:
Great!

We usually omit the full stop if there's just one sentence right?

david-crespo:
Usually, though mentally I have a sort of line where if it gets too long it starts to feel like it needs a period. Generally the rule should be: if it's a sentence, give it a period. We've been leaving it out because most of these are short, but tbh the existing ones look a bit silly too.

image

benjaminleonard:
What can I say, I'm a minimalist

david-crespo:
Fine, I can deal.

image

david-crespo:
Fine, I can deal.

Screenshot 2024-09-04 at 4 48 58 PM

PR #2414, opened 2024-09-04
Add table row action to copy silo ID from system utilization page

### LINKED ISSUE #2344: "Copy ID" on Utilization page
I find the "Copy ID" and the displayed ids useful when working between multiple interfaces (web + api or web + cli). It would be a nice to have if the `/system/utilization` page had something like the "Copy ID" button that the `/system/silos` page has. The context row is a little different here, so maybe it would have to read "Copy Silo ID".

charliepark:
This adds a column to the `system/utilization` page, allowing users to copy the ID of the silo on each row. As this table is constructed a bit differently than our other tables where we pass the columns in, I extracted some existing code into a RowActions component that we can call independently (as we do in this PR). The new component also has an optional prop for `copyIdLabel`, in case we want to use more specific copy than "Copy ID" (as we do in this PR).

Screenshot 2024-09-04 at 4 17 15 PM

Closes #2344

Minor sidenote: We had an existing "did it copy to the clipboard?" test that was skipping for Safari and Firefox. It's still uncooperative for CI contexts, but Firefox is now working, [as of June 2024,](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#browser_compatibility) so we aren't skipping that one any more.

### DIFF
--- a/app/pages/system/UtilizationPage.tsx
+++ b/app/pages/system/UtilizationPage.tsx
@@ -229,7 +229,7 @@ function UsageTab() {
               />
             </Table.Cell>
             <Table.Cell className="action-col w-10 children:p-0" height="large">
-              <RowActions id={silo.siloId} copyIdText="Copy Silo ID" />
+              <RowActions id={silo.siloId} copyIdLabel="Copy Silo ID" />
             </Table.Cell>
           </Table.Row>
         ))}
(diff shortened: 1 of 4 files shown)


david-crespo (approved):
Great. The [diff with whitespace hidden](https://github.com/oxidecomputer/console/pull/2414/files?w=1) is extremely short. Nice about FF supporting the clipboard tests. Only think I'd like to see is a doc comment on the `RowActions` component and/or the `id` and label props saying something like

> If `id` is provided, a copy ID menu item will be automatically included. Use `copyIdLabel` to override the default label.

PR #2415, opened 2024-09-05
tools: try to reduce flake in image upload e2e tests

david-crespo:
Test flakes pretty much always look like this. The assert about them all being ready is sort of ill-conceived anyway.

image

### DIFF
--- a/test/e2e/image-upload.e2e.ts
+++ b/test/e2e/image-upload.e2e.ts
@@ -19,16 +19,14 @@ import {
 // to complete in time, so we just assert that they all start out ready and end
 // up complete
 async function expectUploadProcess(page: Page) {
-  const steps = page.locator('div[data-status]')
-
-  for (const step of await steps.all()) {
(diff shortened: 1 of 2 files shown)


test/e2e/image-upload.e2e.ts
```diff
@@ -234,12 +232,6 @@ test.describe('Image upload', () => {
 
       await page.click('role=button[name="Upload image"]')
 
-      const steps = page.locator('div[data-status]')
-
-      for (const step of await steps.all()) {
-        await expect(step).toHaveAttribute('data-status', 'ready')
-      }
-
```
david-crespo, line 242: When I ran this locally, it passed because `steps.all()` was coming back empty, so the whole loop is being skipped. Boo. This is due to this approach being sort of a hack — normal `expect` will [auto-wait](https://playwright.dev/docs/actionability) a bit for the thing to become visible before asserting, but this `all()` approach just gets the list of things matching at the moment it's called, and if it's empty, it's empty.

PR #2420, opened 2024-09-09
Use renamed functions in place of raw useApiMutation object

charliepark:
This PR migrates a couple of files over from using the raw object returned by `useApiMutation` to using the `mutateAsync` function from within that object. Partially addresses #2367, though there will be more to migrate in time.


david-crespo (approved):
noice

david-crespo (approved):
still good!

PR #2423, opened 2024-09-09
Destructure useApiMutation for createInstance

charliepark:
A small one, in the vein of #2420, working on the files that need to get migrated as a part of #2367.

The line that motivated this change was the `[createInstance.error]` passed in to the `useEffect`, but since we were destructuring the `useApiMutation` response anyway, the others are coming along for the ride.

### DIFF
--- a/app/forms/instance-create.tsx
+++ b/app/forms/instance-create.tsx
@@ -173,7 +173,11 @@ export function CreateInstanceForm() {
   const { project } = useProjectSelector()
   const navigate = useNavigate()
 
-  const createInstance = useApiMutation('instanceCreate', {
+  const {
+    mutateAsync: createInstance,
+    isPending,
+    error,
+  } = useApiMutation('instanceCreate', {
     onSuccess(instance) {
       // refetch list of instances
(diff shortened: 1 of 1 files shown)


app/forms/instance-create.tsx
```diff
@@ -244,10 +248,10 @@ export function CreateInstanceForm() {
   const imageSizeGiB = image?.size ? Math.ceil(image.size / GiB) : undefined
 
   useEffect(() => {
-    if (createInstance.error) {
+    if (error) {
       setIsSubmitting(false)
     }
-  }, [createInstance.error])
```
david-crespo, line 250: This one was not actually a #2367 problem because the dep array is looking at the same `error` value in both the original and the updated version. So in this case I think the vagueness of the name `error` might weigh against this change. I'd say if you want to keep it anyway, rename `error` to `mutationError`. `isPending` is ok to me because RQ is the only thing that uses that word.

charliepark: Good point; no strong feelings on this one, so will close it.

PR #2424, opened 2024-09-10
Remove unnecessary invalidateQueries command

charliepark:
I went back to look at this a bit more, and couldn't replicate the error message flash that I was seeing earlier, regardless of the presence or absence of this block of code. I tried both while updating the IP Pool's name and when updating the description (without the name changing).

Seeing as it's a bit confusing, and likely unnecessary, I figured we could clean it out.

### DIFF
--- a/app/forms/ip-pool-edit.tsx
+++ b/app/forms/ip-pool-edit.tsx
@@ -41,15 +41,6 @@ export function EditIpPoolSideModalForm() {
       queryClient.invalidateQueries('ipPoolList')
       navigate(pb.ipPool({ pool: updatedPool.name }))
       addToast({ content: 'Your IP pool has been updated' })
-
-      // Only invalidate if we're staying on the same page. If the name
-      // _has_ changed, invalidating ipPoolView causes an error page to flash
-      // while the loader for the target page is running because the current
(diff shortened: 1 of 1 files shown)


david-crespo:
It might be necessary. I wouldn’t be surprised if a test fails because it moves faster than the 2 second `staleTime`. Try increasing `staleTime`, you might be able to repro it then.

david-crespo:
Logically, I do think it’s required — even if it does work without it, that’s probably a coincidence we can’t rely on. We could change something that appears unrelated and cause it to break. I’m going to see if I can get it to break without the invalidation. I thought I saw it at least once when we were working on the other PR.

david-crespo:
Here's why it's necessary.

https://github.com/user-attachments/assets/fdd5f11a-ddec-4c25-9212-7b688aea91f8

charliepark:
That makes sense; do we want to add the description to that page?

david-crespo:
Probably eventually. I think we had some trouble because of the IP address count UI element, it looked weird together with the properties table.

charliepark:
That all makes sense. Thank you for highlighting the issue!

PR #2426, opened 2024-09-10
Get rid of useless RHF `useForm` wrapper

david-crespo:
This is totally mechanical. I added a wrapper for `useForm` a while ago so we could change the validation behavior of all the forms at the same time, but a) it currently "overrides" the setting _to the default_ (validate on submit initially and then on change after), and b) we weren't even using it everywhere —  we were often importing directly from RHF anyway because it's the top LSP auto-import suggestion. I don't see us wanting at any point to opt in to validate-on-change or whatever globally.


charliepark (approved):
👍

Issue #2388, opened 2024-08-19
Add combobox for routes


charliepark:
Once #2296 is merged, improve the UI for VPC routes, giving instances and subnets a selectable interface, similar to #2065 

<img width="512" alt="Screenshot 2024-08-27 at 2 35 00 PM" src="https://github.com/user-attachments/assets/3b6bf1d8-9dc6-406f-9cc2-4d29cd4e7f49">

PR #2436, opened 2024-09-12
Extract common fields and refactor VPC router route forms

charliepark:
There was a lot of duplicated code in the VPC router route create/edit forms. I started on #2388, but realized it'd be good to clean the forms up a little first, so this is the result of that refactoring. Open to other refinements, if anything stands out.

### DIFF
--- a/app/forms/vpc-router-route-common.tsx
+++ b/app/forms/vpc-router-route-common.tsx
@@ -8,7 +8,12 @@
 
 import type { UseFormReturn } from 'react-hook-form'
 
-import type { RouterRouteCreate, RouterRouteUpdate } from '~/api'
+import type {
+  RouteDestination,
+  RouterRouteCreate,
+  RouterRouteUpdate,
+  RouteTarget,
+} from '~/api'
 import { DescriptionField } from '~/components/form/fields/DescriptionField'
 import { ListboxField } from '~/components/form/fields/ListboxField'
 import { NameField } from '~/components/form/fields/NameField'
@@ -30,6 +35,26 @@ export const routeFormMessage = {
   noDeletingSystemRouters: 'System routers cannot be deleted',
 }
 
+// VPCs can not be specified as a destination in custom routers
+// https://github.com/oxidecomputer/omicron/blob/4f27433d1bca57eb02073a4ea1cd14557f70b8c7/nexus/src/app/vpc_router.rs#L363
+const destTypes: Record<Exclude<RouteDestination['type'], 'vpc'>, string> = {
+  ip: 'IP',
+  ip_net: 'IP network',
+  subnet: 'Subnet',
+}
+
+// Subnets and VPCs cannot be used as a target in custom routers
+// https://github.com/oxidecomputer/omicron/blob/4f27433d1bca57eb02073a4ea1cd14557f70b8c7/nexus/src/app/vpc_router.rs#L362-L368
+const targetTypes: Record<Exclude<RouteTarget['type'], 'subnet' | 'vpc'>, string> = {
+  ip: 'IP',
+  instance: 'Instance',
+  internet_gateway: 'Internet gateway',
+  drop: 'Drop',
+}
+
+const toItems = (mapping: Record<string, string>) =>
+  Object.entries(mapping).map(([value, label]) => ({ value, label }))
+
 type RouteFormFieldsProps = {
   form: UseFormReturn<RouteFormValues>
   isDisabled?: boolean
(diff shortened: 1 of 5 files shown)


charliepark:
Looking into an issue with editing existing routes.

charliepark:
This uses a useEffect where we had been using an onChange on the form field. We could define the onChange in the `create` and `edit` callsites and pass it through to the common fields, though it seemed cleaner to pass fewer props.

app/forms/vpc-router-route-common.tsx
```diff
@@ -74,3 +88,34 @@ export const targetValueDescription = (targetType: RouteTarget['type']) =>
...
+  targetType,
+  isDisabled,
+}: RouteFormFieldsProps) => (
+  <>
+    {isDisabled && (
+      <Message variant="info" content={routeFormMessage.vpcSubnetNotModifiable} />
+    )}
+    <NameField name="name" control={control} disabled={isDisabled} />
+    <DescriptionField name="description" control={control} disabled={isDisabled} />
+    <ListboxField {...fields.destType} control={control} disabled={isDisabled} />
```
david-crespo, line 108: If `fields` is no longer shared, you can get rid of it and inline these as regular props.

charliepark:
> This uses a useEffect where we had been using an onChange on the form field. We could define the onChange in the `create` and `edit` callsites and pass it through to the common fields, though it seemed cleaner to pass fewer props.

Update: By passing the entire `form` in to the common fields, we can use `form.setValue` directly in the common component within the `onChange`, and so don't need to use a `useEffect`, and don't need to declare anything (`useEffect` or `onChange`) in the callsites.

david-crespo:
Nice, passing form is a good trick. You could conceivably go all the way and share the entire `<SideModalForm>`, and just pass in the props that differ, but that would require a lot more props, so it could be harder to understand.

app/forms/vpc-router-route-common.tsx
```diff
@@ -0,0 +1,105 @@
...
+        name="destination.type"
+        label="Destination type"
+        control={control}
+        items={[
+          // VPCs can not be specified as a destination in custom routers
+          // https://github.com/oxidecomputer/omicron/blob/4f27433d1bca57eb02073a4ea1cd14557f70b8c7/nexus/src/app/vpc_router.rs#L363
+          { value: 'ip', label: 'IP' },
+          { value: 'ip_net', label: 'IP network' },
+          { value: 'subnet', label: 'Subnet' },
+        ]}
```
david-crespo, line 57: Inlining these is clearly more readable, but an advantage of the old way was that we would get a type error if the list of destination or target types changed in the generated API.

```ts
// VPCs can not be specified as a destination in custom routers
// https://github.com/oxidecomputer/omicron/blob/4f27433d1bca57eb02073a4ea1cd14557f70b8c7/nexus/src/app/vpc_router.rs#L363
const destTypes: Record<Exclude<RouteDestination['type'], 'vpc'>, string> = {
  ip: 'IP',
  ip_net: 'IP network',
  subnet: 'Subnet',
}

// Subnets and VPCs cannot be used as a target in custom routers
// https://github.com/oxidecomputer/omicron/blob/4f27433d1bca57eb02073a4ea1cd14557f70b8c7/nexus/src/app/vpc_router.rs#L362-L368
const targetTypes: Record<Exclude<RouteTarget['type'], 'subnet' | 'vpc'>, string> = {
  ip: 'IP',
  instance: 'Instance',
  internet_gateway: 'Internet gateway',
  drop: 'Drop',
}

const toItems = (mapping: Record<string, string>) =>
  Object.entries(mapping).map(([value, label]) => ({ value, label }))
```

charliepark: Yeah, was weighing that. Will revert to get the types back in.

app/forms/vpc-router-route-common.tsx
```diff
@@ -0,0 +1,117 @@
...
+      />
+      <ListboxField
+        name="target.type"
+        label="Target type"
+        control={control}
+        items={toItems(targetTypes)}
+        placeholder="Select a target type"
+        required
+        onChange={(value) => {
+          form.setValue('target.value', value === 'internet_gateway' ? 'outbound' : '')
```
david-crespo, line 97: I see this is a change from the previous logic. Good one, it makes way more sense to always clear it. There is one minor issue but it's tolerable: "changing" the type to the same value it already is fires `onChange` and clears the value (I tried it). The brute force fix would be to check if it's different from the current value before firing `setValue`. I'm also fine without that. I checked the headless listbox docs and it doesn't seem like there's an obvious way to tell it not to fire `onChange` unless the value has actually changed.

david-crespo (approved):
Nice!

app/forms/vpc-router-route-common.tsx
charliepark: I experimented with a few iterations with the brute force "pass a conditional prop through with the frozen originalTargetType and compare it to the form's targetType at the onChange firing time", but things got fairly squirrelly. For now, I'll get this folded in, as I think clearing the value is better as the default action. Absolutely good to revisit it down the road if we find UX friction on the field being problematic.

PR #2437, opened 2024-09-12
Add Transit IPs column to instance NIC table

### LINKED ISSUE #2381: Instance networking tab to support add/remove transit IPs
The use of transit IPs is a special use case of custom routes, illustrated in [Example 4](https://docs.oxide.computer/guides/configuring-guest-networking#_example_4_software_routing_tunnels) of the Networking guide. The optional array attribute `transit_ips` defines which target IPs (of other instances in the same subnet) will have their packets re-routed to the instance linked to these IPs.

charliepark:
This adds a column for Transit IPs to the Instance NIC table …
Screenshot 2024-09-12 at 1 07 11 PM

… and an update to the NIC edit form:
Screenshot 2024-09-17 at 8 49 56 AM

It is worth noting that it is currently the case with the API that `transitIps` can be added to an existing `InstanceNetworkingInterface` — https://github.com/oxidecomputer/console/blob/main/app/api/__generated__/Api.ts#L1902-L1918 — but not passed in when creating a new one — https://github.com/oxidecomputer/console/blob/main/app/api/__generated__/Api.ts#L1801-L1813. We might change this in a future PR once we get clarity about the intent of that feature, but I want to keep this PR scoped to the existing API.

Closes #2381


app/components/ListPlusCell.tsx
```diff
@@ -30,7 +30,7 @@ export const ListPlusCell = ({
   const rest = array.slice(numInCell)
   const content = (
     <div>
-      <div className="mb-2">{tooltipTitle}</div>
+      {tooltipTitle && <div className="mb-2">{tooltipTitle}</div>}
```
david-crespo, line 33: Might be nice with `Other transit IPs` or something as the title? Maybe not though.

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -133,6 +134,21 @@ const staticCols = [
...
+    cell: (info) => {
+      const transitIps = info.getValue()
+      return transitIps?.length ? (
+        <ListPlusCell numInCell={1}>
+          {transitIps.map((ip) => (
+            <div key={ip}>{ip}</div>
+          ))}
+        </ListPlusCell>
+      ) : (
+        <EmptyCell />
```
david-crespo, line 148: This is a good idea — consider adding it to `ListPlusCell` if the existing call sites would benefit

charliepark:
Current state of this UI:

Screenshot 2024-09-13 at 4 15 07 PM

app/forms/network-interface-edit.tsx
```diff
@@ -65,6 +112,76 @@ export function EditNetworkInterfaceForm({
...
+      <FormDivider />
+
+      <div className="flex flex-col gap-3">
+        {/* We have to blow this up instead of using TextField to get better text styling on the label */}
+        <div className="mt-2">
+          <label id="transitIps-label" htmlFor="transitIp" className="text-sans-lg">
+            Transit IPs
+          </label>
+          <TextInputHint id="transitIps-help-text" className="mb-2">
+            Enter an IPv4 or IPv6 address, or{' '}
```
david-crespo, line 124: Love including a docs link here. No need to frame it as a choice between two things.

```suggestion
            Enter an IPv4 or IPv6 address. {' '}
```

Not sure either about the icon. I’m surprised we haven’t run into this question before. I lean slightly against, I think.

app/forms/network-interface-edit.tsx
```diff
@@ -65,6 +112,76 @@ export function EditNetworkInterfaceForm({
...
+            <a
+              href={links.transitIpsDocs}
+              className="inline-flex gap-1 children:last:border-0"
+              target="_blank"
+              rel="noreferrer"
+            >
+              learn more about Transit IPs
```
david-crespo, line 131: If the whole sentence is the link, I would lowercase transit. Capital L is to go with the other suggestion.

```suggestion
              Learn more about transit IPs
```

charliepark:
I passed the Vercel preview link to Angela for review, and she said the new column and updated form look good.

app/components/ListPlusCell.tsx
charliepark: I was ambivalent on this, but realized that I was looking at it on a larger screen:
Screenshot 2024-09-17 at 8 43 55 AM

When looking at it on a smaller screen, it's possible for the tooltip to cover up the column header:
Screenshot 2024-09-17 at 8 44 25 AM

So I think having the header in makes it a bit clearer:
Screenshot 2024-09-17 at 8 43 18 AM

Will add that in.

app/components/ListPlusCell.tsx
```diff
@@ -18,14 +19,20 @@ type ListPlusCellProps = {
...
   children,
   numInCell = 1,
 }: ListPlusCellProps) => {
   const array = React.Children.toArray(children)
+  if (array.length === 0) {
+    return <EmptyCell />
+  }
   const inCell = array.slice(0, numInCell)
   const rest = array.slice(numInCell)
   const content = (
```
david-crespo, line 38: I think we should add a class to the `tooltipTitle` div to style the text. We never noticed this before because the items in the list were all badges, so they were distinct from the title text already. Now that they're text, I think the title should be white or bold or both.

image

charliepark: Updated to `text-sans-semi-md text-default`

Screenshot 2024-09-17 at 9 46 46 AM

charliepark:
For posterity, here's the current state of the UI:

Screenshot 2024-09-17 at 9 48 25 AM

david-crespo:
Hm, what do you think about collapsing it under an `Advanced` thing like we do on some other forms? The argument against would be that the main reason anyone would want to use this form ever is to make a NIC with some transit IPs.

charliepark:
I'm inclined to not tuck it into a collapsible `Advanced` section. The form is such a simple one (name, description), I don't think including the Transit IPs section overwhelms the user. I think it makes sense on something like Instance Create because the form has so much going on.

app/components/form/ClearAndAddButtons.tsx
```diff
@@ -0,0 +1,36 @@
...
+}: ClearAndAddButtonsProps) => (
+  <div className="flex justify-end gap-2.5">
+    <Button variant="ghost" size="sm" disabled={!isDirty} onClick={onClear}>
+      Clear
+    </Button>
+    <Button size="sm" onClick={onSubmit}>
+      {addButtonCopy}
+    </Button>
+  </div>
+)
```
david-crespo, line 36: No big deal either way, but if this is always used with `MiniTable` it might as well live with `MiniTable`.

app/components/form/ClearAndAddButtons.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright Oxide Computer Company
+ */
+
+import { Button } from '~/ui/lib/Button'
+
+type ClearAndAddButtonsProps = {
+  addButtonCopy: string
+  onClear: () => void
+  onSubmit: () => void
+  isDirty: boolean
```
david-crespo, line 15: It was already like this, but I'd rather call this `disableClear` because this component doesn't know or need to know anything about the semantics of the form it is called from.

app/forms/network-interface-edit.tsx
```diff
@@ -65,6 +84,69 @@ export function EditNetworkInterfaceForm({
     >
       <NameField name="name" control={form.control} />
       <DescriptionField name="description" control={form.control} />
+      <FormDivider />
+
+      <div className="flex flex-col gap-3">
+        {/* We have to blow this up instead of using TextField for better layout control of field and ClearAndAddButtons */}
+        <div>
+          <FieldLabel id="transitIps-label" htmlFor="transitIp" optional>
```
david-crespo, line 92: don't think the ID is needed since `htmlFor` is there

app/forms/network-interface-edit.tsx
```diff
@@ -65,6 +84,69 @@ export function EditNetworkInterfaceForm({
...
+            Transit IPs
+          </FieldLabel>
+          <TextInputHint id="transitIps-help-text" className="mb-2">
```
david-crespo, line 95: Is this ID unused? Should it be referenced in a `aria-describedby` or whatever? Looks like this magic only happens `tooltipText` is truthy, which seems kinda weird tbh

https://github.com/oxidecomputer/console/blob/7daaa337c1895b68ea414e7d059656da2b4fe3d1/app/components/form/fields/TextField.tsx#L132

app/forms/network-interface-edit.tsx
charliepark: Agree, though `id` is currently a required prop for `FieldLabel`, and I'm not sure about side effects of making it optional. Can investigate it, though it seems minor to keep it in for now.

charliepark: Possibly related to the `aria-labelledby` magic in your comment on line 95, though not sure yet.

david-crespo:
I think we might have to do some client-side validation of the IPs to get a decent experience around malformed transit IPs. The alternative would be to catch the server's validation error and turn it into something more usable, but I think client-side validation preventing the whole error in the first place would be both easier and more maintainable.

https://github.com/user-attachments/assets/dcdbae97-d6ab-43c4-8298-efc534465582


We do have a `validateIp` function that uses rather elaborate regexes, but I don't think they handle CIDR blocks, so I guess we'll want to add that somehow.

https://github.com/oxidecomputer/console/blob/1625d02a540035e6b10e6222e7ef709691dda7f1/app/util/str.ts#L57-L67

We're already using the impressive and robust-seeming [`ip-num`](https://ip-num.github.io/ip-num/) library in the mock server

https://github.com/oxidecomputer/console/blob/1625d02a540035e6b10e6222e7ef709691dda7f1/mock-api/msw/util.ts#L11

...but we're not yet using it anywhere in the app code. I wonder if we could replace our IPv4 and IPv6 regexes with calls to this library, as well as using it to validate IP networks.

https://github.com/oxidecomputer/console/blob/1625d02a540035e6b10e6222e7ef709691dda7f1/app/forms/ip-pool-range-add.tsx#L57-L58

app/forms/network-interface-edit.tsx
charliepark: Fixed in [aee425a](https://github.com/oxidecomputer/console/pull/2437/commits/aee425abd25b745d87006371c84ba1a07fb112c1)

app/components/form/fields/TextField.tsx
```diff
@@ -129,12 +128,17 @@ export const TextFieldInner = <
...
         error={!!error}
-        aria-labelledby={cn(`${id}-label`, !!tooltipText && `${id}-help-text`)}
-        aria-describedby={tooltipText ? `${id}-label-tip` : undefined}
+        aria-labelledby={`${id}-label ${id}-help-text`}
+        aria-describedby={tooltipText ? `${id}-tooltipText` : undefined}
         onChange={(e) => onChange(transform ? transform(e.target.value) : e.target.value)}
         {...fieldRest}
         {...props}
       />
+      {tooltipText && (
```
charliepark, line 137: I want to revisit this one. It might not work with the deconstructed form, but it could make sense to pass the `id` to the tooltip itself, rather than duplicating the tooltipText in the DOM.

david-crespo:
As discussed in chat, I'm working on some IP network validation utilities. I think we should merge this without validation and add it as a followup.

app/components/form/fields/TextField.tsx
charliepark: I have revisited, and chatted with Crespo. It would require a fair bit of prop drilling to get the `id` passed to the appropriate spot, and the `tooltipText` happens infrequently enough that we're okay with an occasional bit of duplication.

david-crespo (approved):
Nice! We will follow up with IP address and network validation.

PR #2440, opened 2024-09-12
minor: tweak image upload cancel e2e for less flake

david-crespo:
Followup to #2415. Flakes seem less frequent but we still [saw one](https://github.com/oxidecomputer/console/actions/runs/10820606878/job/30021009761).

PR #2441, opened 2024-09-12
chore: patch-level runtime dep bumps

david-crespo:
Bumping a few that are very unlikely to cause issues.


david-crespo:
Come on, feels like #2440 actually made things _worse_.

PR #2446, opened 2024-09-16
chore: Remake `package-lock.json` from scratch

david-crespo:
I accidentally discovered on Friday that if I delete `package-lock.json` and run `npm i` to remake it from scratch, this cuts the file down from 33k lines to 20k. This may be due primarily to lockfile version 3 removing some duplication. I seriously doubt there are any downsides to this but I want to think about it a little more. It would be kind of cool to diff the tree of resolved deps before and after, but I'm not too worried about it.


package.json
```diff
@@ -63,6 +64,7 @@
     "react-hook-form": "^7.53.0",
     "react-is": "^18.3.1",
     "react-merge-refs": "^2.1.1",
+    "react-remove-scroll": "2.5.5",
```
david-crespo, line 67: Had to add these because we're patching them, and regenerating the lockfile would otherwise upgrade them.

david-crespo:
Everything looks good except a couple of type errors due to a change in `react-stately`'s calendar state types. I've been putting off upgrading that library because I didn't feel like figuring it out, but I guess I'll figure it out. The alternative is pinning the current version of react-stately, which seems silly.

PR #2449, opened 2024-09-17
Add makeHeadCell helper function

charliepark:
A rather small improvement, and I'm fine if we decide not to move forward with it.

Essentially, this reduces some of the duplication in `MiniTable`s by pulling the header cells into a helper function. We pass in the relevant strings and they get turned into the appropriate `HeadCell`s, with an optional CSS `className` string.

Screenshot 2024-09-16 at 5 34 27 PM

Screenshot 2024-09-16 at 5 34 40 PM


charliepark:
Eh, after thinking about it, I don't think the savings here really justify the added function. Am glad I poked around with it, but I think we're better off with the existing approach. Will close this one down.

PR #2450, opened 2024-09-17
Scroll entire page instead of content pane

david-crespo:
Trying to extract the simple thing out of #2087. It mostly works, but there are still some issues:

- [ ] In theory this change should let us switch back to the [`<ScrollRestoration />`](https://reactrouter.com/en/main/components/scroll-restoration) thing built into RR, but it's not working yet
- [x] Figuring out little things like full-page layout shift due to modals removing the scroll bar

<details>
  <summary>video of layout shift</summary>
  
  https://github.com/user-attachments/assets/0e5eb7b7-5821-4e9a-84e8-9e28979f706e
</details>


david-crespo:
While messing around with converting dropdowns to use Headless instead of Radix (which we want to do anyway), I figured out you can fix the layout shift by making all dropdowns `modal={false}`. Still need to fix the z-index on the dropdown or the topbar, as shown.

https://github.com/user-attachments/assets/851594a1-8afd-4233-8596-425ca832ef11

david-crespo:
Pagination not landing at the bottom is the only remaining issue that I'm aware of. I had to switch back to our custom scroll restoration hook because I couldn't get the RR one to work unless I wrap their [`window.scrollTo` call](https://github.com/remix-run/react-router/blob/9c32cfa3529071373ff306d7711c34f7052dfc78/packages/react-router-dom/index.tsx#L1881) in a `setTimeout`, which means it's running too early (theirs runs in a layout effect and ours runs in a regular effect).


image

benjaminleonard:
Are we running into this because the query has not come back yet, the table not populated and therefore the page is not long enough to scroll to when the scroll restoration has run?

david-crespo:
I think I assumed that was the case, and I probably did check the scrollable height, but I don’t remember whether I printed the entire DOM at the time `scrollTo` is called to confirm it.

david-crespo:
Will redo later

PR #2451, opened 2024-09-17
Use destructured mutate method for dependency array

charliepark:
I believe this is the last migration needed to close out #2367. I scanned through looking at all instances of `= useApiMutation` to see if their variables were A) not destructured and B) used in a dependency array, and this was the only one that I saw that still needed to be switched over.

### DIFF
--- a/app/pages/project/disks/DisksPage.tsx
+++ b/app/pages/project/disks/DisksPage.tsx
@@ -104,7 +104,7 @@ export function DisksPage() {
     },
   })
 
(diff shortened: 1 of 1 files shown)


charliepark:
The relevant dependency array: https://github.com/oxidecomputer/console/pull/2451/files#diff-991cae7316222eb3ef6394499cb3166acaa3a0814d516f42cb9d5c1144967424R157

david-crespo (approved):

PR #2452, opened 2024-09-17
Radix dropdown -> Headless dropdown

david-crespo:
Radix has given us a bit of trouble here and there and I'd prefer to be all-in on Headless. While working on #2450, decided to poke at converting our dropdown to use Headless. The goal is to keep the API of our wrapper component roughly the same to minimize the diff. Plus we cut the bundle size a tiny bit (384 kB to 376 kB gzipped). Can't drop `@radix-ui/react-focus-guards` yet because it's still used by the dialog lib.

The four spots the dropdown menu is used are:

* `TopBarPicker` (project/instance/vpc picker in header bar)
* User menu in top right corner
* Actions menu on table rows
* More actions menu in top right on detail pages, e.g., instance actions


charliepark (approved):
This looks cleaner, and, in clicking through the vercel deployment, all seems good.

david-crespo:
Still need to fix this. Radix locks scroll when the dropdown is open to prevent this from happening. Headless has a `modal` mode that in theory prevents scroll, but it might only affect page-level scrolling. That would be fine in _this_ PR, but I think in #2450 it causes the layout shift mentioned in the description because it causes the scrollbar to disappear. I think for this PR I'll just fix the z-index so it looks normal when scrolling.

https://github.com/user-attachments/assets/3e05974d-0588-439f-a005-38756f66891c

david-crespo:
5c5cdd1150e01a09592740af8a8e0fb8cd14eda3 seems good:

https://github.com/user-attachments/assets/d9a91291-d118-469e-868b-40ee72292049

PR #2453, opened 2024-09-18
Fix instance disk row menu closing on instance state change

david-crespo:
I don't like this very much but we might be stuck with it. The make actions callback depending on the instance state means it gets remade when the instance state changes, which means the whole table gets nuked when the instance goes from stopping to stopped, causing the row actions menu to be closed.

https://github.com/user-attachments/assets/5ecf8f2b-3d8e-4f02-9843-c7c0eb77d1e8

### DIFF
--- a/app/pages/project/instances/instance/tabs/StorageTab.tsx
+++ b/app/pages/project/instances/instance/tabs/StorageTab.tsx
@@ -18,6 +18,7 @@ import {
   useApiQueryClient,
   usePrefetchedApiQuery,
   type Disk,
+  type Instance,
 } from '@oxide/api'
 import { Storage24Icon } from '@oxide/design-system/icons/react'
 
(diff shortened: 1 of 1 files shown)


app/pages/project/instances/instance/tabs/StorageTab.tsx
```diff
@@ -112,7 +120,7 @@ export function StorageTab() {
   const { data: instance } = usePrefetchedApiQuery('instanceView', instancePathQuery)
 
   const makeActions = useCallback(
-    (disk: Disk): MenuAction[] => [
+    ({ disk, instance }: InstanceDisk): MenuAction[] => [
```
david-crespo, line 123: By making the instance part of the row data (same instance for all rows) we can avoid having to remake this callback whenever the instance state changes. This works, but I can't help feeling there has to be a better way.

app/pages/project/instances/instance/tabs/StorageTab.tsx
```diff
@@ -60,7 +61,11 @@ StorageTab.loader = async ({ params }: LoaderFunctionArgs) => {
   return null
 }
 
-const colHelper = createColumnHelper<Disk>()
+type InstanceDisk = Disk & {
+  instanceState: InstanceState
+}
```
david-crespo, line 66: By making the instance state part of the row data, we can avoid having to remake the row actions callback whenever the instance state changes.

charliepark (approved):
I think it's pretty clever! The comment in the code definitely helps clarify what's going on. Definitely feels like an improvement to the UX to not lose the modal when the state changes.

PR #2458, opened 2024-09-19
cleanup: Remove "inline" pagination

david-crespo:
While looking at #2457 I could not find a place where `inline` pagination is actually used. It's also all wonky, with some components defaulting to `inline` and some components defaulting to `page`. Will comment inline because there's some weirdness, but the goal here is to change nothing about any actually existing usages.

### DIFF
--- a/app/components/Pagination.tsx
+++ b/app/components/Pagination.tsx
@@ -14,13 +14,7 @@ import {
 
 const Tunnel = tunnel('pagination')
 
-interface PaginationProps extends UIPaginationProps {
-  /** If true pagination will be rendered wherever `Pagination.Target` is included */
-  inline?: boolean
-}
-export function Pagination({ inline = false, ...props }: PaginationProps) {
-  if (inline) return <UIPagination {...props} />
-
+export function Pagination(props: UIPaginationProps) {
   return (
     <Tunnel.In>
       <UIPagination className="gutter h-14 py-5" {...props} />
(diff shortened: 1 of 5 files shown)


app/components/Pagination.tsx
```diff
@@ -14,13 +14,7 @@ import {
...
 const Tunnel = tunnel('pagination')
 
-interface PaginationProps extends UIPaginationProps {
-  /** If true pagination will be rendered wherever `Pagination.Target` is included */
-  inline?: boolean
-}
-export function Pagination({ inline = false, ...props }: PaginationProps) {
-  if (inline) return <UIPagination {...props} />
-
+export function Pagination(props: UIPaginationProps) {
```
david-crespo, line 17: `inline` was never true here. The only callsite is in `QueryTable`, which defaults to `page` and was never overridden.

app/table/QueryTable.tsx
```diff
@@ -115,7 +112,6 @@ const makeQueryTable = <Item extends Record<string, unknown>>(
       <>
         <Table table={table} rowHeight={rowHeight} />
         <Pagination
-          inline={pagination === 'inline'}
```
david-crespo, line 118: This was always false because `pagination` defaults to `"page"` and we never set the prop explicitly.

app/ui/lib/Pagination.tsx
```diff
@@ -50,7 +48,6 @@ export const Pagination = ({
     <>
       <div
         className={cn(
-          type === 'page' && 'py-5',
```
david-crespo, line 53: This is the weird bit. In all the calling code above this, we were always `page`, never `inline`, but because this defaults to `inline`, we take this bit out.

PR #2461, opened 2024-09-20
Validate IPs and IP networks

david-crespo:
Followup work for https://github.com/oxidecomputer/console/pull/2437#issuecomment-2362290469

- [x] Basic parsing-based validation with `ip-num`
- ~~Confirm Rust code has the same validation results (https://github.com/oxidecomputer/test-ip-validation)~~
- [x] Abandon `ip-num` because it does _not_ have the same validation results
- [x] Decide on return type of parser function
- [x] Figure out how this should be integrated into forms
- [x] Validate transit IP nets
- [x] Validate IPs and IP nets in route form
- [x] Validate IPs and IP nets in firewall rule form
- [x] Polish up weird behavior on host and target subform on type change, submit success, and clear
- [x] Write some e2e tests

### DIFF
--- a/app/components/form/fields/ComboboxField.tsx
+++ b/app/components/form/fields/ComboboxField.tsx
@@ -10,7 +10,9 @@ import {
   useController,
   type Control,
   type FieldPath,
+  type FieldPathValue,
   type FieldValues,
+  type Validate,
 } from 'react-hook-form'
 
 import { Combobox, type ComboboxBaseProps } from '~/ui/lib/Combobox'
@@ -25,6 +27,7 @@ export type ComboboxFieldProps<
   name: TName
   control: Control<TFieldValues>
   onChange?: (value: string | null | undefined) => void
+  validate?: Validate<FieldPathValue<TFieldValues, TName>, TFieldValues>
 } & ComboboxBaseProps
 
 export function ComboboxField<
@@ -51,9 +54,14 @@ export function ComboboxField<
     : allowArbitraryValues
       ? 'Select an option or enter a custom value'
       : 'Select an option',
+  validate,
   ...props
 }: ComboboxFieldProps<TFieldValues, TName>) {
-  const { field, fieldState } = useController({ name, control, rules: { required } })
+  const { field, fieldState } = useController({
+    name,
+    control,
+    rules: { required, validate },
+  })
   return (
     <div className="max-w-lg">
       <Combobox
--- a/app/forms/firewall-rules-common.tsx
+++ b/app/forms/firewall-rules-common.tsx
@@ -6,6 +6,7 @@
  * Copyright Oxide Computer Company
  */
 
+import { useEffect } from 'react'
 import {
   useController,
   useForm,
@@ -63,7 +64,6 @@ type TargetAndHostFilterType =
 type TargetAndHostFormValues = {
   type: TargetAndHostFilterType
   value: string
-  subnetVpc?: string
 }
 
 // these are part of the target and host filter form;
@@ -239,14 +239,14 @@ type CommonFieldsProps = {
   error: ApiError | null
 }
 
+const targetAndHostDefaultValues: TargetAndHostFormValues = {
+  type: 'vpc',
+  value: '',
+}
+
 export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) => {
   const { project, vpc } = useVpcSelector()
-  const targetAndHostDefaultValues: TargetAndHostFormValues = {
-    type: 'vpc',
-    value: '',
-    // only becomes relevant when the type is 'VPC subnet'; ignored otherwise
-    subnetVpc: vpc,
-  }
+
   // prefetchedApiQueries below are prefetched in firewall-rules-create and -edit
   const {
     data: { items: instances },
@@ -280,6 +280,18 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
     targets.onChange([...targets.value, { type, value }])
     targetForm.reset(targetAndHostDefaultValues)
   })
+  // HACK: we need to reset the target form completely after a successful submit,
+  // including especially `isSubmitted`, because that governs whether we're in
+  // the validate regime (which doesn't validate until submit) or the reValidate
+  // regime (which validate on every keypress). The reset inside `handleSubmit`
+  // doesn't do that for us because `handleSubmit` set `isSubmitted: true` after
+  // running the callback. So we have to watch for a successful submit and call
+  // the reset again here.
+  // https://github.com/react-hook-form/react-hook-form/blob/9a497a70a/src/logic/createFormControl.ts#L1194-L1203
+  const { isSubmitSuccessful: targetSubmitSuccessful } = targetForm.formState
+  useEffect(() => {
+    if (targetSubmitSuccessful) targetForm.reset(targetAndHostDefaultValues)
+  }, [targetSubmitSuccessful, targetForm])
 
   // Ports
   const portRangeForm = useForm({ defaultValues: { portRange: '' } })
@@ -312,10 +324,13 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
     if (!type || !value) return
     if (hosts.value.some((t) => t.value === value && t.type === type)) return
     hosts.onChange([...hosts.value, { type, value }])
-    // TODO: something is not resetting right -- if the IP net field is set to
-    // validate on change due to having an error, it stays that way after submit
     hostForm.reset(targetAndHostDefaultValues)
   })
+  // HACK: see comment above about doing the same for target form
+  const { isSubmitSuccessful: hostSubmitSuccessful } = hostForm.formState
+  useEffect(() => {
+    if (hostSubmitSuccessful) hostForm.reset(targetAndHostDefaultValues)
+  }, [hostSubmitSuccessful, hostForm])
 
   return (
     <>
--- a/app/forms/firewall-rules-common.tsx
+++ b/app/forms/firewall-rules-common.tsx
@@ -261,8 +261,7 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
   // Targets
   const targetForm = useForm({ defaultValues: targetAndHostDefaultValues })
   const targets = useController({ name: 'targets', control }).field
-  const targetType = targetForm.watch('type')
-  const targetValue = targetForm.watch('value')
+  const [targetType, targetValue] = targetForm.watch(['type', 'value'])
   // get the list of items that are not already in the list of targets
   const targetItems = {
     vpc: availableItems(targets.value, 'vpc', vpcs),
@@ -308,8 +307,7 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
   // Host Filters
   const hostForm = useForm({ defaultValues: targetAndHostDefaultValues })
   const hosts = useController({ name: 'hosts', control }).field
-  const hostType = hostForm.watch('type')
-  const hostValue = hostForm.watch('value')
+  const [hostType, hostValue] = hostForm.watch(['type', 'value'])
   // get the list of items that are not already in the list of host filters
   const hostFilterItems = {
     vpc: availableItems(hosts.value, 'vpc', vpcs),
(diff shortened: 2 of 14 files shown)


david-crespo:
Ok, here's something interesting. I tried swapping in the `ip-num` things in our existing `validateIp` function and a bunch of tests fail, indicating it disagrees significantly with the Rust parsing code.

```diff
 export function validateIp(ip: string) {
-  const isv4 = IPV4_REGEX.test(ip)
-  const isv6 = !isv4 && IPV6_REGEX.test(ip)
+  let isv4 = false
+  try {
+    IPv4.fromString(ip)
+    isv4 = true
+  } catch (e) {}
+  let isv6 = false
+  if (!isv4) {
+    try {
+      IPv6.fromString(ip)
+      isv6 = true
+    } catch (e) {}
+  }
   return { isv4, isv6, valid: isv4 || isv6 }
 }
```

image

I didn't notice these differences before because my list of examples for IP networks was much less thorough than my kickass list for plain addresses. When I add the offending address to the list of IP nets, `ip-num` considers it valid, as expected.

image

Rust `oxnet`'s `Ipv4Net` considers it invalid, which is expected because it splits the thing on `/` and uses the underlying `Ipv4Addr` to parse the inner IP address ([source](https://github.com/oxidecomputer/oxnet/blob/7dacd265f1bcd0f8b47bd4805250c4f0812da206/src/ipnet.rs#L373-L385)).

image

So, after all that excitement, I am going to take `ip-num` back out and use our existing IP address thing to parse an IP network. Not sure why I didn't start with that.

david-crespo:
truly, we are jamming.

Uploading 2024-09-24-validate-ips-and-nets.mp4…

app/forms/network-interface-edit.tsx
```diff
@@ -57,12 +58,11 @@ export function EditNetworkInterfaceForm({
...
   const transitIpsForm = useForm({ defaultValues: { transitIp: '' } })
 
-  const submitTransitIp = () => {
-    const transitIp = transitIpsForm.getValues('transitIp')
+  const submitTransitIp = transitIpsForm.handleSubmit(({ transitIp }) => {
     if (!transitIp) return
     form.setValue('transitIps', [...transitIps, transitIp])
     transitIpsForm.reset()
-  }
+  })
```
david-crespo, line 65: Without wrapping this in `handleSubmit`, it doesn't run the `validate` function on submit. This way the inner function never gets called unless validation passes.

app/forms/network-interface-edit.tsx
```diff
@@ -107,10 +107,16 @@ export function EditNetworkInterfaceForm({
                 submitTransitIp()
               }
             }}
+            validate={(value) => {
+              const result = validateIpNet(value)
+              if (result.type === 'error') return result.message
+            }}
```
david-crespo, line 113: I'm thinking I might want to split `validateIpNet` into `parseIpNet` and `validateIpNet`, and the latter just returns a string or undefined so it could go here simply as `validate={validateIpNet}`

app/forms/firewall-rules-common.tsx
```diff
@@ -272,7 +278,7 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
     if (!type || !value) return
     if (targets.value.some((t) => t.value === value && t.type === type)) return
     targets.onChange([...targets.value, { type, value }])
-    targetForm.reset()
+    targetForm.reset(targetAndHostDefaultValues)
```
david-crespo, line 281: I wasn't seeing this reset do what I expected, and I figured out why: we call this reset inside the callback to `handleSubmit`. This does in fact set `isSubmitted` to false. But after it runs our callback, [`handleSubmit` sets `isSubmitted: true`](https://github.com/react-hook-form/react-hook-form/blob/9a497a70accbc5e0be5f4e1c06c0ff00baa963da/src/logic/createFormControl.ts#L1194-L1203). Hence the very brief flash of `false`, then back to `true`.

app/forms/firewall-rules-common.tsx
```diff
@@ -518,7 +546,12 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
           control={hostForm.control}
           valueType={hostType}
           items={hostFilterItems[hostType]}
-          onTypeChange={() => hostForm.setValue('value', '')}
+          // HACK: reset the whole subform, keeping type (because we just set
+          // it). most importantly, this resets isSubmitted so the form can go
+          // back to validating on submit instead of change
+          onTypeChange={() =>
+            hostForm.reset({ type: hostForm.getValues('type'), value: '' })
+          }
```
david-crespo, line 554: I kind of hate this, but it does pretty elegantly express what I want to happen here, namely that we are basically starting over from scratch whenever we change the type. Very weird combination of feeling both wrong and exactly right.

david-crespo:
Still need to add another test or two and fix the clear button bug in the firewall rules inputs, but this is mostly ready to review. It's crazy long but most of the length is tests and fresh additions.

app/components/form/fields/ComboboxField.tsx
```diff
@@ -25,6 +27,7 @@ export type ComboboxFieldProps<
   name: TName
   control: Control<TFieldValues>
   onChange?: (value: string | null | undefined) => void
+  validate?: Validate<FieldPathValue<TFieldValues, TName>, TFieldValues>
```
david-crespo, line 30: Adding this lets us validate names in arbitrary-value comboboxes.

app/util/ip.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  '1:2:3:4:5:6:7:88888',
+  '2001:db8:3:4:5::192.0.2.33', // std::new::Ipv6Net allows this one
+  'fe08::7:8%',
+  'fe08::7:8i',
+  'fe08::7:8interface',
+]
+
+test.each(invalid)('parseIp catches invalid IP: %s', (s) => {
+  expect(parseIp(s)).toStrictEqual({ type: 'error', message: 'Not a valid IP address' })
+})
```
david-crespo, line 80: From here up, the tests are basically unchanged, I just had to extract the lists of IPs so I could reuse them in the IP net tests.

app/util/ip.ts
```diff
@@ -0,0 +1,86 @@
...
+
+// Borrowed from Valibot. I tried some from Zod and an O'Reilly regex cookbook
+// but they didn't match results with std::net on simple test cases
+// https://github.com/fabian-hiller/valibot/blob/2554aea5/library/src/regex.ts#L43-L54
+
+const IPV4_REGEX =
+  /^(?:(?:[1-9]|1\d|2[0-4])?\d|25[0-5])(?:\.(?:(?:[1-9]|1\d|2[0-4])?\d|25[0-5])){3}$/u
+
+const IPV6_REGEX =
+  /^(?:(?:[\da-f]{1,4}:){7}[\da-f]{1,4}|(?:[\da-f]{1,4}:){1,7}:|(?:[\da-f]{1,4}:){1,6}:[\da-f]{1,4}|(?:[\da-f]{1,4}:){1,5}(?::[\da-f]{1,4}){1,2}|(?:[\da-f]{1,4}:){1,4}(?::[\da-f]{1,4}){1,3}|(?:[\da-f]{1,4}:){1,3}(?::[\da-f]{1,4}){1,4}|(?:[\da-f]{1,4}:){1,2}(?::[\da-f]{1,4}){1,5}|[\da-f]{1,4}:(?::[\da-f]{1,4}){1,6}|:(?:(?::[\da-f]{1,4}){1,7}|:)|fe80:(?::[\da-f]{0,4}){0,4}%[\da-z]+|::(?:f{4}(?::0{1,4})?:)?(?:(?:25[0-5]|(?:2[0-4]|1?\d)?\d)\.){3}(?:25[0-5]|(?:2[0-4]|1?\d)?\d)|(?:[\da-f]{1,4}:){1,4}:(?:(?:25[0-5]|(?:2[0-4]|1?\d)?\d)\.){3}(?:25[0-5]|(?:2[0-4]|1?\d)?\d))$/iu
```
david-crespo, line 17: regexes are unchanged, the point was to distinguish between parse and validate and to add the IP net things

app/forms/firewall-rules-common.tsx
```diff
@@ -132,8 +133,10 @@ const DynamicTypeAndValueFields = ({
           items={items}
           allowArbitraryValues
           hideOptionalTag
-          // TODO: validate here, but it's complicated because it's conditional
-          // on which type is selected
+          validate={(value) =>
+            // TODO: is required false correct here? should this function even have that argument?
+            validateName(value, `${capitalize(sectionType)} name`, false)
+          }
```
david-crespo, line 139: I think `required: false` is correct because we don't want this to show a required error when it's left empty after a bad input was attempted.

app/forms/firewall-rules-common.tsx
david-crespo: Fixed by `useEffect` below.

app/forms/firewall-rules-common.tsx
```diff
@@ -518,13 +544,18 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
...
+          // back to validating on submit instead of change
+          onTypeChange={() =>
+            hostForm.reset({ type: hostForm.getValues('type'), value: '' })
+          }
           onInputChange={(value) => hostForm.setValue('value', value)}
           onSubmitTextField={submitHost}
         />
         <MiniTable.ClearAndAddButtons
           addButtonCopy="Add host filter"
-          disableClear={!!hostValue}
```
david-crespo, line 527: WHOOPS

app/forms/ip-pool-range-add.tsx
```diff
@@ -35,20 +33,20 @@ const ipv6Error = { type: 'pattern', message: 'IPv6 ranges are not yet supported
...
+  } else if (first.type === 'v6') {
     errors.first = ipv6Error
   }
 
-  if (!last.valid) {
-    errors.last = invalidAddressError
-  } else if (last.isv6) {
+  if (last.type === 'error') {
+    errors.last = { type: 'pattern', message: last.message }
+  } else if (last.type === 'v6') {
```
david-crespo, line 49: Nothing actually changes in this file, so I didn't change the tests.

app/forms/network-interface-edit.tsx
```diff
@@ -107,12 +114,19 @@ export function EditNetworkInterfaceForm({
                 submitTransitIp()
               }
             }}
+            validate={(value) => {
+              const error = validateIpNet(value)
+              if (error) return error
+
+              if (transitIps.includes(value)) return 'Transit IP already in list'
+            }}
```
david-crespo, line 122: Doing this here instead of in the `handleSubmit` means we get a beautiful validation error inline on attempts to add a dupe.

david-crespo (commented):
Added some explanations.

app/forms/firewall-rules-common.tsx
```diff
@@ -255,8 +264,7 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
   // Targets
   const targetForm = useForm({ defaultValues: targetAndHostDefaultValues })
   const targets = useController({ name: 'targets', control }).field
-  const targetType = targetForm.watch('type')
-  const targetValue = targetForm.watch('value')
+  const [targetType, targetValue] = targetForm.watch(['type', 'value'])
```
charliepark, line 267: 💡

app/forms/firewall-rules-common.tsx
```diff
@@ -272,8 +280,20 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
...
+  // the validate regime (which doesn't validate until submit) or the reValidate
+  // regime (which validate on every keypress). The reset inside `handleSubmit`
+  // doesn't do that for us because `handleSubmit` set `isSubmitted: true` after
+  // running the callback. So we have to watch for a successful submit and call
+  // the reset again here.
+  // https://github.com/react-hook-form/react-hook-form/blob/9a497a70a/src/logic/createFormControl.ts#L1194-L1203
+  const { isSubmitSuccessful: targetSubmitSuccessful } = targetForm.formState
+  useEffect(() => {
+    if (targetSubmitSuccessful) targetForm.reset(targetAndHostDefaultValues)
+  }, [targetSubmitSuccessful, targetForm])
```
charliepark, line 296: That's a clever workaround. A shame that we need to distinguish between the two flows, but premature validation errors are even worse. Nevertheless, this is clever.

charliepark (approved):
Looks great!

PR #2464, opened 2024-09-23
Boot order

### LINKED ISSUE #2335: Boot order UI
API PR: https://github.com/oxidecomputer/omicron/pull/6585

This is looking like a `boot_disk: Option<NameOrId>` on the instance create body and a `boot_disk_id: Uuid` on the instance view response. Then (possibly in a followup PR on top of https://github.com/oxidecomputer/omicron/pull/6321) there will also be a field on an instance update endpoint allowing you to change it. You can always remove the configured boot disk ID (I think), which will result in falling back to whatever the (likely undesirable) behavior is now. There will likely be an added constraint on disk detach, namely that you cannot detach a the boot disk — you have to change or clear the specified boot disk first.

So, the work on our end sounds like:

## Instance create POST body

We already have a concept of which disk is the boot disk; currently we put it first in the list of disks. All we have to do is add `bootDisk: bootDisk.name`.

https://github.com/oxidecomputer/console/blob/e3d424599f5c4c1fc493ede4f4d6c79b478f857f/app/forms/instance-create.tsx#L305-L326

## Instance disks table

* Show which disk is the boot disk
  * Sort it first in the list
  * Probably with a little badge like we do for primary NIC
  image

* Add `Set as boot disk` action to disk row actions
* Disable disk detach button for boot disk (with helpful tooltip)
  * This is the only place we do disk detach; you can't do it, for example, on the main project disks list

image

## Open questions

* When an instance has no boot disk, should we show that somehow?
  * Put an info box somewhere saying "this instance doesn't have an explicit boot disk, so the behavior might be confusing"?
  * Only do that when there is more than one disk?

david-crespo:
Preview: https://console-git-boot-order-oxidecomputer.vercel.app/projects/mock-project/instances/db1/storage

Closes #2335. Depends on https://github.com/oxidecomputer/omicron/pull/6585.

https://github.com/user-attachments/assets/3b1087c1-03e6-42ec-89be-43f890173dd8


david-crespo:
### Instance create works

https://github.com/user-attachments/assets/2ae871b8-9c18-47c4-a93c-77d008847f68

benjaminleonard:
I think separate from the badge treatment, there's something here about having a button and a modal to select the boot disk. I had originally made this for NICs. But I think we need some way hoisted up that isn't just in the more actions menu.

image

See also the designs I had for separate boot disk:

image

app/api/util.ts
```diff
@@ -96,6 +96,9 @@ const instanceActions: Record<string, InstanceState[]> = {
   // https://github.com/oxidecomputer/omicron/blob/6dd9802/nexus/db-queries/src/db/datastore/instance.rs#L865
   delete: ['stopped', 'failed'],
 
+  // TODO: whence my licence to make such a claim
+  update: ['stopped'],
```
iximeow, line 100:
```suggestion
  // https://github.com/oxidecomputer/omicron/blob/144e91a/nexus/db-queries/src/db/datastore/instance.rs#L1039-L1043
  update: ['stopped', 'failed'],
```

failed's especially important so you can unset a boot disk that you want to detach failed instance, and i expect this to continue to be valid (even if not useful) when you can update other bits of an instance

mock-api/msw/handlers.ts
```diff
@@ -542,6 +589,32 @@ export const handlers = makeHandlers({
...
+    if (body.boot_disk) {
+      // Only include project if it's a name, otherwise lookup will error.
+      // This will 404 if the disk doesn't exist, which I think is right.
+      const disk = lookup.disk({
+        disk: body.boot_disk,
+        project: isUuid(body.boot_disk) ? undefined : query.project,
+      })
+
+      // disk must of course be attached to this instance already to be set as the boot disk
+      // TODO: update with better link once PR is merged
```
iximeow, line 604:
```suggestion
      // https://github.com/oxidecomputer/omicron/blob/144e91a/nexus/db-queries/src/db/datastore/instance.rs#L1079-L1081
```

app/forms/instance-create.tsx
```diff
@@ -318,6 +318,7 @@ export function CreateInstanceForm() {
               memory: instance.memory * GiB,
               ncpus: instance.ncpus,
               disks: [bootDisk, ...values.disks],
```
iximeow, line 320: i'd suggest making this just `values.disks`, the fact that you can specify a disk multiple times (`boot_disk` + others, repeatedly in `disks`, some combination of both) seems to have [surprised everyone](https://github.com/oxidecomputer/omicron/pull/6585#issuecomment-2386291174). not gonna go reject this right now, but it feels like something we might reject later?

app/pages/project/instances/instance/tabs/StorageTab.tsx
```diff
@@ -22,26 +23,40 @@ import {
...
-const EmptyState = () => (
+const BootDiskEmptyState = () => (
   <TableEmptyBox>
     <EmptyMessage
       icon={<Storage24Icon />}
-      title="No disks"
-      body="Attach a disk to this instance to see it here"
+      title="No boot disk set"
+      // TODO: boot order docs link
+      body="Read about boot order LINK HERE"
```
charliepark, line 49: Just highlighting that this isn't set yet

app/api/util.ts
david-crespo: Thanks. Already had this done but not pushed

app/forms/instance-create.tsx
david-crespo: Will do. Also need to change the mock API logic so it doesn't error when the boot disk is not in the other disks list.

app/pages/project/instances/instance/tabs/StorageTab.tsx
```diff
@@ -120,29 +135,98 @@ export function StorageTab() {
...
+                path: { instance: instance.id },
+                body: { bootDisk: undefined },
+              }),
+            errorTitle: 'Could not unset boot disk',
+            modalTitle: 'Confirm unset boot disk',
+            // TODO: copy + link to docs
+            modalContent: (
+              <p>
+                Are you sure you want to unset <HL>{disk.name}</HL> as the boot disk? This
+                will TODO COPY RE: WHAT HAPPENS HERE
```
charliepark, line 185: What *does* happen here?

app/pages/project/instances/instance/tabs/StorageTab.tsx
```diff
@@ -120,29 +135,98 @@ export function StorageTab() {
 
   const { data: instance } = usePrefetchedApiQuery('instanceView', instancePathQuery)
 
-  const makeActions = useCallback(
+  const { mutateAsync: instanceUpdate } = useApiMutation('instanceUpdate', {
+    onSuccess() {
+      apiQueryClient.invalidateQueries('instanceView')
```
charliepark, line 140: perhaps add an `addToast({ content: 'Your instance has been updated' })` here?

app/pages/project/instances/instance/tabs/StorageTab.tsx
charliepark: Screenshot 2024-10-01 at 2 23 55 PM

We don't really tell them with this copy how to set the boot disk, or that it's done via the table below. I'm having trouble coming up with an elegant way to say that, though. (Copy might also need to handle the "no attached disks at all" case with the copy, so that could be even more ungainly.)

charliepark: One more thought, not really pinned to this code, but pertinent to the screenshot above: The two states are `attached` and `detached`; is it redundant to have the `STATE` column here? There's no way that a disk could show up here in a `detached` state, right? Or might we add some third state in time?

(I also get it if it looks a little empty without that column.)

david-crespo: I've also noticed that. It is redundant but it also looks empty without it. I think it's ok because it reinforces the concept of attachment.

david-crespo:
Can't tell if I'm cooking or this is ridiculous.

https://github.com/user-attachments/assets/1f391c92-c98a-4c8f-a1fa-a4c5552b4d29

iximeow:
ooh, i do like that it's contextually as-helpful-as-we-can-be.

as for what happens without a boot disk: you get the status quo. so the guest will _probably_ try booting disks in PCI device order (which i don't think we really expose in a useful way), but as evidenced by https://github.com/oxidecomputer/omicron/issues/5112 it's quite possible that one of the disks has an EFI System Partition that ends up with the EFI shell tried before the right disk and then everything gets stuck.

i think an informative way to present that would be something like: "With no boot disk, the instance will try booting based on UEFI variables of the first disk with an EFI System Partition". it's really the same behavior with one or multiple disks. but "first" still means "lowest PCI device number", so i think that's still less actionable than ideal?

this makes me wonder if it's possible to end up in a #5112-like state with even a single disk..

david-crespo:
That’s helpful, though indirectly, because to me the main thing it says is there’s no good way to explain this concisely enough in the UI and we should focus on encouraging them to set a boot disk if there’s more than one and leaning on the docs link for explanation.

david-crespo:
Going with this for now for the multiple non-boot disks case.

image

david-crespo:
### Confirm set boot disk when there isn't one currently

Screenshot 2024-10-02 at 12 32 29 PM

### Confirm set boot disk when there is one currently

Screenshot 2024-10-02 at 12 32 58 PM

### Confirm unset boot disk

Screenshot 2024-10-02 at 12 37 30 PM

app/pages/project/instances/instance/tabs/StorageTab.tsx
david-crespo: Good idea, though I think it would be even better if it was specific to the action — this is used for both set and unset boot disk. So it needs to be triggered at each call site. That might require moving this invalidate there because I think if you pass `onSuccess` as an option at call time it will override this one.

PR #2448, opened 2024-09-16
Add dropdowns to router route form for destinations and targets

charliepark:
This PR converts the text fields in the VPC Router Routes form to be — when appropriate — prepopulated dropdowns.

For Destinations, that's when the Destination Type has been selected as `Subnet`; for Targets, `Instance`s.

Screenshot 2024-09-16 at 12 37 14 PM

Closes #2388

PR #2465, opened 2024-09-23
Update isDisabled prop name for Comboboxes

charliepark:
When initially making the Combobox component, I'd used the `isDisabled` prop name to hint at the boolean type it was expecting. I realized, when refactoring code for https://github.com/oxidecomputer/console/pull/2448/files, that it'd be more convenient if it matched other field types, and was simply `disabled`. This PR switches it over.


david-crespo (approved):
Good. I couldn't remember why it had a different name.

PR #2474, opened 2024-09-25
Allow ReactElements to be used in Combobox dropdowns

### LINKED ISSUE #1837: Instance create: fetch more disks and use comboboxes instead of listboxes
In instance create, the drop down selection for `Additional disks` -> `ATTACH EXISTING DISK" menu
only shows me about 100 disks.

I've created some 1400 disks, and I can't choose the disk I want as it is not showing up.
There should be some way to either show all the disks, or enter a few letters and have the set of 
choices filtered to match what I entered, or.. something else ??

DiskInstance

### LINKED ISSUE #2483: Combobox: can't click to place cursor, can't type spaces
I think the button is eating the event that should be going to the input. Headless's [example](https://headlessui.com/react/combobox) doesn't have this problem, so it must be something we're doing.

https://github.com/user-attachments/assets/56e1a199-cac3-4990-8437-639cfc66d1c5

Ok, haha. Turns out we are using `ComboboxInput` inside of `ComboboxButton`, but you're only supposed to need one or the other. The styles will need some untangling, but I was able to confirm immediately that changing `ComboboxButton` to a `div` does in fact fix clicking to place the cursor and typing spaces.

https://github.com/oxidecomputer/console/blob/e46216aae26609c83e32ff512c1d7a2f055782c2/app/ui/lib/Combobox.tsx#L102-L137

charliepark:
For #1837, we had a few more places that needed to get switched from a Listbox to a Combobox. Interestingly, all of our existing Comboboxes had had the same string used for both the `label` and the `value`, so there were a few places where the Combobox was using the `label` where it would more accurately use the `value`. As these were identical before, it didn't cause any problems. With this update, though, we'll begin using elements that have a different `value` from their `label`, so it's more important that we use the `value`.

One example of this is in the image selector, where we now have three elements — the `value`, the `label`, and an optional `selectedLabel` — a custom string for the ComboboxItem.
Screenshot 2024-09-25 at 4 43 53 PM
Here we see the `selectedLabel` in the input field, where the `label` for each item is the richer HTML in the dropdown.

Closes #1837 
Closes #2483


app/ui/lib/Combobox.tsx
```diff
@@ -16,20 +16,42 @@ import {
...
 import { TextInputHint } from './TextInput'
 
+export type ComboboxItem = { value: string; label: ReactNode; selectedLabel: string }
+
+/** Convert an array of items with a `name` attribute to an array of ComboboxItems
+ *  Useful when the rendered label and value are the same; in more complex cases,
+ *  you may want to create a custom ComboboxItem object (see toImageComboboxItem).
+ */
+export const toComboboxItems = (
+  items?: Array<Disk | Instance | Project | VpcSubnet>
```
david-crespo, line 36: `Array<{ name: string }>` will do it

app/ui/lib/Combobox.tsx
```diff
@@ -74,20 +91,24 @@ export const Combobox = ({
...
-    keys: ['value'],
+    keys: ['selectedLabel', 'label'],
     sorter: (items) => items, // preserve original order, don't sort by match
   })
 
   const zIndex = usePopoverZIndex()
 
+  const getDisplayValueForSelected = useMemo(() => {
+    return getSelectedLabelFromValue(items, selected || '')
+  }, [items, selected])
```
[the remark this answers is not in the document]
charliepark: Currently handling this with a prop passed in, `selectedItemLabel`, which is updated via an `onChange` on the Combobox. ([here](https://github.com/oxidecomputer/console/pull/2474/files#diff-69443021cc6d598a05d9a1167c4f2361d629276d08ab0ffc9ada62258b9df9c8R81))

charliepark:
Still futzing with the width of the dropdown, which the docs suggest can be done with `w-[var(--input-width)]`, but something in the nesting must still be off a bit, or we'll need to approach it with a slightly different method. The demo in the docs seems simple enough, so maybe ours is a little overcomplicated. Will keep playing with it.
Screenshot 2024-10-01 at 5 13 56 PM

david-crespo:
image

Brace for the worst thing you've ever seen...........
...
...
...
...
...


```
w-[calc(var(--input-width)+var(--button-width))]
```

david-crespo:
Oh note also it's missing the focus ring. This is from main:

image

app/ui/lib/Combobox.tsx
```diff
@@ -67,79 +85,83 @@ export const Combobox = ({
...
-          className={cn(
-            `flex h-10 w-full grow items-center justify-between rounded border text-sans-md ring-accent-secondary data-[open]:ring-2`,
-            hasError
-              ? 'focus-error border-error-secondary hover:border-error'
-              : 'border-default hover:border-hover',
-            hasError && 'data-[open]:ring-error-secondary',
-            disabled
-              ? 'cursor-not-allowed text-disabled bg-disabled !border-default'
-              : 'bg-default',
-            disabled && hasError && '!border-error-secondary'
```
david-crespo, line 112: I think some of these error classes need to be put back — red border and focus ring are missing.

image

david-crespo:
The behavior

david-crespo:
The behavior where it automatically focuses the image field on error is broken — In theory, `ComboboxField` needs the same fix `ListboxField` got in https://github.com/oxidecomputer/console/pull/2364. However, I am trying it and it's not working yet.

```diff
diff --git a/app/ui/lib/Combobox.tsx b/app/ui/lib/Combobox.tsx
index 946cd236..0dbfa2d9 100644
--- a/app/ui/lib/Combobox.tsx
+++ b/app/ui/lib/Combobox.tsx
@@ -15,7 +15,7 @@ import {
 } from '@headlessui/react'
 import cn from 'classnames'
 import { matchSorter } from 'match-sorter'
-import { useId, useState, type ReactNode } from 'react'
+import { useId, useState, type ReactNode, type Ref } from 'react'
 
 import { SelectArrows6Icon } from '@oxide/design-system/icons/react'
 
@@ -69,6 +69,8 @@ type ComboboxProps = {
   selectedItemLabel: string
   hasError?: boolean
   onChange: (value: string) => void
+  /** Necessary if you want RHF to be able to focus it on error */
+  inputRef?: Ref<HTMLInputElement>
 } & ComboboxBaseProps
 
 export const Combobox = ({
@@ -86,6 +88,7 @@ export const Combobox = ({
   onInputChange,
   allowArbitraryValues = false,
   hideOptionalTag,
+  inputRef,
   ...props
 }: ComboboxProps) => {
   const [query, setQuery] = useState(selectedItemValue || '')
@@ -147,6 +150,7 @@ export const Combobox = ({
                 : 'bg-default',
               hasError && 'focus-error'
             )}
+            ref={inputRef}
           />
           {items.length > 0 && (
             <ComboboxButton

```

david-crespo:
Ok, it's working now. Weird.

app/ui/lib/Combobox.tsx
charliepark: As of [the latest commit](https://github.com/oxidecomputer/console/pull/2474/commits/d610bc3d8a6f3ab8e95d7c3ac136b001b147cb77):
Screenshot 2024-10-03 at 1 05 32 PM

charliepark:
This also closes #2483

david-crespo (approved):
woohoo!

PR #2477, opened 2024-09-26
Update readme screenshot

charliepark:
We've adjusted a few things on the Instances list page since the last screenshot (April 30th), so this just gets a more up-to-date screenshot in the docs.

**Old:**
readme-screenshot-original

**New:**
readme-screenshot


david-crespo (approved):
Didn't realize we brightened the background a little!

david-crespo:
I was confused about the colors. Turns out we did not change anything, as far as I can tell. Rather, somehow, my original screenshot was darker than it should be. I went way back to a year ago on Vercel and the background color is the same as it is now.

https://console-85zkgfn7q-oxidecomputer.vercel.app/projects

PR #2480, opened 2024-09-27
Fix disableClear bug in Transit IP form; don't add duplicate IPs to list

charliepark:
Found a bug in the logic on the disableClear button for the "add a transit IP" form — checking whether the field has anything in it gives a more reliable signal about whether it should be clearable or not. Also, I was able to add some logic to prevent duplication of IPs in the list.


david-crespo:
Pulled this into #2461 in https://github.com/oxidecomputer/console/pull/2461/commits/a048e32dc53bbd198381af0260b631c4476c2c2e. Thanks for fixing.

PR #2487, opened 2024-10-04
Instance resize

david-crespo:
Based on https://github.com/oxidecomputer/omicron/pull/6774. Wait for that to merge so we can point at omicron main.


david-crespo:
I did something gnarly, but luckily it seems to have worked. I pushed my own version of the omicron branch with omicron main merged in to get everything _else_ up to date, then I bumped the API here with that commit, then I bumped omicron to main on console main, than I merged console main into here.

Short version: this is pretty up to date and once resize is merged in the API we will be in good shape. This still needs a bit of polish, though.

charliepark:
No rush at all; just wanted to note that https://github.com/oxidecomputer/omicron/pull/6774 has been merged.

david-crespo:
I think we could get the info box on one line. Don’t really need the instance name. Maybe just `Current: 8 CPU / 4 GiB` (whatever the canonical unit for CPU is)

IMG_7948

benjaminleonard:
I added that in there because you can resize from the index, at which point it might not be clear which instance you are resizing. Or at least it's helpful to reassure the user of their action.

charliepark:
Screenshot 2024-11-25 at 2 22 22 PM

charliepark:
It seemed useful to me to have the name in the box, unless there's some other way to verify which instance they're updating. I thought putting it in the modal header (Resize instance db1) could work, but that's not a pattern we have in other places in the app.

david-crespo:
Overflow with a long name is kinda barf, maybe I was wrong about one line. The modal just feels so stubby.

image

benjaminleonard:
Fixed the truncation. Design-wise looks good to me.

charliepark:
🎉

Issue #1527, opened 2023-05-23
Need breadcrumbs on nested detail pages


david-crespo:
We have several pages where you can navigate to a detail view and not have a clear way to get back to where you were:

* [Sleds](https://console-l5slchxz1-oxidecomputer.vercel.app/system/inventory/sleds) -> sled detail 
* ~[Silos](https://console-l5slchxz1-oxidecomputer.vercel.app/system/silos) -> silo detail~
* ~[IP pools](https://console-l5slchxz1-oxidecomputer.vercel.app/system/networking/ip-pools) -> IP pool detail~
* ~[VPCs](https://console-l5slchxz1-oxidecomputer.vercel.app/projects/mock-project/vpcs) -> VPC detail~
* ~VPC routers -> router detail w/ routes (coming soon in #2294)~

--- 

(This was the original issue, edited later to add the above, making it more general)

I think you need a picker in the top bar to visually distinguish the sled detail from the top-level inventory view. Silo detail also shown for comparison.

![2023-05-23-sled-detail](https://github.com/oxidecomputer/console/assets/3612203/0d8bab73-171e-4b87-9e8d-7af6ebd92b94)

PR #2488, opened 2024-10-07
Add read-only UI for internet gateways

### LINKED ISSUE #2402: Internet gateways tab on VPC page
Following https://github.com/oxidecomputer/omicron/pull/6475

charliepark:
This PR creates a new tab on the VPC view page, showing Internet Gateways that are a part of that VPC. It lists out the gateways, along with their attached IP Pools, IP Addresses, and any VPC routes that are targeting that particular internet gateway.
Screenshot 2024-12-10 at 3 21 30 PM

A sidemodal gives slightly more details, including the "Internet Gateway IP Address" and "Internet Gateway IP Pool", which is really more like a join table record between the Internet Gateway and an IP Pool. Current limitations stipulate that an Internet Gateway can have up to one attached IP Pool. The sidemodal also includes a table that shows the routes (and their routers) that include this gateway as a target value.
Screenshot 2024-12-10 at 4 41 58 PM

Finally, this PR also updates the router route form, so that a user can set up an internet gateway as the target for a VPC route.
Screenshot 2024-12-10 at 3 22 29 PM

Closes #2402


benjaminleonard:
In general I think we're better off holding for the next release to get the implementation right for churn reasons.

Let's catch up on this today. I might be a helpful eye for this, since I haven't read up on internet gateways yet, so I can play the idiot seeing this for the first time – a role a know and love.

I notice we're also missing breadcrumbs – is that #1527 ?

charliepark:
Good point on the breadcrumbs. Have added those in, but I think your larger point is a good one — there'll be more than just the breadcrumbs to discuss, and holding back on the feature for a release to get it right could be the better call. Will look forward to talking through it.

app/pages/project/vpcs/InternetGatewayPage/InternetGatewayPage.tsx
```diff
@@ -0,0 +1,119 @@
...
+        </div>
+      </PageHeader>
+      <PropertiesTable.Group className="mb-8">
+        <PropertiesTable>
+          <PropertiesTable.Row label="Description">
+            <DescriptionCell text={description} />
+          </PropertiesTable.Row>
+          <PropertiesTable.Row label="ID">
+            <Truncate text={id} maxLength={32} hasCopyButton />
+          </PropertiesTable.Row>
```
charliepark, line 88: This isn't necessary, as we have the `MoreActionsMenu` button in the header with the `Copy ID` button, but I didn't see a clear "fourth" aspect to the internet gateway metadata to highlight. If we do keep this, we can drop the `MoreActionsMenu`, as it's redundant. I suspect dropping this one is the better approach, though, if we can figure out what to put here in its place.

app/pages/project/vpcs/InternetGatewayPage/tabs/InternetGatewayIpPoolsTab.tsx
```diff
@@ -0,0 +1,98 @@
...
+      colHelper.accessor('description', Columns.description),
+      colHelper.accessor('ipPoolId', {
+        header: 'IP Pool Name',
+        cell: (info) => {
+          const ipPool = ipPools.items.find((item) => item.id === info.getValue())
+          return ipPool?.name || <EmptyCell />
+        },
+      }),
+      colHelper.accessor('ipPoolId', {
+        header: 'IP Pool Description',
```
charliepark, line 66: On http://localhost:4000/projects/mock-project/floating-ips we handle these as tooltips, hovering over the IP Pool name. Could do the same here if having them as columns in the table is confusing.

app/util/__snapshots__/path-builder.spec.ts.snap
```diff
@@ -709,6 +709,103 @@ exports[`breadcrumbs 2`] = `
...
+      "path": "/projects/p/vpcs",
+    },
+    {
+      "label": "v",
+      "path": "/projects/p/vpcs/v",
+    },
+    {
+      "label": "g",
+      "path": "/projects/p/vpcs/v/internet-gateways/g",
+    },
```
david-crespo, line 732: `Internet Gateways` is missing here before `g`

app/routes.tsx
```diff
@@ -399,6 +412,33 @@ export const routes = createRoutesFromElements(
...
               </Route>
+              <Route
+                path="internet-gateways/:gateway"
+                element={<InternetGatewayPage />}
+                loader={InternetGatewayPage.loader}
+                handle={makeCrumb(
+                  (p) => p.gateway!,
+                  (p) => pb.vpcInternetGateway(getInternetGatewaySelector(p))
+                )}
+              >
```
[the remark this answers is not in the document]
charliepark: Ah, thank you.

app/pages/project/vpcs/VpcPage/tabs/VpcGatewaysTab.tsx
```diff
@@ -5,3 +5,64 @@
...
+
+const colHelper = createColumnHelper<InternetGateway>()
+
+VpcInternetGatewaysTab.loader = async ({ params }: LoaderFunctionArgs) => {
+  const { project, vpc } = getVpcSelector(params)
+  const query = { project, vpc, limit: ALL_ISH }
+  await Promise.all([
+    apiQueryClient.prefetchQuery('internetGatewayList', { query }),
+    apiQueryClient.prefetchQuery('internetGatewayIpAddressList', { query }),
+    apiQueryClient.prefetchQuery('internetGatewayIpPoolList', { query }),
```
david-crespo, line 30: While merging the `useQueryTable` change in here I noticed the `gateway` query param is not required on these two list endpoints, which is why they can be called here without a type error despite the fact that there is no gateway in scope. This turns out to be a bug in the OpenAPI schema generation — the gateway param is required in the Rust code but not in the OpenAPI schema.

https://github.com/oxidecomputer/omicron/issues/7155

This isn't really a problem for us, we just have to make sure to call it correctly, i.e., with the gateway. Here we can just remove the calls. On the gateway page you have it right.

app/util/__snapshots__/path-builder.spec.ts.snap
```diff
@@ -709,6 +709,55 @@ exports[`breadcrumbs 2`] = `
...
+    {
+      "label": "v",
+      "path": "/projects/p/vpcs/v/firewall-rules",
+    },
+    {
+      "label": "Internet Gateways",
+      "path": "/projects/p/vpcs/v/internet-gateways",
+    },
+  ],
+  "vpcInternetGatewaysNew (/projects/p/vpcs/v/internet-gateways-new)": [],
```
david-crespo, line 760: note the lack of crumbs here

app/routes.tsx
```diff
@@ -375,6 +380,22 @@ export const routes = createRoutesFromElements(
...
+                >
+                  <Route
+                    path=":gateway"
+                    element={<EditInternetGatewayForm />}
+                    loader={EditInternetGatewayForm.loader}
+                    handle={makeCrumb(
+                      (p) => p.gateway!,
+                      (p) => pb.vpcInternetGateway(getInternetGatewaySelector(p))
+                    )}
+                  />
```
david-crespo, line 397: I think this should be a title-only crumb. I tried to follow a rule that side modals don't get a crumb. It looks weird for it to pop up in the background.

david-crespo:
Updated the info box copy and added a docs link:

image

app/pages/project/vpcs/VpcPage/tabs/VpcGatewaysTab.tsx
```diff
@@ -0,0 +1,179 @@
...
+  return routes.items
+    .filter((r) => r.target.type === 'internet_gateway' && r.target.value === gateway)
+    .map((route) => (
+      <Link
+        key={route.name}
+        to={pb.vpcRouterRouteEdit({ project, vpc, router, route: route.name })}
+        className="link-with-underline text-sans-md"
+      >
+        {route.name}
+      </Link>
```
david-crespo, line 70: I find clicking this link to the route edit form very jarring. We have some similar links in other places (linked silos and IP pools both link to each other, disks link to the instances they're attached to) but somehow this feels much more confusing. I thought maybe the problem was being dropped in the form (because you're in a modal and can't see much behind it), so maybe landing in the list of routes would be better, but landing on the router route list page instead is similarly disorienting. This isn't really something wrong with this PR (it's more about the underlying issue of deep nesting in the VPC tree, which is why @benjaminleonard wanted to keep the gateway details in a modal instead of a whole page) so I'm not sure we can fix it here either, unless we were to drop the link altogether. A more elaborate fix would be to open the route details in a side modal on top of the gateways page, but it's never great to have two canonical locations for a thing.

david-crespo: Did we figure out a ballpark likelihood of there being multiple routes targeting a gateway? It doesn't seem like that weird a thing to do. In any case we need to figure out how to render this better as a list — there are 3 here and no spaces between them. I think this is why we were considering not putting this in the cell at all.

image

david-crespo:
Looks pretty good overall. My main open question is how to handle multiple routes in the gateways table cell, possibly by removing that column. Secondary question of whether there's a less jarring thing we can do than linking to the route edit form.

PR #2492, opened 2024-10-09
Fix listbox IDs; better link labels to inputs

### LINKED ISSUE #2417: Fix `id` props in `Listbox` and `Combobox`

charliepark:
Closes #2417 

This does two things. One, it gets rid of the strange empty `id` that had been in the Listbox elements. (We had already fixed Comboboxes in #2474.)

This also better links the Label and Listbox element, so when you click on the label, the focus goes to the Listbox.

### DIFF
--- a/app/ui/lib/Listbox.tsx
+++ b/app/ui/lib/Listbox.tsx
@@ -7,13 +7,12 @@
  */
 import {
   Listbox as HListbox,
-  Label,
   ListboxButton,
   ListboxOption,
   ListboxOptions,
 } from '@headlessui/react'
 import cn from 'classnames'
-import { type ReactNode, type Ref } from 'react'
+import { useId, type ReactNode, type Ref } from 'react'
 
 import { SelectArrows6Icon } from '@oxide/design-system/icons/react'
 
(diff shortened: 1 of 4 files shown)


david-crespo (approved):
Looks good. Do you know why this change fixes the click on the label? More specifically, any idea why it doesn't work on main? On main, the headless `<Label>` component is hooking up the IDs for us, resulting in the correct accessible name on the button, but I confirmed clicks on it don't focus the field. Weird.

image

Here's what it looks like on this branch. Maybe the fact that we have both the `id` + `for` thing _and_ `aria-labelledby` is making it mad on main. Here we only use the former.

image

charliepark:
I'm trying a few different things in dev tools to get main to function properly, but it doesn't seem to cooperate. Removing `aria-labelledby` didn't really change it, from what I could tell. Will keep an eye on it on main, though, to see if there's (for some inexplicable reason) a regression from the current branch's working state. Not expecting that, but it's odd that the current version on main seems like it should be working.

PR #2494, opened 2024-10-09
Dynamic units

charliepark:
Open to general conversation on this.

Currently, we have a few set units that we apply to our Memory (GiB) and Storage (TiB) utilization representations. The important thing is that they should be the same unit on each tile:
Screenshot 2024-10-07 at 5 08 49 PM

As we move to having larger drives in the future, we probably won't be able to assume that GiB and TiB will be reliable values for the Memory and Storage tiles (and other representations). This PR introduces a dynamic way to determine the appropriate unit and to convert the numbers.

I am open to any thoughts on this. For example, is the core assumption here off? Is it more straightforward to a user to say `3000 GiB` rather than recasting it as `TiB`?

There are a few places where we'd need to revisit our existing "toGiB"-style conversions, but I think this is a good proof of concept so we can talk through it.

Here's one spot, for example, where we're always running a "to GiB" conversion (note the bottom row), but where a dynamic converter might make more sense …
Screenshot 2024-10-09 at 1 17 46 AM
… I was focused on embiggening the middle row, but then noticed that since I added the dynamic conversion, both rows now show the proper (TiB) value. (Note: TiB for the bottom row was because it was already large enough for that, not based off of the middle row's new TiB scale.)
Screenshot 2024-10-09 at 1 17 28 AM

### DIFF
--- a/test/e2e/utilization.e2e.ts
+++ b/test/e2e/utilization.e2e.ts
@@ -116,6 +116,37 @@ test.describe('Silo utilization', () => {
   })
 })
 
+test('Utilization shows correct units for CPU, memory, and storage', async ({ page }) => {
+  await page.goto('/system/utilization')
+
+  // Verify the original values for the Memory tile
+  await expect(page.getByText('Memory(GIB)')).toBeVisible()
+  await expect(page.getByText('Provisioned384 GiB')).toBeVisible()
+  await expect(page.getByText('Quota (Total)800 GiB')).toBeVisible()
+
+  // Navigate to the quotas tab
+  await page.goto('system/silos/maze-war?tab=quotas')
+
+  // Verify that there's a row for memory with the correct units
+  await expect(page.getByText('Memory234 GiB300 GiB')).toBeVisible()
+
+  // Edit the quota and verify the new value
+  await page.getByRole('button', { name: 'Edit quotas' }).click()
+  await page.getByRole('textbox', { name: 'Memory (GiB)' }).fill('3000')
+  await page.getByRole('button', { name: 'Update quotas' }).click()
+  // The table has been updated
+  await expect(page.getByText('Memory0.23 TiB2.93 TiB')).toBeVisible()
+
+  // Navigate back to the utilization page without refreshing
+  await page.getByRole('link', { name: 'Utilization' }).click()
+  await expect(page.getByRole('heading', { name: 'Utilization' })).toBeVisible()
+
+  // Verify the updated values for the Memory tile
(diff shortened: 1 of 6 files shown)


test/e2e/utilization.e2e.ts
```diff
@@ -93,8 +93,8 @@ test.describe('System utilization', () => {
     await expectRowVisible(page.getByRole('table'), {
       Silo: 'all-zeros',
       CPU: '0',
-      Memory: '0 GiB',
-      Storage: '0 TiB',
+      Memory: '0 B',
+      Storage: '0 B',
```
charliepark, line 97: A thing to talk about, for example. Is `B` fine here? Do we want, instead, some minimum (`GiB`?)

david-crespo:
I agree it would be nice to get things into more natural units, though sometimes we might be abbreviating too much. Maybe it makes sense to always show some digits after the decimal? I know it's relatively rare to have exactly `7.00 TiB` as the value , but when I see `7 TiB`, I wonder: is that rounded? What is the real number?

benjaminleonard:
Should we use `toLocaleString` to keep the commas for values >= 1,000?

That might be less frequent with the dynamic units, but still. Especially where the units are set within a tile.

My other question is whether those are the right thresholds to switch? I had a vague thought that we might have a preference for a unit and switch if it's significantly higher or lower but that's probably overcomplicating it.

david-crespo:
`round` uses `Intl.NumberFormat` internally, so we should still get the nice commas. I actually agree about complicating the thresholds, I think we could get a better result.

https://github.com/oxidecomputer/console/blob/98103a25f34cbcdeadbf8c62dbd41bfef65ee0a0/app/util/math.ts#L51-L61

charliepark:
On the decimals / commas / return types …
One of the issues with returning strings is that we do sometimes use the values downstream, like in CapacityBar, where we call `const pct = percentage(provisioned, capacity)`. Also, there are a few places where it looks like we use BigInt values, like — also in CapacityBar — where we use `bigint` for the allocated IPv6 addresses. From http://localhost:4000/system/networking/ip-pools/ip-pool-4:
Screenshot 2024-10-09 at 5 29 08 PM

The good news is that the decimals are getting dropped because they're `.00` or .X0` … if they're non-zero, `.XY`, which I suspect they will be most of the time, they'll be rendered.

So — open to thoughts, but — I think we can leave them as-is, as the likelihood that they'll be an even BinaryUnit multiple — while non-zero (wokka wokka) — is probably low.

david-crespo:
We could calculate percentages from the raw byte values.

PR #2495, opened 2024-10-09
Instance resize modal

benjaminleonard:
Building on your PR @david-crespo 

image

As you said, modal seems to work much better here. Kept it as a route so we can link to it from the instances list page. One thing to keep in mind is because it's a link, you can get here even if the instance is not currently stopped – so we should highlight that, even if the API would prevent it.

image

We can reuse the number fields from the instance create form. On that, I think we should remove the number limit and instead let the form validation describe to the user why they cannot create an instance above the CPU and memory threshold rather than just preventing it.

image

Unsure if this error handling is sufficient. Do we want to expand the modal to show errors like the side modal form, instead of the toast?

### DIFF
--- a/app/forms/instance-resize.tsx
+++ b/app/forms/instance-resize.tsx
@@ -7,20 +7,24 @@
  */
 import { useForm } from 'react-hook-form'
 import { useNavigate, type LoaderFunctionArgs } from 'react-router-dom'
-import * as R from 'remeda'
 
 import {
   apiQueryClient,
+  INSTANCE_MAX_CPU,
+  INSTANCE_MAX_RAM_GiB,
+  instanceCan,
   useApiMutation,
   useApiQueryClient,
   usePrefetchedApiQuery,
 } from '@oxide/api'
 
 import { NumberField } from '~/components/form/fields/NumberField'
-import { SideModalForm } from '~/components/form/SideModalForm'
 import { getInstanceSelector, useInstanceSelector } from '~/hooks/use-params'
 import { addToast } from '~/stores/toast'
+import { Message } from '~/ui/lib/Message'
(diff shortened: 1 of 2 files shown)


david-crespo:
Better for it not to be a link and just pop the modal on the list view when you're there, and pop it on the detail view when you're there. It might be as simple as `{resizeModalOpen && <ResizeModal />}` in each page with a `useState` controlling the modal state. I can try it soon if you don't get to it.

benjaminleonard:
Sounds good. Why do you think it's better? Simpler implementation?

david-crespo:
More the UX. I think it’s weird for clicking resize on instance list to take you to a different page when it doesn’t have to.

benjaminleonard:
Ah of course, I guess I was only really looking at it from the root instance view route.

benjaminleonard:
Trying to figure out a treatment that tells you the instance name, since you can do this direct from the instances list and it might not be clear which instance is selected.

image

Similar to silo images demote:
image

image

Side modal resource:
image

image

app/pages/project/instances/actions.tsx
```diff
@@ -119,8 +115,8 @@ export const useMakeInstanceActions = (
         },
         {
           label: 'Resize',
-          onActivate() {
-            navigate(pb.instanceResize(instanceSelector))
+          onActivate: () => {
+            options.onResizeClick && options.onResizeClick(instance)
```
benjaminleonard, line 119: Interestingly this does not fail when run locally.

But also there must be a better way of doing this? `onResizeClick` should not be optional but our declaration of options as `{}` insists on it.

benjaminleonard:
Converted over to a regular non-route modal.

CleanShot 2024-10-10 at 12 43 18

I think that works well.

PR #2497, opened 2024-10-09
Extract IpPoolCell so we can use it in other places

charliepark:
We'll likely be using the IpPoolCell in the new Network Gateways page; this extracts it out to its own component.

We were already following the "fetch the list and then seed the ReactQuery cache with relevant data" approach, so this is really just moving the component around.


david-crespo (approved):
Nice. Something worth trying: `if (!content) return child` on line 88 inside `Tooltip`. That would eliminate the conditional in the cell and seems like it could be useful elsewhere.

https://github.com/oxidecomputer/console/blob/f0c161de77958a6f3a5881772a194a218fa5eedf/app/ui/lib/Tooltip.tsx#L87-L94

Issue #2361, opened 2024-08-09
Error when trying to attach a floating IP when an instance is starting


augustuswm:
If you attempt to attach a floating IP while an instance is in the starting state you will get the following error:

![image](https://github.com/user-attachments/assets/d2e18ec3-a96e-40c3-9c42-53128a4de5f3)

This is an expected behavior, but in the future we should likely prevent this action until the instance is in a state that will accept the request.

david-crespo:
We should try and get a better error message out the API there too. It’s possible that we are getting one and not displaying it because it’s a 503. Also is 503 really the right status code for this?

david-crespo:
This would be pretty easy to fix by adding something like `modifyIp: ['running', 'rebooting', 'stopped']` to the list here and disabling attach/detach in other other states:

https://github.com/oxidecomputer/console/blob/efcaba790188949af73e8e0bdd6e82cdf76f1fe2/app/api/util.ts?plain=1#L92-L123

The problem is I need to nail down exactly which states allow IP attach/detach. It looks like [this](https://github.com/oxidecomputer/omicron/blob/62c7672eb850d8561e91252beff7e312d86970d3/nexus/src/app/sagas/instance_common.rs?plain=1#L228-L323) is probably the relevant code. @gjcolombo or @FelixMcFelix, does `['running', 'rebooting', 'stopped']` sound good for states you're allowed to attach or detach IPs in?

gjcolombo:
I think that's right, from my recollection of what the code you linked is trying to do (I agree that it's the relevant bit here).

david-crespo:
This feels like mostly an API problem — a better status code and error message would show up in the console automatically.

PR #2505, opened 2024-10-15
minor: refactor `instanceCan` and `diskCan` for better type safety

david-crespo:
I noticed while working on #2361 that `instanceCan` allows any string as a key, which means you can do `instanceCan.detaaaaaach(instance)` and not get a type error. The goal here was to make that a type error without affecting the API, so we don't have to change any call sites. ~~After some back and forth with myself around several ways one could make this work, I landed on this, which is more verbose than I prefer, but it's fine.~~ I did it.

Rather than try to explain the other stuff I tried, here is a fun and exciting video of me suffering.

https://github.com/user-attachments/assets/c289c8eb-f090-42ec-824e-3892d43d66e2

### DIFF
--- a/app/api/util.spec.ts
+++ b/app/api/util.spec.ts
@@ -7,7 +7,7 @@
  */
 import { describe, expect, it, test } from 'vitest'
 
-import { genName, parsePortRange, synthesizeData } from './util'
+import { diskCan, genName, instanceCan, parsePortRange, synthesizeData } from './util'
 
 describe('parsePortRange', () => {
   describe('parses', () => {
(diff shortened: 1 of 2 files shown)


app/api/util.spec.ts
```diff
@@ -136,3 +136,22 @@ describe('synthesizeData', () => {
...
+
+test('diskCan', () => {
+  expect(diskCan.delete({ state: { state: 'creating' } })).toBe(false)
+  expect(diskCan.delete({ state: { state: 'attached', instance: 'xyz' } })).toBe(false)
+  expect(diskCan.delete({ state: { state: 'detached' } })).toBe(true)
+
+  // @ts-expect-error typechecker rejects actions that don't exist
+  // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+  diskCan.abc
+})
```
david-crespo, line 157: Confirmed the `@ts-expect-error`s fail on main:

Screenshot 2024-10-15 at 2 03 47 PM

david-crespo:
Well, recording that video made me realize how to do it the nice way. So that's nice.

PR #2506, opened 2024-10-16
How about we don't hold the entire image in memory

### LINKED ISSUE #2270: Investigate: image upload consumes a lot of memory
Below is just suggestive — I didn't look far enough into it to be sure there's an issue, but it doesn't look very good. I also wasn't able to get the issue in Chrome, so maybe v8 is smarter about this than SpiderMonkey.

---

Inspired by https://schiener.io/2024-05-29/react-query-leaks, I was curious whether I could get the console to hold onto a bunch of memory. I didn't see an issue with normal browsing, which makes sense because it is mostly about closures containing large request bodies, and our request bodies are tiny.

However, image upload sends the entire contents of an image in chunks, so that seemed like a good candidate. Sure enough, taking snapshots showed memory increasing and not getting GCed, even when I cancel the upload and navigate around for a bit. All the memory is in "strings", which makes sense because we are reading each chunk as a base64ed string.

https://github.com/oxidecomputer/console/blob/13ce748fb0331177788db74316e9e61482447bd7/app/forms/image-upload.tsx#L390-L415

2024-06-07-memory-snapshots-image-upload

Using Firefox nightly, I was able to get a profile showing a rather huge (I would bet overstated) amount of memory being used during image upload.

[2024-06-07 11.24 firefox-nightly-profile.json.gz](https://github.com/user-attachments/files/15742522/2024-06-07.11.24.firefox-nightly-profile.json.gz)

Screenshot 2024-06-07 at 11 32 28 AM

david-crespo:
Closes #2270

Reduce `gcTime` on upload chunk React Query mutation from default of 5 minutes to zero.

As I say in the comment, I don't even understand why the mutation cache exists — it's not like we're reading from it. The query cache is a different story. So I don't see a downside to turning it off for this mutation. I have asked about in the Tanstack Discord and on Twitter, so I'll see if anything helpful comes out of that.

https://tanstack.com/query/v5/docs/reference/MutationCache
https://tanstack.com/query/v5/docs/framework/react/reference/useMutation
https://tkdodo.eu/blog/mastering-mutations-in-react-query

### Before

Taking some snapshots in FF throughout a big upload. Memory goes up and up.

Screenshot 2024-10-15 at 3 58 10 PM

### After

Memory fluctuates around a relatively small value, consistent with GC functioning as desired.

image

### With shorter but non-zero `gcTime`

As a test, I set `gcTime: 60000` (1 minute) and saw as expected that memory use still plateaus, but at a higher number because it holds onto everything it can process in one minute.

image

### DIFF
--- a/app/forms/image-upload.tsx
+++ b/app/forms/image-upload.tsx
@@ -215,7 +215,13 @@ export function CreateImageSideModalForm() {
 
   const createDisk = useApiMutation('diskCreate')
   const startImport = useApiMutation('diskBulkWriteImportStart')
-  const uploadChunk = useApiMutation('diskBulkWriteImport')
+
+  // gcTime: 0 prevents the mutation cache from holding onto all the chunks for
+  // 5 minutes. It can be a ton of memory. To be honest, I don't even understand
+  // why the mutation cache exists. It's not like the query cache, which dedupes
+  // identical queries made around the same time.
+  // https://tanstack.com/query/v5/docs/reference/MutationCache
+  const uploadChunk = useApiMutation('diskBulkWriteImport', { gcTime: 0 })
 
   // synthetic state for upload step because it consists of multiple requests
   const [syntheticUploadState, setSyntheticUploadState] =


david-crespo:
I can't see a downside here, so I'm going to just merge this. If we learn something new we have plenty of time to revert before the next release.

david-crespo:
I think we're in the clear. We don't need these mutations in any cache. We might not need _any_ mutations in the cache, but we can think about that another time. The idea of looking at mutation state without having to pass the original mutation object around is pretty cool.

image

https://x.com/TkDodo/status/1846779928768836033

PR #2510, opened 2024-10-17
Automatic ACS URL in IdP forms w/ copy button

### LINKED ISSUE #2472: Compute ACS Url in IdP form
When creating a SAML identity provider connection, the ACS URL that the user enters should always be of the form:

`http(s)://<silo>.sys.<subdomain>/login/<silo>/saml/<idp>`

where `<silo>` is the Silo name, `<subdomain>` is the subdomain assigned to the rack, and `<idp>` is the name of the IdP connection. This means that when a user is filling out the form, we have all of the necessary information to compute this value once they enter a value in to the name field.

By computing this we can also make the field readonly to prevent accidental changes.

charliepark:
https://console-git-automaticacsurl-oxidecomputer.vercel.app/system/silos/maze-war/idps-new

This adds a click-to-copy component as an option for inputs, and then uses that new feature in the IDP form to allow the user to copy an auto-generated ACS URL.

Augustus had the very good idea (in #2472):
>When creating a SAML identity provider connection, the ACS URL that the user enters should always be of the form:
>
>`http(s)://<silo>.sys.<subdomain>/login/<silo>/saml/<idp>`
>
>where <silo> is the Silo name, <subdomain> is the subdomain assigned to the rack, and <idp> is the name of the IdP connection.

This PR, then, sets up that form to have the ACS value automatically generated.

Screenshot 2024-10-17 at 3 35 26 PM

Closes #2472

---

There are a two things I have questions on.

1. I think the visual treatment of the "copy" button could use a little tweaking, but I've had a really hard time getting Tailwind to cooperate. Specifically, giving the button (or a wrapper div for it) a left-border and possibly a different background color.
2. `fieldProps.value as string` … would rather not cast it, but none of the other things I tried seemed to handle it.


david-crespo:
`value` has a funky type because it gets it from the native input, but we could override it to just be `string`:

```diff
diff --git a/app/ui/lib/TextInput.tsx b/app/ui/lib/TextInput.tsx
index 786c1ee6..5e64bcef 100644
--- a/app/ui/lib/TextInput.tsx
+++ b/app/ui/lib/TextInput.tsx
@@ -8,6 +8,7 @@
 import { announce } from '@react-aria/live-announcer'
 import cn from 'classnames'
 import React, { useEffect } from 'react'
+import type { Merge } from 'type-fest'
 
 import { CopyToClipboard } from './CopyToClipboard'
 
@@ -34,14 +35,18 @@ export type TextAreaProps =
 // it makes a bunch of props required that should be optional. Instead we simply
 // take the props of an input field (which are part of the Field props) and
 // manually tack on validate.
-export type TextInputBaseProps = React.ComponentPropsWithRef<'input'> & {
+export type TextInputBaseProps = Merge<
+  React.ComponentPropsWithRef<'input'>,
+  {
     // error is used to style the wrapper, also to put aria-invalid on the input
     error?: boolean
     disabled?: boolean
     className?: string
     fieldClassName?: string
     copyable?: boolean
+    value?: string
   }
+>
 
 export const TextInput = React.forwardRef<
   HTMLInputElement,
@@ -88,7 +93,7 @@ export const TextInput = React.forwardRef<
         />
         {copyable && (
           <CopyToClipboard
-            text={fieldProps.value as string}
+            text={fieldProps.value}
             className="flex h-full items-stretch border-l border-solid px-3 border-default hover:border-hover"
           />
         )}
```

It doesn't seem like we're passing anything other than string to it anywhere because there are no type errors when I do that, except `text` is still unhappy because it's not allowed to be undefined. 

```
$ ts
app/ui/lib/TextInput.tsx:96:13 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

96             text={fieldProps.value}
               ~~~~

  app/ui/lib/CopyToClipboard.tsx:19:3
    19   text: string
         ~~~~
    The expected type comes from property 'text' which is declared here on type 'IntrinsicAttributes & Props'


Found 1 error in app/ui/lib/TextInput.tsx:96
```

In this particular case, we know it won't be undefined because we're constructing it and passing it in, but the only way to ensure that in the code is something like

```ts
{copyable && fieldProps.value && <CopyToClipboard />}
```

Which I guess is fine, though it risks us accidentally doing something weird later where we set `copyable` but don't pass in a value or pass in a falsy value. Maybe a slightly stronger check would be `fieldProps.value !== undefined`.

Another option would be to not use a real input at all here, and instead construct a little div that _looks_ like an input, then we don't have to modify `TextInput` to support this.

david-crespo:
Kinda sad that you can't see the URL without copying it and pasting it somewhere but I guess that's ok. Not sure what visual treatment would let you see the whole thing.

image

charliepark:
Tooltip?
Screenshot 2024-10-18 at 10 18 35 AM

app/forms/idp/create.tsx
```diff
@@ -111,8 +117,15 @@ export function CreateIdpSideModalForm() {
         name="acsUrl"
         label="ACS URL"
         description="Service provider endpoint for the IdP to send the SAML response"
+        // When creating a SAML identity provider connection, the ACS URL that the user enters
+        // should always be of the form: http(s)://<silo>.sys.<subdomain>/login/<silo>/saml/<name>
+        // where <silo> is the Silo name, <subdomain> is the subdomain assigned to the rack,
+        // and <name> is the name of the IdP connection
+        value={`https://${silo}.sys.${subdomain}/login/${silo}/saml/${name}`}
```
david-crespo, line 124: Looks like I was wrong about this working fine. It works fine to display it in the input, but it doesn't appear to be syncing with the RHF form state like I thought it would. Not sure if there's a way to make it do that or maybe we are actually stuck with the `useEffect`. I thought about using an `onChange` on the name field, which is usually preferable, but at least as it is implemented now, passing `onChange` to `NameField` breaks it by overrides the `onChange` that hooks the field up to the RHF state.

image

app/forms/idp/create.tsx
```diff
@@ -22,6 +22,11 @@ import { pb } from '~/util/path-builder'
 
 import { MetadataSourceField, type IdpCreateFormValues } from './shared'
 
+const host = window.document.location.host.toString()
```
david-crespo, line 25: should probably be `window.location.hostname`

app/forms/idp/create.tsx
```diff
@@ -22,6 +22,11 @@ import { pb } from '~/util/path-builder'
...
+const subdomain = host.includes('localhost')
+  ? 'r3.oxide-preview.com'
+  : host.split('.').slice(2).join('.')
```
david-crespo, line 28: We need to make super sure of this one.

app/forms/idp/create.tsx
charliepark: Ah, crumbs; well, easily revertable. I guess we know now why we set the form the way we do!

app/forms/idp/create.tsx
charliepark: Will move it to a util and get some tests in

app/forms/idp/create.tsx
david-crespo: It's not that hard to make `onChange` stop being such a jerk, but it's not a complete fix — `calling `setValue` in `onChange` does work, but you also have to set the initial value. I don't love it, just an idea.

```diff
diff --git a/app/components/form/fields/TextField.tsx b/app/components/form/fields/TextField.tsx
index 9163a02e..ab576c89 100644
--- a/app/components/form/fields/TextField.tsx
+++ b/app/components/form/fields/TextField.tsx
@@ -118,7 +118,11 @@ export const TextFieldInner = <
         type={type}
         error={!!error}
         aria-labelledby={`${id}-label ${id}-help-text`}
-        onChange={(e) => onChange(transform ? transform(e.target.value) : e.target.value)}
+        onChange={(value) => {
+          const transformed = transform ? transform(value) : value
+          onChange(transformed)
+          props.onChange?.(transformed)
+        }}
         {...fieldRest}
         {...props}
       />
diff --git a/app/forms/idp/create.tsx b/app/forms/idp/create.tsx
index 601abc65..4d1e44e5 100644
--- a/app/forms/idp/create.tsx
+++ b/app/forms/idp/create.tsx
@@ -22,10 +22,8 @@ import { pb } from '~/util/path-builder'
 
 import { MetadataSourceField, type IdpCreateFormValues } from './shared'
 
-const host = window.document.location.host.toString()
-const subdomain = host.includes('localhost')
-  ? 'r3.oxide-preview.com'
-  : host.split('.').slice(2).join('.')
+const host = window.location.hostname
+const subdomain = host === 'localhost' ? 'localhost' : host.split('.').slice(2).join('.')
 
 const defaultValues: IdpCreateFormValues = {
   type: 'saml',
@@ -63,8 +61,16 @@ export function CreateIdpSideModalForm() {
     },
   })
 
-  const form = useForm({ defaultValues })
-  const name = form.watch('name')
+  function getAcsUrl(name: string) {
+    return `https://${silo}.sys.${subdomain}/login/${silo}/saml/${name}`
+  }
+
+  const form = useForm({
+    defaultValues: {
+      ...defaultValues,
+      acsUrl: getAcsUrl(defaultValues.name),
+    },
+  })
 
   return (
     <SideModalForm
@@ -111,17 +117,20 @@ export function CreateIdpSideModalForm() {
       submitError={createIdp.error}
       submitLabel="Create provider"
     >
-      <NameField name="name" control={form.control} />
+      <NameField
+        name="name"
+        control={form.control}
+        // When creating a SAML identity provider connection, the ACS URL that the user enters
+        // should always be of the form: http(s)://<silo>.sys.<subdomain>/login/<silo>/saml/<name>
+        // where <silo> is the Silo name, <subdomain> is the subdomain assigned to the rack,
+        // and <name> is the name of the IdP connection
+        onChange={(name) => form.setValue('acsUrl', getAcsUrl(name))}
+      />
       <DescriptionField name="description" control={form.control} required />
       <TextField
         name="acsUrl"
         label="ACS URL"
         description="Service provider endpoint for the IdP to send the SAML response"
-        // When creating a SAML identity provider connection, the ACS URL that the user enters
-        // should always be of the form: http(s)://<silo>.sys.<subdomain>/login/<silo>/saml/<name>
-        // where <silo> is the Silo name, <subdomain> is the subdomain assigned to the rack,
-        // and <name> is the name of the IdP connection
-        value={`https://${silo}.sys.${subdomain}/login/${silo}/saml/${name}`}
         required
         disabled
         control={form.control}
@@ -135,6 +144,7 @@ export function CreateIdpSideModalForm() {
         description="Service provider endpoint for log out requests"
         required
         control={form.control}
+        defaultValue={`https://${silo}.sys.${subdomain}/login/${silo}/saml/`}
       />
       {/* TODO: help text */}
       <TextField
diff --git a/app/ui/lib/TextInput.tsx b/app/ui/lib/TextInput.tsx
index bb635b50..ab2c8bf4 100644
--- a/app/ui/lib/TextInput.tsx
+++ b/app/ui/lib/TextInput.tsx
@@ -8,6 +8,7 @@
 import { announce } from '@react-aria/live-announcer'
 import cn from 'classnames'
 import React, { useEffect } from 'react'
+import { type Merge } from 'type-fest'
 
 import { CopyToClipboard } from './CopyToClipboard'
 import { Tooltip } from './Tooltip'
@@ -35,14 +36,18 @@ export type TextAreaProps =
 // it makes a bunch of props required that should be optional. Instead we simply
 // take the props of an input field (which are part of the Field props) and
 // manually tack on validate.
-export type TextInputBaseProps = React.ComponentPropsWithRef<'input'> & {
+export type TextInputBaseProps = Merge<
+  React.ComponentPropsWithRef<'input'>,
+  {
     // error is used to style the wrapper, also to put aria-invalid on the input
     error?: boolean
     disabled?: boolean
     className?: string
     fieldClassName?: string
     copyable?: boolean
+    onChange?: (value: string) => void
   }
+>
 
 export const TextInput = React.forwardRef<
   HTMLInputElement,
@@ -58,6 +63,7 @@ export const TextInput = React.forwardRef<
       fieldClassName,
       copyable,
       as: asProp,
+      onChange,
       ...fieldProps
     },
     ref
@@ -77,6 +83,7 @@ export const TextInput = React.forwardRef<
         )}
         aria-invalid={error}
         disabled={disabled}
+        onChange={(e) => onChange?.(e.target.value)}
         {...fieldProps}
       />
     )
```

david-crespo: probably better to stick with the `useEffect` for now 😁

charliepark:
Screenshot 2024-10-25 at 1 52 27 PM

charliepark:
Screenshot 2024-10-25 at 2 10 10 PM

charliepark:
As discussed in chat … using a full-height button and making the entire area clickable:
Screenshot 2024-10-28 at 9 15 00 AM

charliepark:
Now matches the other dropdown buttons:
Screenshot 2024-10-28 at 9 21 00 AM

charliepark:
I noticed that the combobox border was running full-height, like the number field / copyable field …
Screenshot 2024-10-28 at 11 33 38 AM
… so I updated it to reflect the regular Listbox style, with the top and bottom gap …
Screenshot 2024-10-28 at 11 33 49 AM

david-crespo:
We don't have an e2e test exercising the IdP create form! I will add one.

david-crespo (commented):
this is awesome. It works really well. The only thing I wanna do before merging is make sure there’s really no way the user can be accessing the console through some other domain (through a proxy, maybe) that could cause us to be wrong to pull the delegated domain out of `window.location`.

david-crespo:
Based on some discussion in [chat](https://matrix.to/#/!oSMXTSrgambaNFEemN:oxide.computer/$02f2sYpr97aUIDs6fc8Q4aHLD7j3cb2qSDsRk7EPf1Y?via=oxide.computer&via=unix.house&via=matrix.org), it seems like my concern about a proxy is founded. 

> If the proxy is doing TLS termination and initiating a new TLS connection to us, then yes, I think what you're describing would be a problem.

The idea would be that there is a "true" rack hostname that the IdP is supposed to use when talking to the rack, but the user is hitting the rack through a proxy that gives them a different one, so putting the latter in the ACS URL would be wrong. This only works if the proxy is doing TLS termination because otherwise it could not change the hostname on the incoming request, and Nexus would reject it because Nexus relies on the hostname being the one it knows about.

This all seems kind of unlikely, so one option is to do nothing and see if it becomes a problem. If it is, the user can always use the CLI instead. I don't love that because even though it's unlikely, it's a wrinkle that pops up at rack setup time, which is an especially critical time for customer perception of the product.

### Options

1. Don't change anything. In the unlikely event this comes up, tell the user to use the CLI to create the IdP, and then fix the console UI in the next release based on what we learn about the use case.
2. Start with the field disabled and auto-generated, have add a checkbox or edit button under it that would allow an override if necessary.
    - If checkbox: when unchecked we just make it a regular text field, i.e., don't touch it when the IdP name changes. If they unchecked it and checked it again, it would have to revert back to the generated version
3. Don't disable the field, auto-generate it only as long as the field has not been touched
    - This is bad because it's not clear about when it's being auto-generated vs. not.

Any other ways we could do it?

charliepark:
Screen Recording 2024-11-07 at 2 42 58 PM mov

Open to copy suggestions on both (1) microcopy above ACS URL input and (2) checkbox label

david-crespo:
I'm thinking we can overexplain in the help text and that will let us keep the checkbox label short. I'm also realizing that "Service provider" is a lot less clear than what that really means, which is the rack. Maybe something more like:

> Oxide endpoint for the identity provider to send the SAML response. URL is generated from the current URL, silo name, and provider name according to a standard format.

And then what if the checkbox says "Override standard format" and defaults to unchecked?

charliepark:
Screenshot 2024-11-07 at 3 40 27 PM

app/forms/idp/create.tsx
```diff
@@ -62,6 +65,24 @@ export function CreateIdpSideModalForm() {
   })
 
   const form = useForm({ defaultValues })
+  const name = form.watch('name')
+
+  const acsUrlForm = useForm({ defaultValues: { generateUrl: true } })
```
david-crespo, line 70: This works, but it would be simpler as a regular `useState` and a checkbox instead of a CheckboxField

david-crespo (approved):
Awesome. Try the `useState` thing, see if it’s nicer, but this is good to go.

Issue #1564, opened 2023-06-09
Validate both-or-neither on signing keypair on IdP create


david-crespo:
In the IdP create UI, we have two file fields for the signing keypair. In the API, these are [optional _as a pair_](https://github.com/oxidecomputer/omicron/blob/051cd95975b83626d6f801c566404892549b1fbb/nexus/types/src/external_api/params.rs#L599-L602). If (as concluded in https://github.com/oxidecomputer/omicron/issues/3155 after some confusion on my part) we do want to have first-class support for the no-keypair option, I think the UI should be more explicit about how you need either both or neither. Maybe something that collapses both together, or a radio, or something. Even if we don't do that, we should validate both-or-neither client-side with a helpful message.

<img src="https://github.com/oxidecomputer/omicron/assets/3612203/cb723ec8-8acf-459d-aae7-6a0ce7b0a4f4" width="466">

PR #2511, opened 2024-10-21
Reorder IDP form elements

### LINKED ISSUE #2471: [proposal] Reorganize IdP form
Currently the IdP create form lists all of the form elements as a single column of form elements. We have seen multiple customers run in to confusion on exactly what each field is for, and I think we could partially improve this by reorganizing the form. Instead of a straight list of elements, we can group them based on object they are configuring (either the Service Provider (rack) or the IdP). As an initial idea I would propose:

**General**
* Name
* Description
* Technical Contact Email

**Service Provider**
* Service provider client ID
* ACS URL
* Single Logout (SLO) Url
* Signed Requests (maybe a check-box that enables the cert / key fields)
  * Public Cert
  * Private Key

**Identity Provider**
* Entity ID
* Group attribute name
* Metadata source

charliepark:
So far, this branch is only modifying the create form, though once feedback is in, I'll modify the update form to match.

This addresses #2471, where Augustus proposed updating the IDP form to more logically group the various components. The main changes here are …
- adding headers for the various sections
- reordering the fields
- adding a checkbox for signed requests, which then reveals the public cert / private key fields

Here's the updated layout …
Screenshot 2024-10-21 at 10 11 37 AM

… and with the "Signed requests" box checked …
Screenshot 2024-10-21 at 10 11 53 AM

I'm open to any other modifications; just opening this up for discussion and moving the ball down the field.

Closes #2471


benjaminleonard:
I'm trying to recall if we're using headings for any pane forms currently. It's a reasonable addition but there are broader implications in terms of rolling it out to every form that chunks the contents like this.

This might be a reason to move this into a full-page like the instance create form. What do you think? That might give us  space to add some more inventive UI also.

david-crespo:
Yeah, the firewall rules form probably feels better in the context of the list. It would be a little jarring on its own page. The main problem with the form is the length, so I'm not sure a full page form would help much anyway, though I guess you could do something like putting target type and value pickers side by side. I wonder if there's room for a wider side modal form.

charliepark:
Still hoping for some font-size guidance on that `1.5xl` size font (Ben is checking it out), but as this is closer, I'm going to take it out of draft mode.

charliepark:
Seems like adding a link to https://docs.oxide.computer/guides/system/completing-rack-config#_configure_silo_identity_provider might be helpful

david-crespo:
Looking great. I think we can figure out the headers later. If this went out in v12 with the big headers, I'd be ok with it.

This should be tweaked, though:

* I think the checkbox should be checked by default as nearly everyone in production will want to include certs (confirm @augustuswm?)
* `(Optional)` on the checkbox doesn't make sense, the checkbox is not optional, it is always either checked or unchecked

Really I'm thinking we can probably do without the checkbox since the fields will nearly always want to be used, and they're already optional anyway.

image

app/util/links.ts
```diff
@@ -19,6 +19,8 @@ export const links = {
   imagesDocs: 'https://docs.oxide.computer/guides/creating-and-sharing-images',
   preparingImagesDocs:
     'https://docs.oxide.computer/guides/creating-and-sharing-images#_preparing_images_for_import',
+  identityProvidersDocs:
+    'https://docs.oxide.computer/guides/system/completing-rack-config#_configure_silo_identity_provider',
```
david-crespo, line 23: Not suggesting a change here, just musing — the linked table is pretty good. It's a shame the generated doc for the endpoint is not nearly as helpful. It's kind of hard to see how it could be as good, since the table includes hand-written notes about, e.g., how the name will be used in context. I wonder if there's a way we could add our own notes in the API structs in addition to the default `Name` docs. Unfortunately these comes from another struct:

```rs
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
pub struct SamlIdentityProviderCreate {
    #[serde(flatten)]
    pub identity: IdentityMetadataCreateParams,
```

image

image

augustuswm:
I think it is actually the inverse. I would expect more installs to not use signed requests than those that would. I definitely don't think we have the intention of making it mandatory as I would not be surprised to encounter IdPs that do not support signed requests. The inverse (signed responses) is mandatory, but that is handled by the IdP metadata url / xml.

david-crespo:
Ah, ok! I was thinking of responses. If we want to encourage it, I think it’s ok to leave the checkbox out and just validate the fields together so they error if they’re empty while the other is not

david-crespo:
new headings look great, thanks @benjaminleonard 

image

david-crespo:
Followup on validating the keypair: https://github.com/oxidecomputer/console/issues/1564#issuecomment-2460336231. Current state is worse than I thought.

augustuswm:
Thank you for all the work on this. It looks so much better and I think a lot more streamlined.

david-crespo:
Good idea, will add.

PR #2512, opened 2024-10-21
Add realtime typing restrictions on NameField

charliepark:
Users are prevented — when submitting a create form — from using spaces, uppercase letters, or anything besides `[a-z0-9-]` in their `Name` fields.

This PR moves that restriction to the actual `NameField` itself, so when a user types in a space, it's entered in as a dash, when they enter an uppercase letter, it's added as a lowercase letter, and when they enter any non-allowed character, nothing is added to the field.


david-crespo (approved):
very nifty idea

PR #2514, opened 2024-10-22
Add descriptions to IP Pool dropdowns

### LINKED ISSUE #2041: Indicate to operator that IP pool description is visible to end users in linked silos
These descriptions could be aimed at end users to help them choose a pool.

### LINKED ISSUE #2042: Display IP pool description on IP pool pickers
Other side of #2041. Descriptions could help end users know what they're looking at.

charliepark:
This adds the IP Pool description to the IP Pool dropdown visible to the end user on the silo …
Screenshot 2024-10-22 at 3 51 33 PM
http://localhost:4000/projects/mock-project/floating-ips-new

And adds a note to the IP Pool form …
Screenshot 2024-10-22 at 3 53 32 PM
http://localhost:4000/system/networking/ip-pools-new

Closes #2041 
Closes #2042

### DIFF
--- a/app/forms/floating-ip-create.tsx
+++ b/app/forms/floating-ip-create.tsx
@@ -19,8 +19,8 @@ import {
 } from '@oxide/api'
 
 import { AccordionItem } from '~/components/AccordionItem'
+import { ComboboxField } from '~/components/form/fields/ComboboxField'
 import { DescriptionField } from '~/components/form/fields/DescriptionField'
-import { ListboxField } from '~/components/form/fields/ListboxField'
 import { NameField } from '~/components/form/fields/NameField'
 import { SideModalForm } from '~/components/form/SideModalForm'
 import { useProjectSelector } from '~/hooks/use-params'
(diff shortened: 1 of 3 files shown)


charliepark:
I added some vertical space to the combobox for the Select an Image dropdown as well:
Screenshot 2024-10-22 at 4 23 38 PM

app/forms/floating-ip-create.tsx
```diff
@@ -19,34 +19,40 @@ import {
...
+const toComboboxItem = (p: SiloIpPool): ComboboxItem => {
+  const value = p.name
+  const selectedLabel = p.name
+  const label = (
+    <div className="flex flex-col gap-1">
+      <div>
+        {p.name}
+        {p.isDefault && (
+          <>
+            {' '}
```
david-crespo, line 43: space might not be necessary if the `ml` is handling it

app/forms/ip-pool-create.tsx
```diff
@@ -51,6 +52,10 @@ export function CreateIpPoolSideModalForm() {
     >
       <NameField name="name" control={form.control} />
       <DescriptionField name="description" control={form.control} />
+      <Message
+        variant="info"
+        content="IP pool names and descriptions are visible to end users in linked silos."
+      />
```
david-crespo, line 58: I like this — since we have so much space and the form is so simple, we phave room to add a few more words about why or when users will see this. Also I think we can just call them "users" instead of end users. Something like this?

> IP pool names and descriptions are shown to users in linked silos in order to help them choose a pool while allocating IPs for instances.

charliepark:
Crespo and I talked IRL and he pointed out that — since there won't be a surfeit of pools — it probably makes sense for these to all be Listboxes, rather than Comboboxes. Still with the descriptions, though.

david-crespo:
The badge color needs a tweak when the item is selected.

### Fine when not highlighted



### Weird when highlighted

charliepark:
I'll work on something a bit more sophisticated via CSS, but this first stab could have been worse.
Screenshot 2024-10-28 at 3 45 13 PM
Screenshot 2024-10-28 at 3 47 18 PM
Screenshot 2024-10-28 at 3 47 28 PM
Screenshot 2024-10-28 at 3 47 37 PM

charliepark:
This wasn't actually bad at all.
badges

david-crespo (approved):
Awesome. I’ll leave it to @benjaminleonard to tweak the selected badge color if he wants.

benjaminleonard:
I've got it somewhere, not sure where of an example with the selected badge. But should be solid instead of dim.

Not sure if we have the accent solid variant as a class. But it might be easier to extract it out into its own class and then conditionally apply that instead? Then `badgeColors` is just applying that class.

david-crespo:
It's nice to do it in CSS, otherwise the item has to be aware of whether it's selected.

charliepark:
Here's where we landed in chat:
Screenshot 2024-10-29 at 10 55 43 AM
Screenshot 2024-10-29 at 10 55 50 AM

PR #2516, opened 2024-10-23
Trim descriptions

### LINKED ISSUE #2515: make sure descriptions can't consist of an empty string
It's possible, when creating an IP Pool, to add a description consisting of ` ` (a single space). We should run `trim()` on the description at create/update time, or possibly as a `transform` in the `onChange` for the field, to make sure than an empty description doesn't get passed through.

charliepark:
For discussion …

This PR fixes an edge case with DescriptionFields. Currently, they can be created with an empty space (or some other combination of whitespace) as the only text in the description. This PR updates the DescriptionField, so that when it is blurred, the value is trimmed.

I like this approach, as it keeps the change scoped to the DescriptionField component, whereas if we handled it at the form layer, we'd need to add a function in each place in the app where we use the DescriptionField. This way, it's handled within the one file. That being said, forcing an update to `props.control._formValues.description` feels off.

If we move forward with it …
Closes #2515

### DIFF
--- a/app/components/form/fields/DescriptionField.tsx
+++ b/app/components/form/fields/DescriptionField.tsx
@@ -16,7 +16,18 @@ export function DescriptionField<
   TFieldValues extends FieldValues,
   TName extends FieldPath<TFieldValues>,
 >(props: Omit<TextFieldProps<TFieldValues, TName>, 'validate'>) {
-  return <TextField as="textarea" validate={validateDescription} {...props} />
+  return (
+    <TextField
+      as="textarea"
+      validate={validateDescription}
+      onBlur={(event) => {
+        const trimmedDescription = event.target.value.trim()
+        event.target.value = trimmedDescription
+        props.control._formValues.description = trimmedDescription
+      }}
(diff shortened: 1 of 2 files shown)


benjaminleonard:
I imagine it's a bit more work, but can we perhaps address this (https://github.com/oxidecomputer/omicron/issues/4929) and set it to undefined if it's just whitespace.

app/components/form/fields/DescriptionField.tsx
```diff
@@ -16,7 +16,18 @@ export function DescriptionField<
...
-  return <TextField as="textarea" validate={validateDescription} {...props} />
+  return (
+    <TextField
+      as="textarea"
+      validate={validateDescription}
+      onBlur={(event) => {
+        const trimmedDescription = event.target.value.trim()
+        event.target.value = trimmedDescription
+        props.control._formValues.description = trimmedDescription
+      }}
```
david-crespo, line 27: I think blur is too specific of a place for this — shouldn't `transform` work?

PR #2518, opened 2024-10-23
Virtual list entry for arbitrary values in comboboxes

### LINKED ISSUE #2447: Workshop arbitrary values help text
> Select an option or enter a custom value

What we have is decent, but I just want to flag that we should keep mulling it over before v11 in case we come up with something better. It might help to explain somewhere (in this help text or elsewhere) that the match works by string, and therefore it can be any string, even one that doesn't match any instances _yet_. At least, that's what I want to convey — doesn't mean we have to say it exactly like that.

https://github.com/oxidecomputer/console/blob/ce242e7764709d201576a950c0539f78439c4578/app/forms/firewall-rules-common.tsx#L124-L129

### LINKED ISSUE #2507: Add virtual item to list arbitrary values in comboboxes
When the user is entering a value that is not in the list for a combobox that allows arbitrary values, we could make a sort of virtual entry in the list with some text that helps clarify what is going on. Inspired by a [Zed PR](https://github.com/zed-industries/zed/pull/18712).

Image

charliepark:
Update how the `allowArbitraryValue` functionality works within comboboxes.

Essentially, this adds a new line below the input that says "Use {typed-in text}". Technically, what they've entered into the input is already set as the value, so this is more of an affordance for the UX, so the user has more confidence that they're selecting that option.

arbitrary

It works both to hit `Enter` and to select the `Use db2` option.

I experimented with a "clear" button as well, but the UX for that wasn't really better than just deleting the input field's contents.
(Note: the following screenshot is NOT what is currently represented in this PR)
Screenshot 2024-10-23 at 4 30 55 PM

Closes #2507 
Closes #2447


charliepark:
I'm seeing some funny interactions with the text input, but am having trouble isolating the issue. Essentially, the text input doesn't always clear from the comboboxes when the "Add host filter" button is clicked. I don't think this has to do with the combobox, but will be worth sorting out.

app/ui/lib/Combobox.tsx
```diff
@@ -177,13 +177,21 @@ export const Combobox = ({
...
-        {items.length > 0 && (
+        {(items.length > 0 || allowArbitraryValues) && (
           <ComboboxOptions
             anchor="bottom start"
             // 14px gap is presumably because it's measured from inside the outline or something
-            className={`ox-menu pointer-events-auto ${zIndex} relative w-[calc(var(--input-width)+var(--button-width))] overflow-y-auto border !outline-none border-secondary [--anchor-gap:14px] empty:hidden`}
+            className={`ox-menu pointer-events-auto ${zIndex} relative w-[calc(var(--input-width)+var(--button-width))] overflow-y-auto border !outline-none border-secondary [--anchor-gap:13px] empty:hidden`}
             modal={false}
           >
+            {allowArbitraryValues && filteredItems.length === 0 && (
```
david-crespo, line 187: Maybe it should show up even when there are matches as long as it doesn't exactly match any of the matches?

charliepark: Ah, so if `tacocat` is an existing option in the list, and they've typed in `taco`, we should show `tacocat` as well as `Use taco`. Sounds good. Will sort that out.

david-crespo:
I'm leaning back toward having a special treatment for the virtual entry, like `Use` or even `Use custom value: ` ("custom" is tolerable but might not be the best word). The problem is kinda there all the time, but comes out most when you type in a letter that's not in any of the entries, so the list goes from having N real entries to having one virtual one. Kind of jarring.

https://github.com/user-attachments/assets/115a23a2-88a7-4fd6-8ba3-bb44c08c16e1

david-crespo:
Looking good. It would be nice if we could fade the `Use custom value`

image

like in this screenshot which you said is of something other than this PR.

image

I'm experimenting. The label is a `ReactNode` already so it might be pretty easy.

david-crespo:
Ugh, the [test failure](https://github.com/oxidecomputer/console/actions/runs/11711075138/job/32618829894#step:9:101) is real. It's caused by my attempt to fix the blur and refocus thing. The fact that we're not clearing the query on blur means it is also not cleared when we change host filter types.

https://github.com/oxidecomputer/console/blob/45b8585a2ba94b3226fefb23b9d6228baf22ed02/app/ui/lib/Combobox.tsx#L126-L127

david-crespo:
Found another bug that is probably also a consequence of getting rid of the `onClose`:

1. type in custom value
2. press enter, or blur, or click the item — any works
3. click add target
4. focus field

`def` is still there! This only happens with VPC type because the others revert back to VPC on submit, which clears the field because of the `key` added in a79be2e66ccfb05971acbafb0af21d537963953c.

https://github.com/user-attachments/assets/a2de76ec-f206-451c-9f7f-61764edfd96c

charliepark:
Looks really good, @david-crespo. I tried breaking it in a few different ways that seemed to occasionally break it earlier, and all seem to be patched.

PR #2520, opened 2024-10-28
Add Group Attribute Name field to IdP view

### LINKED ISSUE #2473: Display group_attribute_name in IdP view
Currently we do not display the `group_attribute_name` field from an IdP configuration in the view IdP flyout. This is a helpful value to display when debugging connection issues.

charliepark:
This field was already present, just hidden in the "edit" (though not really editable) view.

I also adjusted the help text's CSS to have `text-balance`, so we avoid having ragged descriptions.
Screenshot 2024-10-28 at 11 59 26 AM

Closes #2473

### DIFF
--- a/app/forms/idp/edit.tsx
+++ b/app/forms/idp/edit.tsx
@@ -104,14 +104,13 @@ export function EditIdpSideModalForm() {
         control={form.control}
         disabled
       />
-      {/* TODO: add group attribute name when it is added to the API
-          <TextField
-            name="groupAttributeName"
-            label="Group attribute name"
-            description="Name of SAML attribute where we can find a comma-separated list of names of groups the user belongs to"
-            control={form.control}
-            disabled
-          /> */}
+      <TextField
+        name="groupAttributeName"
+        label="Group attribute name"
+        description="Name of the SAML attribute sent by the IdP containing a comma-separated list of groups for the authenticated user"
+        control={form.control}
+        disabled
+      />
       {/* TODO: Email field, probably */}
       <TextField
         name="technicalContactEmail"
--- a/app/forms/idp/edit.tsx
+++ b/app/forms/idp/edit.tsx
@@ -107,7 +107,7 @@ export function EditIdpSideModalForm() {
       <TextField
         name="groupAttributeName"
         label="Group attribute name"
-        description="Name of the SAML attribute sent by the IdP containing a comma-separated list of groups for the authenticated user"
+        description="Name of the SAML attribute in the IdP response listing the user’s groups"
         required
         control={form.control}
         disabled
--- a/app/forms/idp/edit.tsx
+++ b/app/forms/idp/edit.tsx
@@ -108,6 +108,7 @@ export function EditIdpSideModalForm() {
         name="groupAttributeName"
         label="Group attribute name"
         description="Name of the SAML attribute sent by the IdP containing a comma-separated list of groups for the authenticated user"
+        required
         control={form.control}
         disabled
       />
--- a/app/forms/idp/create.tsx
+++ b/app/forms/idp/create.tsx
@@ -133,7 +133,7 @@ export function CreateIdpSideModalForm() {
       <TextField
         name="groupAttributeName"
         label="Group attribute name"
-        description="Name of SAML attribute where we can find a comma-separated list of names of groups the user belongs to"
+        description="Name of the SAML attribute in the IdP response listing the user’s groups"
         control={form.control}
       />
       {/* TODO: Email field, probably */}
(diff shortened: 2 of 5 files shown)


app/forms/idp/edit.tsx
```diff
@@ -104,14 +104,14 @@ export function EditIdpSideModalForm() {
...
-            name="groupAttributeName"
-            label="Group attribute name"
-            description="Name of SAML attribute where we can find a comma-separated list of names of groups the user belongs to"
-            control={form.control}
-            disabled
-          /> */}
+      <TextField
+        name="groupAttributeName"
+        label="Group attribute name"
+        description="Name of the SAML attribute sent by the IdP containing a comma-separated list of groups for the authenticated user"
```
david-crespo, line 110: I don't think they're actually comma-separated. It's all XML.

```xml
<saml:AttributeStatement>
  <saml:Attribute Name="groups" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">
    <saml:AttributeValue>SRE</saml:AttributeValue>
    <saml:AttributeValue>Admins</saml:AttributeValue>
  </saml:Attribute>
</saml:AttributeStatement>
```

https://github.com/oxidecomputer/omicron/blob/5bf05de207bf05201aea7c07f24ddc7724707008/nexus/tests/integration_tests/data/saml_response_with_groups.xml#L94-L98

Maybe we can be a little less concrete about the format anyway. It doesn't matter here:

> Name of the SAML attribute in the IdP response listing the user's groups

We could go even shorter to get it on one line, but I don't think this is consistent with our more typically breezy style.

> Name of SAML attribute in IdP response listing user's groups

Worth noting that `text-balance` looks worse and worse the shorter the thing gets. I'd take it back out.

The docs [also](https://docs.oxide.computer/guides/system/completing-rack-config#_configure_silo_identity_provider) say "comma-separated". I will make a note to confirm it's wrong and fix it.

image

david-crespo:
The create form has a field for this too — its description should match. Yet another case where it would be convenient to share the fields (and in this case it would have helped catch this mismatch) but it feels kinda boilerplatey to get the types right on shared stuff. We should think about how to make it easier, or if we can't, maybe at least commit to a particular pattern here for sharing fields and document it so we don't have to think about it. Out of scope for this PR of course.

https://github.com/oxidecomputer/console/blob/4e967560d1a090ea18b6b8127eb9763fde555987/app/forms/idp/create.tsx#L133-L138

app/forms/idp/edit.tsx
david-crespo: Docs fix: https://github.com/oxidecomputer/docs/pull/484

david-crespo (approved):
Not a blocker, but I'd recommend adding `groupAttributeName` to the mock SAML IdP so we can see it in the form.

https://github.com/oxidecomputer/console/blob/4e967560d1a090ea18b6b8127eb9763fde555987/mock-api/silo.ts#L76-L88

Might also be worth adding a line to this expect asserting that it shows up. Don't use the `text=` way, do the `expect(page.getByRole('textbox', { name: 'Group attribute name' })).toHaveValue('groups')` way.

https://github.com/oxidecomputer/console/blob/4e967560d1a090ea18b6b8127eb9763fde555987/test/e2e/silos.e2e.ts#L181-L187

app/forms/idp/edit.tsx
jmpesp: It can be a single AttributeStatement, like you posted, but it can also be multple:

```xml
      <saml:Attribute Name="groups" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">
        <saml:AttributeValue>Admins</saml:AttributeValue>
      </saml:Attribute>
      <saml:Attribute Name="groups" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">
        <saml:AttributeValue>SRE</saml:AttributeValue>
      </saml:Attribute>
```

the Keycloak setting for this is (under the "groups" mapping):

keycloak_single_group_attribute

It _can_ also be a comma separated list of groups, if the IDP sends that. Some applications apparently do require that (eg: https://support.okta.com/help/s/question/0D54z00009Jau72CAB/group-attribute-statements-combine-all-groups-into-a-comma-separated-string?language=en_US)

david-crespo: Very helpful, thanks. Do all of those work the same way with respect to the attribute name? In which case just not mentioning the details seems best.

jmpesp: > Do all of those work the same way with respect to the attribute name? In which case just not mentioning the details seems best.

I believe so yeah, both use the same `Attribute Name="groups"`, and it'd be a bug not to. Agree on eliding that detail yeah. Even so if we ever encounter an IDP that only sends comma separated lists Nexus can handle it.

PR #2522, opened 2024-10-28
Fix dropdown z index

### LINKED ISSUE #2521: Table menu button dropdown goes behind bottom bar
When clicking the three dot 'kebab' menu button on a table, if the button is high enough on the page that it opens downward, but low enough that it would meet the bottom 'prev / next' bar, it will go behind the bar.

Image

charliepark:
Per #2521, action menu dropdowns were stacking behind the footer bar, which has a z-index of `20`. 
image

This PR updates the imported classes for `.DropdownMenuContent` to use a z-index of `25`. We also had a few overrides of the default z-index in TopBar and TopBarPicker, but since those are now using the same z-index, we can remove the overrides.

Screenshot 2024-10-28 at 2 35 21 PM

I have clicked through as many scenarios as I can think of regarding popovers and modals, and all seem to be working as expected.

Closes #2521.

### DIFF
--- a/app/ui/styles/components/menu-button.css
+++ b/app/ui/styles/components/menu-button.css
@@ -7,7 +7,7 @@
  */
 
 .DropdownMenuContent {
-  @apply z-popover min-w-36 rounded border p-0 bg-raise border-secondary;
+  @apply z-topBarDropdown min-w-36 rounded border p-0 bg-raise border-secondary;
 
   & .DropdownMenuItem {
     @apply block w-full cursor-pointer select-none border-b py-2 pl-3 pr-6 text-left text-sans-md text-secondary border-secondary last:border-b-0;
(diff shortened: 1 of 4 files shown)


app/ui/styles/components/menu-button.css
```diff
@@ -7,7 +7,7 @@
  */
 
 .DropdownMenuContent {
-  @apply z-popover min-w-36 rounded border p-0 bg-raise border-secondary;
+  @apply z-topBarDropdown min-w-36 rounded border p-0 bg-raise border-secondary;
```
david-crespo, line 10:
```suggestion
  /* we want menu popover to be on top of top bar and pagination bar too */
  @apply z-topBarDropdown min-w-36 rounded border p-0 bg-raise border-secondary;
```

PR #2529, opened 2024-10-30
Bring back topbar breadcrumbs

### LINKED ISSUE #2509: Make breadcrumbs more recognisable as breadcrumbs
Current implementation, whilst essentially the same does not always obviously read as breadcrumbs. Let's switch to a less avant garde design.

Before:
Image

After:
Image

charliepark:
This PR switches from the current topbar navigation — using dropdowns that are a little too clever — and moves to a more standard breadcrumb navigation.

Screenshot 2024-10-30 at 10 41 01 AM

Screenshot 2024-10-30 at 10 41 13 AM

I expect there'll be a few tweaks to make re: spacing / sizing, and making the element responsive / more intelligent about narrow viewports.

Closes #2509


david-crespo:
I think this looks great. Huge improvement. A couple of issues to work out:

### No project detail page

This has always been a little weird, but this brings it out. Because there is nothing interesting about a project besides its name and description, we have always used the instance list as a project homepage. But that means both the project link and the instances link go to the same thing. There are some possible workarounds, but it might be easiest to add a project detail page. On the other hand such a page would probably not be that useful.

https://github.com/user-attachments/assets/16d018cc-8f70-4c52-a40c-625b75bbec92


### Back arrow when there's nothing to go back to

Just a little weird, but I'm not sure what else we would do. Also it's slightly weird for `Projects` to be the only white thing on the page.

image

It's less weird when there are other elements in the breadcrumbs because the point is the contrast, though maybe still slightly weird when you think about it. Though as you can see here we have a couple of other white elements here and there, like the section headings. Doesn't feel like there's a clear principle to it.

image

charliepark:
Have removed the arrow from single-level root pages, and also set the higher contrast link to only show up when there are multiple levels.
Screenshot 2024-10-30 at 3 16 54 PM
Screenshot 2024-10-30 at 3 17 01 PM

benjaminleonard:
Addressing some of the visual things now. As an aside, was looking into how a clickable back button might work...tricky and still not quite right.

```tsx
  const matches = useMatches()
  // Finds the next available route to jump to
  // If there's no handle we have no breadcrumb
  // and skip: true means we skip that route
  const parentRoute =
    matches
      .filter(
        (match) => !!match.handle || (match.handle && !(match.handle as RouteHandle).skip)
      )
      .map((match) => match.pathname)
      .slice(0, -1)[0] || ''
```

You then add a skip to routes you don't want to go back to – in this case we'd skip over the instance page since we're on a tab route.

```tsx
<Route
  element={<InstancePage />}
  loader={InstancePage.loader}
  handle={{ skip: true }}
>
```

I can get the parent route mostly okay, but the redirects make it a little funky. Must be a better way.

app/components/TopBar.tsx
```diff
@@ -38,7 +38,7 @@ export function TopBar({ children }: { children: React.ReactNode }) {
       {/* Height is governed by PageContainer grid */}
       {/* shrink-0 is needed to prevent getting squished by body content */}
       <div className="z-topBar border-b bg-default border-secondary">
-        <div className="mx-3 flex h-[60px] shrink-0 items-center justify-between">
+        <div className="mx-3 flex h-[54px] shrink-0 items-center justify-between">
```
benjaminleonard, line 41: Was going to ask if it was worth making this a var, but that might only be worthwhile if we need to use it in more places.

app/components/TopBarBreadcrumbs.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+export const TopBarBreadcrumbs = () => {
+  const [, firstPathItem, secondPathItem, thirdPathItem, , fifthPathItem] =
+    window.location.pathname.split('/')
+
+  // console.log(paths, paths.slice(0, -1).join('/'))
+  const { project } = useParams()
+  // if there's no secondPathItem on the silo section, there's no page to go "back" to
+  // the secondPathItem is top-level within the system section and therefore we check for thirdPathItem
+  const isTopLevel = (firstPathItem === 'system' && !thirdPathItem) || !secondPathItem
```
benjaminleonard, line 23: Might be a better way to do this...

app/components/TopBarBreadcrumbs.tsx
```diff
@@ -0,0 +1,185 @@
...
+  return (
+    <nav
+      className="flex items-center gap-0.5 overflow-clip pr-4 text-sans-md"
+      aria-label="Breadcrumb navigation"
+    >
+      <PrevArrow12Icon
+        className={cn('mx-1.5 flex-shrink-0 text-quinary', isTopLevel && 'opacity-40')}
```
benjaminleonard, line 30: Went for opacity to fade out the prev button when it's no longer applicable instead.

benjaminleonard:
On the high contrast thing – I've set it to `text-secondary` for now. Note that when I make the `text-default` token changes it'll be in more places and we can use it here without it looking unusual.

david-crespo:
This version also has a problem with routes nested under route tabs I noted in my refactor: https://github.com/oxidecomputer/console/pull/2531#issuecomment-2452269208

https://github.com/user-attachments/assets/223e09f0-b3fc-408c-9f65-ae388b637b5d

app/components/TopBar.tsx
david-crespo: WELP

https://github.com/oxidecomputer/console/blob/b4920b1157c5f63a8d5a39adcdb62c541ca870ec/app/pages/project/instances/instance/SerialConsolePage.tsx#L138

app/layouts/ProjectLayout.tsx
```diff
@@ -90,10 +79,7 @@ export function ProjectLayout({ overrideContentPane }: ProjectLayoutProps) {
     <PageContainer>
       <TopBar>
         <SiloSystemPicker value="silo" />
-        <ProjectPicker project={project} />
-        {instance && <InstancePicker />}
-        {vpc && <VpcPicker />}
-        {router && <VpcRouterPicker />}
+        <Breadcrumbs />
```
david-crespo, line 82: All the `TopBar` calls look nearly the same now, so we can eventually move all this into `TopBar`, maybe give it a prop that's `'system' | 'silo'` prop. Doesn't have to be in this PR.

app/components/TopBar.tsx
charliepark: womp

test/e2e/instance-serial.e2e.ts
```diff
@@ -19,7 +19,7 @@ test('serial console can connect while starting', async ({ page }) => {
   // now go starting to its serial console page while it's starting up
   await expect(page).toHaveURL('/projects/mock-project/instances/abc/storage')
   await page.getByRole('tab', { name: 'Connect' }).click()
-  await page.getByRole('link', { name: 'Connect' }).click()
+  await page.getByRole('link', { name: 'Connect' }).nth(1).click()
```
david-crespo, line 22: This is fine, but another approach that might work is use `main` or the tab content pane to select a container and then the connect link.

app/components/TopBar.tsx
charliepark: I dropped this down from 60 to 54, but this is technically for delineating between the body and the footer. @benjaminleonard do we want the serial console footer to match the header, at 54px? Or to stay at 60?

benjaminleonard: I think in that case it's making the layout the full height of the screen, minus the size of the top bar. The bottom bar on the serial console seems a bit chunky but it's matching the instance create bottom bar. Might adjust elsewhere.

benjaminleonard:
I think we might have time to make the system / silo picker change integrated into this before v12 is released. Do you want to merge this and follow up with that in a separate PR?

david-crespo:
I think we should do it in a followup because this is a net improvement on its own and we do have plenty of time.

As I said in chat: I think for this to be mergeable, we need to take out the `crumb` handles for the side modal form routes, or mark them “title only” or something so they don't show up as UI breadcrumbs. Even if we keep them as crumbs, we need to go through and make sure we’re treating them consistently.
Might be cool to write an e2e test that tests the breadcrumbs on approximately every page.

david-crespo:
Here is me demoing approximately all the crumbs. They look pretty good except for the ones where there's a crumb that doesn't actually link to a page, as @charliepark pointed out. (I also noticed issues on SSH key and floating IP pages, fixed in dfb314399ebf9304f3e2e99b6cc0713090cee449).


https://github.com/user-attachments/assets/b59e084f-c651-468e-a9a0-966807cd7b9a



And here is me having some fun with the top bar height CSS var.

https://github.com/user-attachments/assets/a385021d-8113-4e05-9931-50de6fd6ddc4

david-crespo:
I experimented with a `link: false` annotation for the `<Route>`s that don't have their own pages, but it felt weird and bad not to be able to click them. I'm just going to try and make sure they at least don't flash weird blank pages on click.

david-crespo:
One approach that would work for avoiding blank screen flashes on redirecting nodes is hacking in a way for the crumb for `/projects/mock-project` to actually point to `/projects/mock-projects/instances`. That would likely mean annotating the route handle with another field `crumbPath` that can override `match.pathname` as the link path for the crumb. This is pretty straightforward to do.

A more evil approach is to make the loaders for the `/projects/mock-project` route node match the loaders for `/projects/mock-projects/instances` so they are loaded before we hit `/projects/mock-project` and therefore we don't see a flash because the redirect is instantaneous. This works in my testing, but it's much uglier and more error-prone because it can easily become wrong if the loaders for the target route change.

```diff
-      <Route path="projects/:project" element={<Navigate to="instances" replace />} />
+      <Route
+        path="projects/:project"
+        loader={(args) =>
+          Promise.all([ProjectLayout.loader(args), InstancesPage.loader(args)])
+        }
+        element={<Navigate to="instances" replace />}
+      />
```

app/util/path-builder.ts
```diff
@@ -119,8 +115,6 @@ export const pb = {
   ipPoolEdit: (params: IpPool) => `${pb.ipPool(params)}/edit`,
   ipPoolRangeAdd: (params: IpPool) => `${pb.ipPool(params)}/ranges-add`,
 
-  inventory: () => '/system/inventory',
-  rackInventory: () => '/system/inventory/racks',
```
david-crespo, line 123: The deleted ones in here were never used. Some were used to construct other paths but never used directly.

charliepark:
This is all looking great.

PR #2530, opened 2024-10-31
Tell password managers to stop trying to fill `name` field

david-crespo:
This 1Password thing on `name` fields in our forms is so silly.

image

I've noticed this forever but didn't realize how easy it would be to fix. 

https://www.stefanjudis.com/snippets/turn-off-password-managers/


charliepark (approved):
nice

PR #2531, opened 2024-11-01
Breadcrumbs powered by `useMatches()`

david-crespo:
This more or less works already, but needs some tweaks to the route config to make it produce the correct results. The route config changes are simple, but they change the indentation, so the [whitespace-hidden diff](https://github.com/oxidecomputer/console/pull/2531/files?w=1) is much better.

image

- [x] Basic `useMatches()` thing works
- [x] Add Projects crumb
- [ ] Add a mechanism for hiding crumbs for tab routes, or just remove the crumbs from the tabs


david-crespo:
This is pretty good, though there are still a few weird things, mostly around route tabs. Also edit and create forms. I think we could do without crumbs for those. I just didn't want to change the route config too much yet.

https://github.com/user-attachments/assets/12475fa6-b8b9-41e6-a4d1-28660f99e147

Route tabs already had crumbs defined so they would show up in the page title. We can easily remove them, and I tried that, but then you end up with this weirdness with nested ones, where you can click into the router detail, where you clearly want the crumb, and then when you click back up, the crumb disappears:

https://github.com/user-attachments/assets/e1694af5-7c9f-4555-8198-183f2284b2e5

charliepark (approved):
I'm having trouble getting this to run locally, but the code looks good.

Edit: Seems like the tests are missing some `Projects` in the expected strings. I can add those in.

PR #2533, opened 2024-11-06
Add side modal heading

benjaminleonard:
https://console-git-side-modal-title-oxidecomputer.vercel.app/projects/mock-project/vpcs/mock-vpc/firewall-rules-new

---

Added `SideModal.Heading` – example usage on Firewall Rules side modal to be used also within #2511.

image

The size looks good to me, making it semi seems to have made the difference in distinguishing it.

Also in terms of usage – I think a heading should always be preceded by a divider, things are going to get messy if we drop it in to group inputs willy nilly.

### DIFF
--- a/app/forms/firewall-rules-common.tsx
+++ b/app/forms/firewall-rules-common.tsx
@@ -425,8 +425,8 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
           <>
             <p>
               Targets determine the instances to which this rule applies. You can target
-              instances directly by name, or specify a VPC, VPC subnet, IP, or IP subnet,
-              which will apply the rule to traffic going to all matching instances.
+              instances directly or specify a VPC, VPC subnet, IP, or IP subnet, which will
+              apply the rule to traffic going to all matching instances.
             </p>
             <p className="mt-2">
               Targets are additive: the rule applies to instances matching{' '}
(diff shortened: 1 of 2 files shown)


app/forms/firewall-rules-common.tsx
```diff
@@ -523,45 +529,48 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
...
+        <FieldLabel id="portRange-label" htmlFor="portRange" className="mb-2">
+          Protocol filters
+        </FieldLabel>
+        {/* <legend className="mb-2 mt-4 text-sans-lg">Protocol filters</legend> */}
         <ProtocolField control={control} protocol="TCP" />
         <ProtocolField control={control} protocol="UDP" />
         <ProtocolField control={control} protocol="ICMP" />
       </fieldset>
 
-      <div className="flex flex-col gap-3">
```
benjaminleonard, line 532: This stuff is unnecessary because the overall thing already has the spacing, it also causes a bit of havok since the spacing here is different.

app/forms/firewall-rules-common.tsx
```diff
@@ -415,46 +417,50 @@ export const CommonFields = ({ control, nameTaken, error }: CommonFieldsProps) =
...
-          variant="info"
-          content={
-            <>
+      <SideModal.Heading>Targets</SideModal.Heading>
+
+      <Message
+        variant="info"
+        content={
+          <>
+            <p>
```
david-crespo, line 426: needs spacing

image

PR #2536, opened 2024-11-07
Fix radix warnings on side modal

david-crespo:
In local dev, Radix was giving a spurious error and a technically true but pointless warning on our side modal. This PR gets rid of both.

image

### DIFF
--- a/app/ui/lib/SideModal.tsx
+++ b/app/ui/lib/SideModal.tsx
@@ -30,7 +30,7 @@ export function usePopoverZIndex() {
 }
 
 export type SideModalProps = {
-  title?: string
+  title: string
   subtitle?: ReactNode
   onDismiss: () => void
   isOpen: boolean
@@ -53,7 +53,6 @@ export function SideModal({
   animate = true,
   errors,
 }: SideModalProps) {
-  const titleId = 'side-modal-title'
   const AnimatedDialogContent = animated(Dialog.Content)
 
   const config = { tension: 650, mass: 0.125 }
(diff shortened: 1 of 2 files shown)


app/ui/lib/SideModal.tsx
```diff
@@ -81,40 +80,39 @@ export function SideModal({
...
-                                  </>
-                                )
-                              }
-                              title={errors.length > 1 ? 'Errors' : 'Error'}
-                            />
-                          </div>
-                        )}
-                      </>
+                  <div className="items-top mb-4 mt-8">
+                    <Dialog.Title className="flex w-full items-center justify-between break-words pr-8 text-sans-2xl">
```
david-crespo, line 90: `Dialog.Title` without `asChild` renders as an `h2`. Putting it directly around the title itself makes a lot more sense here.

app/ui/lib/SideModal.tsx
```diff
@@ -30,7 +30,7 @@ export function usePopoverZIndex() {
 }
 
 export type SideModalProps = {
-  title?: string
+  title: string
```
david-crespo, line 33: Title did not need to be optional. We always have a title.

app/ui/lib/SideModal.tsx
```diff
@@ -53,7 +53,6 @@ export function SideModal({
   animate = true,
   errors,
 }: SideModalProps) {
-  const titleId = 'side-modal-title'
```
david-crespo, line 56: This is only need to hook up the accessible title to the dialog, but Radix takes care of that for us.

image

app/ui/lib/SideModal.tsx
```diff
@@ -81,40 +80,39 @@ export function SideModal({
...
-                                        <li key={idx}>{error}</li>
-                                      ))}
-                                    </ul>
-                                  </>
-                                )
-                              }
-                              title={errors.length > 1 ? 'Errors' : 'Error'}
-                            />
-                          </div>
-                        )}
```
david-crespo, line 115: why were errors inside the title????

app/ui/lib/SideModal.tsx
```diff
@@ -81,40 +80,39 @@ export function SideModal({
...
+                                {errors.map((error, idx) => (
+                                  <li key={idx}>{error}</li>
+                                ))}
+                              </ul>
+                            </>
+                          )
+                        }
+                        title={errors.length > 1 ? 'Errors' : 'Error'}
+                      />
+                    </div>
```
david-crespo, line 115: I confirmed everything looks the same as `main` when there are errors.

image

PR #2537, opened 2024-11-07
Add help text + request signing section header to IdP form

david-crespo:
Following up on https://github.com/oxidecomputer/console/pull/2511#issuecomment-2460418920. Let's be sooooooo helpful.

image

image

### DIFF
--- a/app/forms/idp/create.tsx
+++ b/app/forms/idp/create.tsx
@@ -156,7 +156,6 @@ export function CreateIdpSideModalForm() {
             <code>
               /login/{silo}/saml/{name.trim() || 'idp-name'}
             </code>
-            .
           </>
         }
       />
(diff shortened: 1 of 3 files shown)


david-crespo:
Moved on top of #2510 so we can get extra snazzy.

image

app/forms/idp/create.tsx
```diff
@@ -120,7 +122,36 @@ export function CreateIdpSideModalForm() {
...
+        name="name"
+        control={form.control}
+        description={
+          <>
+            A short name for the provider in our system. Users will see it in the path to
+            the login page:{' '}
+            <code>
+              /login/{silo}/saml/{name.trim() || 'idp-name'}
+            </code>
+            .
```
charliepark, line 151: If we're ending the copy with `<code>`, I think it looks better (and clearer) without the ultimate period. Thoughts about leaving that off?

david-crespo: I have it in there because it’s two sentences. Otherwise I would leave it out.

david-crespo: Hm, yeah I think you’re right. I’ll take it out.

PR #2540, opened 2024-11-08
Fix broken UX for the Enter key in comboboxes

### LINKED ISSUE #2534: Enter key in comboboxes in firewall rules submits form
The conversion of the firewall rule targets and hosts to comboboxes regressed https://github.com/oxidecomputer/console/pull/1928, which made it so pressing enter in the "subform" would only submit the subform, i.e., add the target or host to the list, and not submit the form. I noticed while working on https://github.com/oxidecomputer/console/pull/2518, but it was true before that.

charliepark:
This fixes a regression introduced in earlier combobox work, where hitting the Enter key in a subform (like the Firewall Rules create form) would try to submit the entire form, rather than just work within the smaller subform.

The current UX in this PR uses a ref and — upon the user hitting enter — moves the focus to the "add" button, which the user can use to add that selection to the subform. I played with having enter just submit the subform, but it felt a little surprising — a little "fast" — to have enter both select the item from the dropdown and add it to the subform in one fell swoop.

In adding the ref, I saw that it would be more efficient to move the relevant `MiniTable.ClearAndAddButtons` into the `DynamicTypeAndValueFields` (now re-titled as `TargetAndHostFilterSubform`), so I moved a bunch of duplicated code for **targets** and **host filters** into that subform.

Closes #2534

### DIFF
--- a/app/forms/firewall-rules-common.tsx
+++ b/app/forms/firewall-rules-common.tsx
@@ -109,11 +109,8 @@ const TargetAndHostFilterSubform = ({
   const field = useController({ name: `${sectionType}s`, control }).field
 
   const submitSubform = subform.handleSubmit(({ type, value }) => {
-    // TODO: do this with a normal validation
-    // ignore click if empty or a duplicate
-    // TODO: show error instead of ignoring click
     if (!type || !value) return
-    if (field.value.some((f) => f.value === value && f.type === type)) return
+    if (field.value.some((f) => f.type === type && f.value === value)) return
     field.onChange([...field.value, { type, value }])
     subform.reset(targetAndHostDefaultValues)
   })


david-crespo:
Nice refactor.

> it felt a little surprising — a little "fast" — to have enter both select the item from the dropdown and add it to the subform in one fell swoop.

I agree with this: it shouldn't select and submit in one keypress, but I think we can accomplish that without focusing the add button. What I pictured is: when the combobox options are open, enter selects the highlighted one without submitting the subform. At this point, focus is still in the input, but with the options popover closed. Pressing enter a second time will submit the subform.

Might be kind of annoying to implement, though — you have to figure out how to handle enter differently depending on whether the options are open or not.

app/forms/firewall-rules-common.tsx
```diff
@@ -89,31 +84,80 @@ const getFilterValueProps = (targetOrHostType: TargetAndHostFilterType) => {
...
+  } = usePrefetchedApiQuery('instanceList', { query: { project, limit: ALL_ISH } })
+  const {
+    data: { items: vpcs },
+  } = usePrefetchedApiQuery('vpcList', { query: { project, limit: ALL_ISH } })
+  const {
+    data: { items: vpcSubnets },
+  } = usePrefetchedApiQuery('vpcSubnetList', { query: { project, vpc } })
+
+  const subform = useForm({ defaultValues: targetAndHostDefaultValues })
+  const field = useController({ name: `${sectionType}s`, control }).field
```
david-crespo, line 109: Completely nuts that TS can do this.

charliepark:
I recommend hiding whitespace changes if looking at the code after the last commit: https://github.com/oxidecomputer/console/pull/2540/files?diff=split&w=1

app/ui/lib/Combobox.tsx
```diff
@@ -149,107 +151,112 @@ export const Combobox = ({
...
+                      // it would be a lot more verbose. those can only be used with TW classes,
+                      // not our .is-selected and .is-highlighted, so we'd have to copy the pieces
+                      // of those rules one by one. Better to rely on the shared classes.
+                      <div
+                        className={cn('ox-menu-item', {
+                          'is-selected': selected && query !== item.value,
+                          'is-highlighted': focus,
+                        })}
+                      >
+                        {item.label} {open}
```
david-crespo, line 247: probably don't to leave this in

charliepark: Good catch

app/ui/lib/Combobox.tsx
```diff
@@ -136,6 +143,17 @@ export const Combobox = ({
...
+    // Prevent form submission when the user presses Enter inside a combobox.
+    // The combobox component already handles Enter keypresses to select items,
+    // so we only preventDefault when the combobox is closed.
+    if (e.key === 'Enter' && !open) {
+      e.preventDefault()
+      onEnter?.(e)
+    }
+    onKeyDown?.(e)
+  }
+
```
david-crespo, line 156: Beautiful. If you inline this, you don't have to pass `open` because it's closed over. I think we could even do without the `onKeyDown` prop on `Combobox` if we're not using it.

david-crespo (approved):
Woohoo!

PR #2541, opened 2024-11-08
Prevent creating multiple disks with the same name

### LINKED ISSUE #2519: Disallow duplicate disk names on instance create form
Reported by @leftwo. Alan noticed that we don't prevent you from adding two disks to create with the same name, but the API doesn't like it, so you get an error on submit. We should prevent duplicates from being added.

It's even worse than Alan said — something very odd happens when I have a create and an attach with the same name and try to delete it. I have a video but it looks like I'll have to upload it later when I'm off airplane wi-fi.

charliepark:
Currently, it's possible to "create" multiple new disks with the same name in the instance create form. This is a little goofy, because the user isn't actually creating the new disks at this point in the form, they're queuing them up to be created shortly. So we have an issue where multiple matching domain names are problematic, and deleting one of those disks will end up deleting both disks.

This PR fixes that by adding validation check to the name field for disk-create, as well as a validation on the boot disk name in the instance-create form. It was already the case that an existing disk name couldn't be used for the boot disk name, but we were not checking the names of the liminal disks (set to be created, but not yet actually created). We are now.

Closes #2519


charliepark:
see-it-works

app/forms/disk-create.tsx
```diff
@@ -64,12 +64,14 @@ type CreateSideModalFormProps = {
    */
   onDismiss: (navigate: NavigateFunction) => void
   onSuccess?: (disk: Disk) => void
+  unavailableDiskNames?: string[]
```
david-crespo, line 67: I do like this a lot better without adding disks from the API in here. I'd add a comment here on the prop saying this is for client-side use-cases like the form embedded in instance create, where submitting the form does not tell you whether it's taken.

app/forms/instance-create.tsx
```diff
@@ -278,10 +284,18 @@ export function CreateInstanceForm() {
         required={false}
         control={control}
         disabled={isSubmitting}
+        validate={(name) => {
+          // don't allow the user to use an existing disk name for the boot disk's name
+          if ([...namesOfDisksToBeCreated, ...existingDiskNames].includes(name)) {
```
david-crespo, line 289: Very small thing, but I think it would be easier to follow if this was done next to the other disk name list logic. You can even wrap it in a function to avoid mapping over a potentially large list of disks on every render.

```ts
const otherDisks = useWatch({ control, name: 'otherDisks' })

function unavailableDiskNames() {
  const existingDiskNames = allDisks.map((disk) => disk.name)
  const namesOfDisksToBeCreated = otherDisks
    .filter((disk) => disk.type === 'create')
    .map((disk) => disk.name)
  return [...namesOfDisksToBeCreated, ...existingDiskNames]
}
```

david-crespo: I just confirmed that typing in the disk name field of disk create does not render the overall form, so it is better to leave it as-is.

david-crespo (approved):

PR #2542, opened 2024-11-09
Generate boot disk name from image name

### LINKED ISSUE #2476: Generate boot disk name from image name instead of ID
`abc-ae46ddf5-a8d5-40fa-bcda-fca-71c32d` is not the most useful name. We could stick "silo" or "project" in there to mark whether it was a silo or project image, but that would probably be more confusing than not. We already have the generated random part to avoid name collisions.

https://github.com/oxidecomputer/console/blob/d634c8f03f45761920a0c818ff6d71a8bd8edde1/app/forms/instance-create.tsx#L82-L97

charliepark:
This update changes how the fallback boot disk name is generated. Currently, we use the ID of the image used as the boot disk, but we'd like to use the image's name instead. The user is still able to set a boot disk name, as normal.

We are also looking into a means of showing the image used as the boot disk so that info is available whether the user names their boot disk or not, but that's a separate issue from this one.

Closes #2476

### DIFF
--- a/app/forms/instance-create.tsx
+++ b/app/forms/instance-create.tsx
@@ -22,6 +22,7 @@ import {
   usePrefetchedApiQuery,
   type ExternalIpCreate,
   type FloatingIp,
+  type Image,
   type InstanceCreate,
   type InstanceDiskAttachment,
   type NameOrId,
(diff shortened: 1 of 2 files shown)


charliepark:
Screenshot 2024-11-08 at 4 40 35 PM

charliepark:
Verified that a longer-than-expected name doesn't break things:
Screenshot 2024-11-11 at 10 06 25 PM

david-crespo (approved):
nice!

Issue #1505, opened 2023-05-19
Serial console search


david-crespo:
You can't ctrl-f the contents of the serial console scrollback — you have to use the search addon and write an explicit UI that calls `searchAddon.findNext()`.

https://www.npmjs.com/package/@xterm/addon-search

Issue #1489, opened 2023-05-15
Want easier way to copy serial console output from a machine in a reboot loop


citrus-it:
I have a VM which is failing to start, crashing after the OS is loaded.
Copying the error message using the console serial console was tricky because the console clears when the VM resets and I could not scroll back.

The VM is `wilma` which is in the `andy` project in the `google` silo, and it is currently rebooting continuously. The text I wanted to capture was:

```
OmniOS r151046 Version omnios-r151046-82ebda23c9 64-bit
Copyright (c) 2012-2017 OmniTI Computer Consulting, Inc.
Copyright (c) 2017-2023 OmniOS Community Edition (OmniOSce) Association.
NOTICE: Performing full ZFS device scan!
NOTICE: Cannot read the pool label from '/pci@0,0/pci1af4,2@4/blkdev@0,0:b'
NOTICE: spa_import_rootpool: error 5
Cannot mount root on /pci@0,0/pci1af4,2@4/blkdev@0,0:b fstype zfs

panic[cpu0]/thread=fffffffffbca07a0: vfs_mountroot: cannot mount root

Warning - stack not written to the dump buffer
fffffffffbca6650 fffffffffbaf0087 ()
fffffffffbca6690 genunix:main+137 ()
fffffffffbca66a0 unix:_locore_start+88 ()

skipping system dump - no dump device configured
rebooting...
```

Presumably the CLI serial console will make this easier, but is there any improvement we can make on the console side? Even a pause button would be a big help here.

PR #2543, opened 2024-11-11
Mock serial console with MSW

### LINKED ISSUE #2071: Make clear serial console isn't supposed to work in MSW mode
There is a to-do to handle errors on connection failure. That would sort of cover us here, but we could include a special message when `process.env.MSW` is true.

https://github.com/oxidecomputer/console/blob/7490104a3509b4d58afcc03fb9cee425caea24cc/app/pages/project/instances/instance/SerialConsolePage.tsx#L57-L67

### LINKED ISSUE #2072: Mock serial console with MSW (once they add websockets)
MSW are actively working on websockets mocking: https://github.com/mswjs/msw/pull/2011

They have an RC release for 2.3 and the PR is ready for review, so I expect this to be out in the next few weeks.

david-crespo:
Closes #2071
Closes #2072 

MSW finally added websockets mocking in [v2.6](https://github.com/mswjs/msw/releases/tag/v2.6.0). It's very easy to use in a basic way, but it will take a little more work to get it to act more like the real API does.

- [x] Basic connection works
- [ ] Only connect in instance states that would let you connect in the real API
- [ ] More realistic initial response and a little behavior on typing

This is not merely cosmetic! This will help enormously with serial console feature work like #1505 and #1489 because it means we can write Playwright tests that exercise those features.

### DIFF
--- a/mock-api/msw/handlers.ts
+++ b/mock-api/msw/handlers.ts
@@ -5,7 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
-import { delay } from 'msw'
+import { delay, ws } from 'msw'
 import * as R from 'remeda'
 import { validate as isUuid, v4 as uuid } from 'uuid'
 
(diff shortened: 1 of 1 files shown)


david-crespo:
Initial version:

https://github.com/user-attachments/assets/907d030a-e37f-400a-92ee-9a51c2a1208b

david-crespo:
Closing for now so it's not just sitting in the list.

PR #2544, opened 2024-11-11
Update system/silo picker for new breadcrumbs

david-crespo:
Followup to #2529 

### Fleet viewer

https://github.com/user-attachments/assets/b296c174-2140-42bd-83b1-6afb010036bd

### Not fleet viewer

Screenshot 2024-11-15 at 3 16 15 PM


charliepark (approved):
This looks great. 

I think regular fleet admin users might get confused at first to not have the left-side controls, so we should A) be sure to mention this change in the v12 release notes and B) give Izzy and crew a heads-up in case they get questions on it from admins.

benjaminleonard (approved):

PR #2545, opened 2024-11-12
cleanup: use RR route module exports to clean up `routes.tsx`

david-crespo:
I didn't bother doing all of them, but I did a bunch of the ones that make the route file shorter as a proof of concept. The route config is really verbose and it makes it a bit unwieldy, especially after #2529. This change is better than it looks in terms of line count because we are adding a few lines outside the route config to save lines inside it:

```
$ git diff --stat main... -- app/routes.tsx
 app/routes.tsx | 285 ++++++++++++++++++++------------------------------------------------
 1 file changed, 82 insertions(+), 203 deletions(-)
```

However, this change has another big advantage besides looking nice: by conforming to RR's route module API, we can trivially convert `<Route {...ProjectsPage}>` to `<Route lazy={() => import('./pages/ProjectsPage')}` to get bundle splitting when and where we want it. This all also will make it easier to convert to the [lovely new route config format](https://reactrouter.com/dev/framework/start/routing) in React Router v7.

<details>
<summary>Example use of <code>&lt;Route lazy&gt;</code></summary>

image

</details>

### DIFF
--- a/app/routes.tsx
+++ b/app/routes.tsx
@@ -24,7 +24,7 @@ import { CreateImageFromSnapshotSideModalForm } from './forms/image-from-snapsho
 import { CreateImageSideModalForm } from './forms/image-upload'
 import { CreateInstanceForm } from './forms/instance-create'
 import { CreateIpPoolSideModalForm } from './forms/ip-pool-create'
-import { EditIpPoolSideModalForm } from './forms/ip-pool-edit'
+import * as IpPoolEdit from './forms/ip-pool-edit'
 import { IpPoolAddRangeSideModalForm } from './forms/ip-pool-range-add'
 import { CreateProjectSideModalForm } from './forms/project-create'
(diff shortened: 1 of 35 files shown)


app/routes.tsx
```diff
@@ -134,21 +130,12 @@ export const routes = createRoutesFromElements(
       </Route>
 
       <Route path="system" element={<SystemLayout />} loader={SystemLayout.loader}>
-        <Route
-          element={<SilosPage />}
-          loader={SilosPage.loader}
-          handle={makeCrumb('Silos', pb.silos())}
-        >
+        <Route {...SilosPage} handle={makeCrumb('Silos', pb.silos())}>
```
david-crespo, line 133: We could export handle from the module too, but I think that would make it impossible to reason about the crumbs by looking at the config. The components and loaders don’t have this problem.

Issue #2546, opened 2024-11-12
Disallow combobox custom values made only of whitespace


david-crespo:
Probably enough to trim it, or I guess we could disallow typing spaces entirely since none of values can have spaces in them.

![Image](https://github.com/user-attachments/assets/956d425d-8370-41da-bb34-a664917eae7b)

PR #2549, opened 2024-11-13
Set up normalizeName string utility function

charliepark:
We are already restricting some of the characters that users can add to the Name field. While working on #2546 I figured that it might be helpful to have this string manipulation available in an external function, so we can restrict what's typed in to the Combobox field. Because we'll want to give users the option of typing a number in as the initial character ("I'm trying to filter for Images with `2024` in them"), the `nameSyntax` function has an optional second prop. Left off, the default action is to only permit a letter in the zeroth position. If `true` is passed in to `allowNonLetterStart`, the user can enter either a number or a dash. Non-allowed characters are dropped.

### DIFF
--- a/app/components/form/fields/NameField.tsx
+++ b/app/components/form/fields/NameField.tsx
@@ -7,7 +7,7 @@
  */
 import type { FieldPath, FieldValues } from 'react-hook-form'
 
-import { capitalize, nameSyntax } from '~/util/str'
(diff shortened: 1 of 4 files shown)


david-crespo (approved):
Good idea! I'd rename the function — "syntax" is not the right word for these constraints. A maximally literal option would be `constrainNameChars`. A more abstract option is `normalizeName`.

PR #2550, opened 2024-11-13
Disallow firewall-rule subform submission when target / host filter value is missing

### LINKED ISSUE #2547: Firewall rules: disallow submitting empty target or host
This is moved around a bit in #2540 but the problem is the same: the `if (!type || !value) return` is doing what it's supposed to do, but the form still resets on empty submit. Not sure why.

https://github.com/oxidecomputer/console/blob/31520a2e2a59a15b371d8cca6891af46da67203e/app/forms/firewall-rules-common.tsx#L323-L330

https://github.com/user-attachments/assets/743fb133-90c1-449d-b0d2-f1a1d428ec46

charliepark:
We already had the logic in place to disable the `clear` button in the subform, we just hadn't applied it to the submit button. This PR applies the same prop logic to both the `clear` and the `add` button — you can't take either action if there's no value in the subform; if the value is there, you can take either action.

Closes #2547

### DIFF
--- a/test/e2e/firewall-rules.e2e.ts
+++ b/test/e2e/firewall-rules.e2e.ts
@@ -160,6 +160,9 @@ test('firewall rule form targets table', async ({ page }) => {
 
   const addButton = page.getByRole('button', { name: 'Add target' })
 
+  // addButton should be disabled until a value is added
+  await expect(addButton).toBeDisabled()
+
   // add targets with overlapping names and types to test delete
 
   await targetVpcNameField.fill('abc')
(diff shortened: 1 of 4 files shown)


test/e2e/firewall-rules.e2e.ts
```diff
@@ -181,7 +184,10 @@ test('firewall rule form targets table', async ({ page }) => {
   // add a subnet by selecting from a dropdown; make sure 'mock-subnet' is present in the dropdown,
   // even though VPC:'mock-subnet' has already been added
   await selectOption(page, 'Target type', 'VPC subnet')
+  // addButton should be disabled again, as type has changed but no value has been entered
+  await expect(addButton).toBeDisabled()
   await selectOption(page, subnetNameField, 'mock-subnet')
+  await expect(addButton).toBeEnabled()
```
charliepark, line 190: This is technically covered by the `click()` and `expectRowVisible` on the next two lines, but I think having this line here (along with the `expect…Disabled` two lines up) makes it clear that the `selectOption` on the line in the middle of them is what enabled the button.

david-crespo (approved):

PR #2553, opened 2024-11-14
normalize combobox inputs; disallow spaces

### LINKED ISSUE #2546: Disallow combobox custom values made only of whitespace

charliepark:
In #2549, we set up a function (`normalizeName`) that allows us to normalize inputs so they conform to the relevant string patterns for the Name field. This PR uses that same function and applies it to the values entered in to comboboxes. As a result of this, #2546 is fixed, as the function doesn't allow for whitespace characters.

This is also the usecase where the `normalizeName` second parameter (`allowNonLetterStart`) comes in, as people might need to search for a number inside a combobox (e.g. "filter the combobox for all instance names that include `22`").

Closes #2546


app/ui/lib/Combobox.tsx
```diff
@@ -189,17 +191,25 @@ export const Combobox = ({
...
+                selectedItemValue
+                  ? allowArbitraryValues
+                    ? selectedItemValue
+                    : selectedItemLabel
+                  : query
               }
               onChange={(event) => {
+                // Pass in true for the second prop so that the user can start their filtering query with a number.
+                // A validation will still prevent them from submitting the form with a number at the beginning.
+                const value = normalizeName(event.target.value, true)
```
david-crespo, line 208: Maybe this function shouldn't be hard-coded in here. It's true that currently the values are all names, but this is meant to be a generic `Combobox`, not a `NameCombobox`. How about a `transform` prop like `TextField`?

https://github.com/oxidecomputer/console/blob/f3d38103253dab2d2670725a10a7d91b26e7937f/app/components/form/fields/TextField.tsx#L49-L50

https://github.com/oxidecomputer/console/blob/f3d38103253dab2d2670725a10a7d91b26e7937f/app/components/form/fields/TextField.tsx#L121

app/ui/lib/Combobox.tsx
```diff
@@ -189,17 +195,25 @@ export const Combobox = ({
...
+              value={
+                selectedItemValue
+                  ? allowArbitraryValues
+                    ? selectedItemValue
+                    : selectedItemLabel
+                  : query
               }
               onChange={(event) => {
+                // Pass in true for the second prop so that the user can start their filtering query with a number.
+                // A validation will still prevent them from submitting the form with a number at the beginning.
```
david-crespo, line 211: comment no longer accurate

david-crespo (approved):

PR #2555, opened 2024-11-14
Add crumb for sled ID using middle-truncated UUID

david-crespo:
I left out the crumb for the sled because they don't have names, and a UUID in the crumb looks ridiculous. But it occurred to me I could just truncate it and it would be fine. Much better than nothing. Nice demo of the crumbs snapshot test, too.

image


charliepark (approved):
nice

benjaminleonard:
Nice, I like it. The rare use for a middle truncation.

PR #2560, opened 2024-11-18
chore: bump playwright to 1.49

david-crespo:
Playwright 1.49 has an interesting new Chrome version that upgrades the way they do headless mode so that it's more authentic Chrome, but potentially slightly slower. It could have breaking changes, but it works fine for me locally and didn't seem slower either. I also wanted to upgrade for a new Safari version since we've been seeing some Safari flake lately.

https://github.com/microsoft/playwright/releases/tag/v1.49.0

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -33,6 +33,7 @@ export default {
           permissions: ['clipboard-read', 'clipboard-write'],
         },
         ...devices['Desktop Chrome'],
+        channel: 'chromium',
       },
     },
     {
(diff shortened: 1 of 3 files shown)


playwright.config.ts
```diff
@@ -33,6 +33,7 @@ export default {
           permissions: ['clipboard-read', 'clipboard-write'],
         },
         ...devices['Desktop Chrome'],
+        channel: 'chromium',
```
david-crespo, line 36: This opts into the new headless Chrome thing.

david-crespo:
The same test failed in both Chrome and Safari, which suggests non-flake. Neither failed for me locally, though. Will have to look into it.

david-crespo:
Actually that test is the same once that keeps flaking in Safari. Ugh.

PR #2561, opened 2024-11-19
Set date pickers to match other input elements' heights

charliepark:
Woof. We recently made comboboxes / listboxes larger, to match the height of input fields. The DatePicker component had an independent height, though, and so it wasn't matching the height of the listbox on the Metrics page. This fixes that, so DatePickers now match the other input components' heights.

Broken:
Screenshot 2024-11-18 at 4 48 40 PM

Fixed:
Screenshot 2024-11-18 at 4 48 30 PM


david-crespo (approved):
nice. whoops

PR #2567, opened 2024-11-20
refactor: new `useQueryTable` with legit types, returns the query data

david-crespo:
This is built on top of the `queryOptions` work in #2566. The types in `QueryTable` have always bothered me, as does returning a component instead of an element, but I could never figure out how to fix it.

https://github.com/oxidecomputer/console/blob/0f922b2ed3af596beae37adaae4ecff545002ed0/app/table/QueryTable.tsx#L65-L77

This change:

* Fixes the types so we don't use `any`
* Uses a much more intuitive mechanism to enforce that we can only use QueryTable with list endpoints (simply require `TData extends { items: TItem[] }`)
* Returns the React Query query result containing the data, so in cases where we are currently making a second `useApiQuery` call where we painstaking ensure the params match (see below), we don't have to do that anymore
* Keeps previous page around while loading next page, preventing blank screen flash, and adds a small loading spinner (see #1861)

https://github.com/oxidecomputer/console/blob/0f922b2ed3af596beae37adaae4ecff545002ed0/app/pages/ProjectsPage.tsx#L64-L67

The new file is 80 lines instead of 120 and strictly better across the board. There are only 20 `useQueryTable` calls in the app. It would be pretty quick to convert them all. We could do that here or merge this and convert in batches.

### DIFF
--- a/app/api/hooks.ts
+++ b/app/api/hooks.ts
@@ -21,6 +21,7 @@ import {
   type UseQueryOptions,
   type UseQueryResult,
 } from '@tanstack/react-query'
+import * as R from 'remeda'
 import { type SetNonNullable } from 'type-fest'
 
 import { invariant } from '~/util/invariant'
@@ -124,16 +125,32 @@ export const getApiQueryOptions =
       ...options,
     })
 
+// Managed here instead of at the display layer so it can be built into the
+// query options and shared between loader prefetch and QueryTable
 export const PAGE_SIZE = 25
+
+/**
+ * This primarily exists so we can have an object that encapsulates everything
+ * useQueryTable needs to know about a query. In particular, it needs the page
+ * size, and you can't pull that out of the query options object unless you
+ * stick it in `meta`, and then we don't have type safety.
+ */
+export type PaginatedQuery<TData> = {
+  optionsFn: (
+    pageToken?: string
+  ) => UseQueryOptions<TData, ApiError> & { queryKey: QueryKey }
+  pageSize: number
+}
 
 /**
  * This is the same as getApiQueryOptions except for two things:
  *
  *   1. We use a type constraint on the method key to ensure it can
  *      only be used with endpoints that return a `ResultsPage`.
- *   2. Instead of returning the options directly, it returns a function that
- *      takes `limit` and `pageToken` and merges them into the query params so
- *      that these can be passed in by `QueryTable`.
+ *   2. Instead of returning the options directly, it returns a paginated
+ *      query config object containing the page size and a function that
+ *      takes `limit` and `pageToken` and merges them into the query params
+ *      so that these can be passed in by `QueryTable`.
  */
 export const getListQueryOptionsFn =
   <A extends ApiClient>(api: A) =>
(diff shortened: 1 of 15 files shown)


david-crespo:
Because `useQueryTable` now returns the data, I discovered a bug on `main` due to the refetch being driven by [its own `useQuery` call](https://github.com/oxidecomputer/console/blob/c425880458522b045996836790e50bd2ae28c88d/app/pages/project/instances/InstancesPage.tsx#L80-L82) instead of using the same data as `QueryTable`. When you move to a page other than first, `QueryTable` has that page's instances in scope, but the one looking for transitioning instances is permanently stuck on the first page. So if you're on the second page, and there are transitioning instances but none on the first page, it does not poll as you expect.

https://github.com/user-attachments/assets/f64d6d94-8a59-4b7b-9271-6e6f0512bbcf

app/api/hooks.ts
```diff
@@ -123,6 +124,37 @@ export const getApiQueryOptions =
...
+ *      that these can be passed in by `QueryTable`.
+ */
+export const getListQueryOptionsFn =
+  <A extends ApiClient>(api: A) =>
+  <
+    M extends string &
+      {
+        // this helper can only be used with endpoints that return ResultsPage
+        [K in keyof A]: Result<A[K]> extends ResultsPage<unknown> ? K : never
+      }[keyof A],
```
david-crespo, line 145: What this does:

image

app/pages/project/snapshots/SnapshotsPage.tsx
```diff
@@ -52,12 +54,12 @@ const EmptyState = () => (
   />
 )
 
+const snapshotList = (project: string) => getListQFn('snapshotList', { query: { project } })
```
david-crespo, line 57: We can now share the query options between the prefetch in the loader and the query table, and we don't  have to specify `PAGE_SIZE` manually here to make sure it matches the query inside QueryTable.

app/table/QueryTable2.tsx
```diff
@@ -0,0 +1,79 @@
...
+  emptyState: React.ReactElement
+  // React Table does the same in the type of `columns` on `useReactTable`
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+  columns: ColumnDef<TItem, any>[]
+}
+
+// require ID only so we can use it in getRowId
+export function useQueryTable<TItem extends { id: string }>({
+  optionsFn,
+  pageSize = PAGE_SIZE,
```
david-crespo, line 39: There's something off that I need to figure out about the fact that `PAGE_SIZE` has a default both here _and_ in `getListQueryOptionsFn`. It means you can manually override the page size in the loader prefetch call and a) it won't apply to the QueryTable, and b) the only way you find out about the mismatch is that the prefetch check in QueryTable fails. It's not the worst (it's more or less what we have now) but it doesn't feel right — it doesn't feel like it's taking advantage of the query options sharing.

I feel like ideally you would not have to specify page size on `useQueryTable` at all; instead, you would set it on the query options object shared by the loader and the query table. It's easy to build the page size into the options object — the only real obstacle is that inside `useQueryTable` we need the `pageSize` for UI reasons (whether to enable the next button), so if we build pageSize into the query options, we need to be able to pull it back out of there. There is a `meta` key on query options that we could put `pageSize` on.

david-crespo:
Figured out https://github.com/oxidecomputer/console/pull/2567#discussion_r1852706175 in 8cab460ba3fab7641edb14b4199cb3ffa6832c81. This video shows how the page size for a given page, both loader and `useQueryTable` call, is controlled in a single spot. There is no need to keep the page size in sync manually. The global default page size is controlled at the data layer by the helper that produces the paginated query config object. `useQueryTable` doesn't know anything about the default, it just takes a prop.

The error on hot reload when I change the value is an `ensurePrefetch` failure inside of `useQueryTable` because that query changes, but the loader does not refire, so what's in the cache no longer matches the page size. We don't have to worry about this in production until.

https://github.com/user-attachments/assets/f1c625d4-8512-49d9-86d0-29021495a2d7

app/api/hooks.ts
```diff
@@ -123,6 +125,62 @@ export const getApiQueryOptions =
...
+    // pathOr plays nice when the properties don't exist
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const limit = R.pathOr(params as any, ['query', 'limit'], PAGE_SIZE)
+    return {
+      optionsFn: (pageToken?: string) => {
+        const newParams = { ...params, query: { ...params.query, limit, pageToken } }
+        return getApiQueryOptions(api)(method, newParams, {
+          ...options,
+          // identity function so current page sticks around while next loads
+          placeholderData: (x) => x,
```
david-crespo, line 177: I just remembered the problem with this, which made me reluctant to do it everywhere, though doing it in just a few random tables (what we currently do) is probably even worse. When we don't do the blank flash, we don't scroll to top when the next page loads.

Screenshot 2024-11-19 at 12 31 43 PM

david-crespo: GOT EM

https://github.com/user-attachments/assets/f58b0473-67d0-4310-b65d-408e37d6c402

david-crespo:
Loading spinner. We're in business.

https://github.com/user-attachments/assets/57c15945-34d7-4d32-b4cf-d895428aa0cb

PR #2576, opened 2024-11-22
Improve external ips row

### LINKED ISSUE #2570: Improve instance page's external IPs row
I was fairly sure we were handling truncation and also that we were listing Floating IPs before the Ephemeral IP in the "external IPs" minitable row on Instance pages, but we apparently aren't doing either of those.

Image

There was also a request to consider some way to distinguish between Floating and Ephemeral IPs in that row (even though that info is on the Networking tab). Possibly a modal, or an `E` badge or something.

charliepark:
This PR moves the Ephemeral IP to the end of the list of IPs in the minitable at the top of the screen. It also caps the number of IPs shown, with an overflow badge (`+3` or whatever) that directs the user to the Networking page.

Showing two IPs, with the Floating IP first:
Screenshot 2024-11-22 at 2 35 12 PM
When there are more than 2 IPs:
Screenshot 2024-11-22 at 2 35 25 PM

Closes #2570


charliepark:
Added a little spacing to the left of the badge.

Would be open to adding a slash between the last IP and the `+1` badge, if that feels better.

charliepark:
Screenshot 2024-11-22 at 3 48 23 PM

benjaminleonard:
The badge is more visible, but we're not using it in other cases of overflow. Let's match the treatment elsewhere that just uses the mono. We might need some container to distinguish it from the addresses themselves but let's start there.

david-crespo:
I love the idea of linking to the networking tab. What if we used an info icon to link to the networking tab even when there is no overflow? That could let us avoid having to indicate which are floating or ephemeral inline.

david-crespo:
I don't like the mono `+1` link, it looks too close to an IP — reminds me of a country code in a phone number. I don't think badge works either because badges are never links (can't think of any). So I'm not sure what the visual treatment should be. Need help @benjaminleonard.

image

app/components/ExternalIps.tsx
```diff
@@ -6,13 +6,21 @@
...
 import { EmptyCell, SkeletonCell } from '~/table/cells/EmptyCell'
 import { CopyableIp } from '~/ui/lib/CopyableIp'
+import { Slash } from '~/ui/lib/Slash'
 import { intersperse } from '~/util/array'
+import { pb } from '~/util/path-builder'
 import type * as PP from '~/util/path-params'
 
+/** Move ephemeral IP (if present) to the end of the list of external IPs */
+export const orderIps = (ips: ExternalIp[]) =>
+  ips.sort((a) => (a.kind === 'ephemeral' ? 1 : -1))
```
david-crespo, line 22: `import * as R from 'remeda'` and use [`R.sortBy`](https://remedajs.com/docs/#sortBy) to avoid the mutation (like the comment said). I think this mutates what's in the query cache, which freaks me out.

david-crespo:
Very happy with the final treatment. Looks and feels right.

image

app/pages/project/instances/instance/tabs/NetworkingTab.tsx
```diff
@@ -371,7 +372,7 @@ export function Component() {
 
   const ipTableInstance = useReactTable({
     columns: useColsWithActions(staticIpCols, makeIpActions),
-    data: eips?.items || [],
+    data: orderIps(eips.items),
```
david-crespo, line 375: I think you need to memoize this to avoid the table blowing up on every render. This would manifest, for example, as the row actions menu closing on every instance poll (which I need to put in more tests for, as mentioned in https://github.com/oxidecomputer/console/pull/2618#issuecomment-2539806844).

david-crespo (approved):

PR #2577, opened 2024-11-23
cleanup: delete unused API helper things

david-crespo:
Followup to 

* #2567 
* #2571 
* #2575

### DIFF
--- a/app/api/hooks.ts
+++ b/app/api/hooks.ts
@@ -9,14 +9,11 @@ import {
   hashKey,
   queryOptions,
   useMutation,
-  useQueries,
   useQuery,
-  type DefaultError,
   type FetchQueryOptions,
   type InvalidateQueryFilters,
   type QueryClient,
   type QueryKey,
-  type UndefinedInitialDataOptions,
   type UseMutationOptions,
   type UseQueryOptions,
   type UseQueryResult,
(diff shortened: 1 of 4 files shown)


app/api/hooks.ts
```diff
@@ -289,35 +278,6 @@ export const getUseApiMutation =
...
-              api[method](params, { signal }).then(handleResult(method)),
-            throwOnError: (err: ApiError) => err.statusCode === 404,
-            ...options,
-            // Add params to the result for reassembly after the queries are returned
-            select: (data) => ({ ...data, params }),
-          }) satisfies UseQueryOptions<Result<A[M]> & { params: Params<A[M]> }, ApiError>
-      ),
-    })
-  }
-
```
david-crespo, line 320: This (and `useApiQueries`) wasn't used. I kept it around in case we needed it again, but I think if we wanted this now, we'd use the query options helper with `map` and pass the resulting array of options objects to the built-in `useQueries`.

app/api/hooks.ts
```diff
@@ -29,17 +26,12 @@ import { invariant } from '~/util/invariant'
...
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
-export type Params<F> = F extends (p: infer P) => any ? P : never
-export type Result<F> = F extends (p: any) => Promise<ApiResult<infer R>> ? R : never
-export type ResultItem<F> =
-  Result<F> extends { items: (infer R)[] }
-    ? R extends Record<string, unknown>
-      ? R
-      : never
-    : never
```
david-crespo, line 42: `ResultItem` was only used by the old QueryTable, I think.

charliepark: oof; that's an ugly one

charliepark (approved):

david-crespo:
The test that failed in Chrome failed twice the same way. It passes locally, but the flake might still be an indicator of something, like maybe I introduced some extra renders that were previously optimized out. Though it would have to come from a a previous PR because one doesn't change any runtime code that's actually used.

image

PR #2579, opened 2024-11-26
Update silo creation form to require a TLS certificate

### LINKED ISSUE #2578: Uploading a TLS certificate should be mandatory during silo creation
Currently the console does not require a TLS certificate to be uploaded during the silo creation workflow.

This is a mandatory field in the API and failure to provide a cert may leave the newly created silo broken in unpredictable ways.

charliepark:
This fixes a bug in the silo creation form, where a silo could be created without entering in TLS information.

The sub-form already had the TLS fields marked as required, but those requirements don't bubble up to the higher-order form. This adds a requirement to the body of the form, that the `tls_certificates` array be not-empty.

Augustus had an interesting idea about verifying Common Name and Subject Alternative Names in the cert fields, but this PR doesn't address that yet.

Closes #2578


charliepark:
Chatting through this with Ben, and he's noting that it would be good for us to consider a `minItems: 1` on the API for `tls_certificates`, so this is protected via CLI as well. 

All of the current types in Nexus that have a `minItems` value also have an identical `maxItems` value, so are of a fixed length. So if approaching this in Rust, we might need to approach it differently than in other spots in the code.

benjaminleonard:
> Chatting through this with Ben, and he's noting that it would be good for us to consider a `minItems: 1` on the API for `tls_certificates`, so this is protected via CLI as well.

Rather I meant that the `minItem` requirement is then documented in the API docs. Which we could probably also do with a comment.

benjaminleonard:
This is a simple one, worth squeezing into R12? @david-crespo @charliepark

david-crespo:
Yeah, probably.

app/components/form/fields/TlsCertsField.tsx
```diff
@@ -26,8 +27,13 @@ export function TlsCertsField({ control }: { control: Control<SiloCreateFormValu
...
 
   const {
-    field: { value: items, onChange },
-  } = useController({ control, name: 'tlsCertificates' })
+    field: { value: items, onChange, ref },
+    fieldState: { error },
+  } = useController({
+    control,
+    name: 'tlsCertificates',
+    rules: { required: true },
```
david-crespo, line 35: This seems to work, but I was surprised about it since `[]` is truthy. Found this in the [docs](https://react-hook-form.com/docs/useform/register) recommending `validate` over `required` for object and array values. I'll make that tweak now.

image

david-crespo (approved):
This all looks good to me. I'm still a little concerned about requiring a field in the UI that the API does not actually require, but the fix for that is almost certainly to have the API 400 if the list of certs has less than one element.

This looks a little tight, could use some space @benjaminleonard ?

image

benjaminleonard:
Ship it, I feel like I want to rework that pattern where mini-table is hidden when there's no elements anyway, doesn't sit quite right.

PR #2582, opened 2024-11-27
Initial cert validation test

### LINKED ISSUE #2580: TLS cert soft failure
https://github.com/oxidecomputer/console/issues/2578#issuecomment-2499184245

> Similar to the "soft-failures" we have on image import, we could probably decode the certificate and verify that either the the CN or SAN covers the computed silo url `{silo}.sys.{domain}`. Less useful for long term users, but likely nice for initial silo creation.

Seems like a quick win, design work is minimal since we can just reuse the pattern from image import. @augustuswm have you got a test cert you could send to play with.

benjaminleonard:
Fixes #2580 and related to #2578 

Idea from @augustuswm to soft validate the cert. Unfortunately I think we need a library to parse the cert but I lazy imported it to avoid adding to the main bundle since most users will not see this.

Principally the risk here is what I dont know about certs:

1. wildcard handling might need to be sophisticated
2. certs that arrange slightly different in a way that `parseCertificate()` doesn't catch

But feels, like the image soft validation, that it could be a good quality of life improvement for something that is potentially very painful when not done correctly. We might want to label the `CN` and `SAN` on the messages list of found names.

image


app/components/form/fields/TlsCertsField.tsx
```diff
@@ -147,3 +176,88 @@ const AddCertModal = ({ onDismiss, onSubmit, allNames }: AddCertModalProps) => {
...
+}
+
+function matchesDomain(pattern: string, domain: string): boolean {
+  const patternParts = pattern.split('.')
+  const domainParts = domain.split('.')
+
+  if (patternParts.length !== domainParts.length) return false
+
+  return patternParts.every(
+    (part, i) => part === '*' || part.toLowerCase() === domainParts[i].toLowerCase()
```
benjaminleonard, line 204: Would it just be the first part that accepts a wildcard?

package.json
```diff
@@ -51,10 +52,12 @@
     "classnames": "^2.5.1",
     "date-fns": "^3.6.0",
     "filesize": "^10.1.6",
+    "i": "^0.3.7",
     "lodash.throttle": "^4.1.1",
     "match-sorter": "^6.3.4",
     "md5": "^2.3.0",
     "mousetrap": "^1.6.5",
+    "npm": "^10.9.1",
```
david-crespo, line 60: lol you `npm i npm i`

benjaminleonard: haha what an idiot

app/components/form/fields/TlsCertsField.tsx
benjaminleonard: I presume it would be the left-most wildcard subdomain is the only allowable?

Would need adjusting, and some unit tests also.

augustuswm: Correct only the left most subdomain can be a wildcard. For example `*.r3.oxide-preview.com`  does not match `silo.sys.r3.oxide-preview.com`

benjaminleonard:
Added tests, improved wildcard handling and also a message in case the certificate cannot be parsed.

image

Message copy needs work.

charliepark:
Is this your card? https://github.com/oxidecomputer/console/blob/require_tls_cert/app/forms/idp/util.ts#L16-L17

augustuswm:
This looks awesome, thanks for picking it up. I think initial rack install will certainly benefit from it.

benjaminleonard:
> Is this your card? https://github.com/oxidecomputer/console/blob/require_tls_cert/app/forms/idp/util.ts#L16-L17

Do we need some copy to caveat incase that is not correct. I remember some talk about a proxy perhaps throwing it off?

app/components/form/fields/TlsCertsField.tsx
```diff
@@ -103,10 +114,18 @@ type AddCertModalProps = {
...
-  const { control, handleSubmit } = useForm<CertFormValues>({ defaultValues })
+const AddCertModal = ({ onDismiss, onSubmit, allNames, siloName }: AddCertModalProps) => {
+  const { watch, control, handleSubmit } = useForm<CertFormValues>({ defaultValues })
+
+  const file = watch('cert')
+
+  const { data: certValidation } = useQuery({
+    queryKey: ['validateImage', ...(file ? [file.name, file.size, file.lastModified] : [])],
+    queryFn: file ? () => file.text().then(parseCertificate) : skipToken,
+  })
```
david-crespo, line 128: @charliepark pointed out we might want to set the stale time directly here. I'm going to experiment with that and also maybe tweaking the query key.

david-crespo:
Tweaked the copy and fixed a bug where it would only do validation if the silo name was non-empty in the parent form. I changed it so it only opts out of the part of the validation that requires the silo name.

image

benjaminleonard:
Updated from main and a little review:

**Major**
- Replaced `r2.oxide-preview.com` with `getDelegatedDomain(window.location)` as per @charliepark's snippet

**Smaller polish**
- Renamed the stale `validateImage` query key, dropped the redundant match arm
- `matchesDomain` now normalizes case (wildcard branch was case-sensitive) and trailing dots up front, and rejects the pathological `*.` pattern.
- Warns when the uploaded cert is already past its `notAfter`. Stacks with the domain-mismatch warning when both apply.

**Tests**
- Vitest: wildcard + uppercase, trailing dot, `*.` rejection, multi-PEM bundle, whitespace/BOM, expiry parse.
- Playwright (`silos.e2e.ts`): asserts "Couldn't parse certificate" on the existing garbage upload, then uploads a valid matching cert (no notice fires) and a real expired non-matching PEM (both notices fire), exercising the FileField → useQuery → parseCertificate → matchesDomain → notice pipeline end to end.

---

Can we get one last sanity check and revisit merging?

david-crespo:
I took out the dynamic import in 765210b796207700ccce3633cffbac405401a98e because even though technically certs aren't required by the API in the silo create body, in practice they are necessarily pretty much all the time. So that import will always happen, might as well absorb it into the chunk for the form. Had the robot compare before/after:

image

benjaminleonard:
LGTM

david-crespo:
Let's give this this a last look but I think it's ready to go.

PR #2583, opened 2024-11-27
Use `min()` for modal top position

### LINKED ISSUE #2523: Use `min()` for modal top position
On taller screens the modal seems like it comes down a bit far (at 50%). Also, if the height of the modal changes (with some dynamic content) at all the position of say an input will no longer be where it was previously.

We could use `min()` – that way we make it a fixed position at the top and then have it snap to 50% if the modal moves to a threshold.

Image

benjaminleonard:
Fixes #2523 

Feels much more comfortable IMO.


charliepark (approved):
Shame about that merge conflict 😄🫠

New positioning looks good, though!

PR #2586, opened 2024-11-27
Make pressing enter submit instance resize modal form

### LINKED ISSUE #2585: Instance resize modal: enter doesn't submit the form
It uses a form, but the submit button isn't inside it, so it needs to be hooked up.

https://github.com/oxidecomputer/console/blob/059c55142ba29e4f691247def92493b7ef3f8df2/app/pages/project/instances/instance/InstancePage.tsx#L343

david-crespo:
Closes #2585 

`Modal.Footer` has more lines of props than lines of code. Not my favorite.

### DIFF
--- a/app/ui/lib/Modal.tsx
+++ b/app/ui/lib/Modal.tsx
@@ -9,6 +9,7 @@ import * as Dialog from '@radix-ui/react-dialog'
 import { animated, useTransition } from '@react-spring/web'
 import cn from 'classnames'
 import React, { forwardRef, useId } from 'react'
+import type { MergeExclusive } from 'type-fest'
 
 import { Close12Icon } from '@oxide/design-system/icons/react'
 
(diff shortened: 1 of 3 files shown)


app/ui/lib/Modal.tsx
```diff
@@ -151,6 +157,8 @@ Modal.Footer = ({
         {cancelText || 'Cancel'}
       </Button>
       <Button
+        type={formId ? 'submit' : 'button'}
+        form={formId}
```
david-crespo, line 161: This is a hack version, but it would be ideal if the button had no `onClick` and we relied on it triggering the form's `onSubmit`. Unfortunately that would require even an more confusing mix of props.

app/ui/lib/Modal.tsx
```diff
@@ -125,6 +126,21 @@ Modal.Body = classed.div`py-2 overflow-y-auto`
...
+type FooterProps = {
+  children?: React.ReactNode
+  onDismiss: () => void
+  actionType?: 'primary' | 'danger'
+  actionText: React.ReactNode
+  actionLoading?: boolean
+  cancelText?: string
+  disabled?: boolean
+} & MergeExclusive<{ formId: string }, { onAction: () => void }>
+
```
david-crespo, line 143: yoooo `MergeExclusive` is my new best friend

charliepark: 🤯

PR #2589, opened 2024-12-02
Add view/edit SSH key page

benjaminleonard:
Can't edit anything here, but it's helpful to have this to be able to see the public key – and it's easy to eventually allow editing when the API supports it.

image


app/forms/ssh-key-edit.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <DescriptionField name="description" control={form.control} disabled />
+      <TextField
+        as="textarea"
+        name="publicKey"
+        label="Public key"
+        required
+        rows={8}
+        control={form.control}
+        disabled
+      />
```
david-crespo, line 72: If there's any change the user might want to copy this, it's probably worth adding a button for it. Triple-clicking with the disabled cursor feels unpleasant.

test/e2e/ssh-keys.e2e.ts
```diff
@@ -19,6 +19,29 @@ test('SSH keys', async ({ page }) => {
...
   ])
 
+  // click name to open side modal
+  await page.getByRole('link', { name: 'm1-macbook-pro' }).click()
+
+  // verify side modal content
+  const modal = page.getByRole('dialog')
+  await expect(modal).toBeVisible()
+  await expect(modal.getByText('Edit SSH key')).toBeVisible()
+  await expect(modal.getByText('m1-macbook-pro')).toBeVisible()
```
david-crespo, line 29: test failure is because this matches multiple things

app/forms/ssh-key-edit.tsx
benjaminleonard: Sure, wasn't sure if it was outside the scope of this PR, modifying textarea.

Might be worth also adding it to the more actions menu.

test/e2e/utils.ts
```diff
@@ -36,18 +36,18 @@ export async function map<T>(
 
 type Selector = string | Locator
 
-const toLocator = (page: Page, selector: Selector): Locator =>
-  typeof selector === 'string' ? page.locator(selector) : selector
-
-export async function expectVisible(page: Page, selectors: Selector[]) {
+// note if `Locator` is used instead of string, the container is ignored
+export async function expectVisible(container: Page | Locator, selectors: Selector[]) {
```
benjaminleonard, line 40: What do you think of this @david-crespo ?

Alternatively we could remove `Locator` as an option and only use strings. It's only really used in two test files and the change would be trivial.

benjaminleonard:
Implemented in a bit of a hack, but works well. If we made the form editable, would need to remove or hook it up to the input with watch.

CleanShot 2024-12-03 at 10 59 15

test/e2e/utils.ts
david-crespo: Just use regular `expect().toBeVisible()`. I want to get rid of this utility anyway. I don’t use it in any new code.

PR #2596, opened 2024-12-03
Remove Ladle

david-crespo:
It's been broken for a while, I haven't looked at it in well over a year. All it gives me is type and lint errors to fix for no reason. Will wait until @benjaminleonard weighs in.


benjaminleonard:
Murder away, were we a bigger team id advocate to keep it but I think it's less than useful now.

PR #2597, opened 2024-12-04
refactor: start getting rid of `apiQueryClient`

david-crespo:
#2566 added a more flexible and powerful way of making typesafe API calls with React Query, but didn't convert things like `invalidateQueries` and `setQueryData` to use the new style to avoid churn. But we do eventually want to do that and get rid of `apiQueryClient` entirely. This PR aims to figure out what that will look like.


app/pages/ProjectsPage.tsx
```diff
@@ -64,9 +65,7 @@ export function Component() {
 
   const { mutateAsync: deleteProject } = useApiMutation('projectDelete', {
     onSuccess() {
-      // TODO: figure out if this is invalidating as expected, can we leave out the query
-      // altogether, etc. Look at whether limit param matters.
-      apiQueryClient.invalidateQueries('projectList')
+      invalidate('projectList')
```
david-crespo, line 68: So, ideally we could do this as something like this:

```ts
queryClient.invalidateQueries(projectList.optionsFn())
```

or (to be clearer about what's going on)

```ts
queryClient.invalidateQueries({ queryKey: projectList.optionsFn().queryKey })
```

The problem here is that that `queryKey` has way too much specificity. Even without a page token specified, it will have the `limit` on it:

https://github.com/oxidecomputer/console/blob/e4f70b01b1c14fd194d306b45a1ca87efa0f1468/app/api/hooks.ts?plain=1#L168-L169

which means it won't invalidate any other fetches for the project list that have a different limit, and it won't invalidate any cached queries for pages other than the first one. This shows the wisdom of our approach of just using the method key alone without any parameters, which has the effect of invalidating any cached `projectList` fetch, regardless of the details.

I have an idea, though. If the list option object has another method that returns just the method key, we could use that.

app/pages/ProjectsPage.tsx
```diff
@@ -77,11 +69,8 @@ export function Component() {
         onActivate: () => {
           // the edit view has its own loader, but we can make the modal open
           // instantaneously by preloading the fetch result
-          apiQueryClient.setQueryData(
-            'projectView',
-            { path: { project: project.name } },
-            project
-          )
+          const queryOptions = apiq('projectView', { path: { project: project.name } })
+          queryClient.setQueryData(queryOptions.queryKey, project)
```
david-crespo, line 73: I was surprised to discover this does actually enforce the type of `project` matches the endpoint. 

image

`setQueryData` is inferring it from the type of the options object.

image

image

app/api/client.ts
```diff
@@ -49,6 +53,24 @@ export const useApiMutation = getUseApiMutation(api.methods)
...
+   * Note that we only take a single argument, `method`, rather than allowing
+   * the full query key `[query, params]` to be specified. This is to avoid
+   * accidentally overspecifying and therefore failing to match the desired query.
+   * The params argument can be added in if we ever have a use case for it.
+   */
+  invalidateEndpoint(method: keyof typeof api.methods) {
+    this.invalidateQueries({ queryKey: [method] })
+  }
+}
+
```
david-crespo, line 73: This approach feels good, much better than a separate `invalidate` function (which I tried in 622bc80c076211d3d1adc10e18a7fa4dc6fa0bc5) or a separate `apiQueryClient`, like we have now. I like that it adds a new method and does not touch existing methods.

app/forms/project-edit.tsx
```diff
@@ -23,29 +18,31 @@ import { getProjectSelector, useProjectSelector } from '~/hooks/use-params'
...
   const editProject = useApiMutation('projectUpdate', {
     onSuccess(project) {
       // refetch list of projects in sidebar
-      // TODO: check this invalidation
-      queryClient.invalidateQueries('projectList')
+      queryClient.invalidateEndpoint('projectList')
       // avoid the project fetch when the project page loads since we have the data
-      queryClient.setQueryData('projectView', { path: { project: project.name } }, project)
+      const { queryKey } = projectView({ project: project.name })
+      queryClient.setQueryData(queryKey, project)
```
david-crespo, line 45: I think this is kind of cool. Even though it's two lines, it's much more explicit about what's going on.

app/forms/project-edit.tsx
```diff
@@ -23,29 +18,31 @@ import { getProjectSelector, useProjectSelector } from '~/hooks/use-params'
 import { addToast } from '~/stores/toast'
 import { pb } from '~/util/path-builder'
 
+const projectView = ({ project }: { project: string }) =>
+  apiq('projectView', { path: { project } })
```
david-crespo, line 22: I'm realizing that these functions from selectors to query args are totally generic, likely to be used in multiple places, and could even be generated. In the meantime I'll try collecting some in a file.

david-crespo:
I'm happy with `invalidateEndpoint`. I like the direction, so I'm going to merge. Unfortunately the codebase will continue to be a mix of new and old until everything is converted, but I think it's tolerable.

PR #2598, opened 2024-12-05
Add truncation to displayed IP in CopyableIp

charliepark:
When working on Internet Gateways, I noticed that IPv6 addresses could expand beyond the space we want them to when rendering a CopyableIP. (We could expand this to convert `CopyableIp` to just `Ip`, and have `isCopyable` and `isLinked` props on it, if we decide that makes more sense, and we could then use this more liberally, though I don't know that we have too many uses where the IP is in a space-constrained situation where it doesn't need to be copyable.)

Regardless, here's how it looks:
Broken:
Screenshot 2024-12-05 at 5 05 55 AM
Fixed:
Screenshot 2024-12-05 at 5 15 37 AM

(Sidenote, that copy icon looks 1-2 px too low to me, though if higher, it might read as an "open in external tab" icon?)


david-crespo (approved):
let's go

PR #2600, opened 2024-12-05
TS ESM Tailwind config

benjaminleonard:


tailwind.config.ts
```diff
@@ -78,11 +75,10 @@ module.exports = {
             },
           }))
           .reduce((p, c) => ({ ...p, ...c }), {}),
-        variants
```
benjaminleonard, line 81: Typescript says `variants` is not there?

PR #2606, opened 2024-12-06
refactor: create path params types module, rename current path params to "selectors"

david-crespo:
What we were calling path params were really API selectors, and we weren't exporting the actual path params (with all fields required) so that we could use them in places. This PR renames the existing PP to Sel and creates a new PP that is the real PP.


charliepark (approved):
nice!

PR #2608, opened 2024-12-06
minor: hook up `Dialog.Title` better in modal to fix Radix errors

david-crespo:
Fixed the `DialogTitle` error, which, I was dismayed to see, is not suppressed in production. Maybe for good reason, because it got me to fix it.

Screenshot 2024-12-06 at 2 07 30 PM

Note that the accessible title was actually _not_ broken in production — the problem here is that we were hooking up the title ID to the dialog ourselves rather than letting Radix dialog handle it, so it thought we weren't doing it.

image

### DIFF
--- a/app/ui/lib/Modal.tsx
+++ b/app/ui/lib/Modal.tsx
@@ -8,7 +8,6 @@
 import * as Dialog from '@radix-ui/react-dialog'
 import { animated, useTransition } from '@react-spring/web'
 import cn from 'classnames'
-import React, { forwardRef, useId } from 'react'
 import type { MergeExclusive } from 'type-fest'
 
 import { Close12Icon } from '@oxide/design-system/icons/react'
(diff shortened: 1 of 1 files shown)


app/ui/lib/Modal.tsx
```diff
@@ -85,12 +83,12 @@ export function Modal({
...
                 >
-                  <Dialog.Title asChild>
-                    <ModalTitle id={titleId}>{title}</ModalTitle>
+                  <Dialog.Title className="border-b px-4 py-4 text-sans-semi-lg bg-secondary border-b-secondary">
+                    {title}
                   </Dialog.Title>
                   {children}
                   <Dialog.Close
-                    className="absolute right-2 top-4 flex items-center justify-center rounded p-2 hover:bg-hover"
+                    className="absolute right-2 top-3.5 flex items-center justify-center rounded p-2 hover:bg-hover"
```
david-crespo, line 91: It was very slightly too low.

PR #2617, opened 2024-12-12
Prevent the sidemodal form from immediately erroring when using the action menu

### LINKED ISSUE #2584: When you open a form from the quick menu, you get a field validation error immediately
Somehow the enter keypress to open the form must also be hitting the form and submitting it.

https://github.com/user-attachments/assets/abbac6d9-23fd-4276-8e18-d89cdbe7d982

charliepark:
This adds an `event.preventDefault()` to the Action Menu when the user hits the enter key. This prevents the event from bubbling up and accidentally submitting the form prematurely.

Screen Recording 2024-12-11 at 4 53 04 PM mov

Closes #2584


david-crespo:
I’d put the test in a file dedicated to the action menu. We’ll probably have more of those.

david-crespo (approved):

PR #2618, opened 2024-12-12
Fix instance disk menus closing on poll

david-crespo:
We've broken this several times by accident, so this time I added a test. Very straightforward.


david-crespo:
I was going to add tests for instance list and instance networking tab but we can do that later since they're not broken.

PR #2625, opened 2024-12-14
Lighten solid neutral badge bg

david-crespo:
I find it a little dim.

## Before

Screenshot 2024-12-13 at 9 53 49 PM

## After

Screenshot 2024-12-13 at 9 53 37 PM

I think it makes #2624 look better too, that's why I thought of it.

image


david-crespo:
On my phone, the screenshot of the dim one looks all right, but my less beautiful external monitor is probably more typical.

benjaminleonard (approved):

Issue #994, opened 2022-06-29
Add minitable of non-editable fields to edit modals


just-be-dev:
I've noticed in the designs that we have minitable summaries at the top of edit forms that we currently don't have implemented. We should add these?

![image](https://user-images.githubusercontent.com/3087225/176338310-8882add3-41d4-4f1e-83be-e3ec3e02d57a.png)
![image](https://user-images.githubusercontent.com/3087225/176338343-d38789ba-b939-484a-bffb-67d1d5ed2e6a.png)

david-crespo:
This is still missing on almost all the edit forms (except for the fake ones like [image edit](https://oxide-console-preview.vercel.app/projects/mock-project/images/image-1/edit), ironically). I do think it would be cool to show the fields that are not editable: ID, timestamps, and then other stuff relevant to the thing like IP address and attached instance on a floating IP.

david-crespo:
As discussed on #2630, it's a bit noisy and can distract a bit from the form fields — to consider this done I'd want to take a pass at the design to a) separate the table from the fields better, b) shrink the table a bit or collapse it (Ben doesn't like that) or maybe just take out a couple of rows (like updated at).

PR #2630, opened 2024-12-18
Add minitable to Floating IP side modal

charliepark:
Back in #994, Justin noted the absence of some resource details at the top of side modals. (That is, non-editable details that might pertain to the editable fields of the form.) This PR just picks up the Floating IP side modal and adds the relevant info.

Screenshot 2024-12-18 at 2 23 34 PM

Screenshot 2024-12-18 at 3 38 30 PM

Well, hm. After adding all of the data (id, dates), that's an awful lot of forehead on the form. I wonder if we should put it in a pre-collapsed, expandable element

Happy to pick up more, but figured this was a tightly-scoped one.


david-crespo:
Yeah, I had the same reaction about it being a lot of noise. Collapsing it is an interesting idea. Ben will have to take a look.

benjaminleonard:
I'm not convinced making it collapsible is an improvement. I like to avoid tucking too much away because discoverability is harder and scrolling is less effort than clicking.

If we did modify this, it might be to pull the timestamps and the ID out ... possibly, but would need to play with it. I'd suggest leaving it, and if it gets worse we can reconsider.

david-crespo (approved):

PR #2631, opened 2024-12-18
Use EmptyCell when we'd have an empty cell in InstanceLinkCell

charliepark:
This was something I noticed when working on #2630, but figured I'd handle separately, as it isn't relevant to that PR.

This PR just modifies the InstanceLinkCell component to use an EmptyCell when appropriate.

Here's the before:
Screenshot 2024-12-18 at 3 46 31 PM

Here's the after:
Screenshot 2024-12-18 at 3 46 03 PM


david-crespo (approved):
nice

PR #2633, opened 2024-12-19
Add PropertiesTable.IdRow and PropertiesTable.DateRow

charliepark:
This is a quick refactor that I'm not wedded to by any means, but thought could clean up some of our PropertiesTables, and act as a pattern for future table rows if we like what it does. 

The UI doesn't change at all, apart from a few places where we had the full ID not wrapped in a `Truncate` (though I suspect we actually want that to be consistent, as it is with this change). Otherwise, the PropertiesTable rows all look as they did before.

In this screenshot, the ID, Created, and Updated rows have been updated.
Screenshot 2024-12-18 at 4 32 15 PM


david-crespo (approved):
Nice. `PropertiesTable` is so verbose.

PR #2634, opened 2025-01-06
Make it harder to forget to submit targets/filters subform

### LINKED ISSUE #2627: Make it harder to forget to submit targets/filters subform
@iliana reports (and I think @charliepark have discussed this before) that it feels like the form is just ignore your target because typed it in but forgot to click Add Target. There is probably something we can do about this. It's probably too much to refuse to submit the overall form if there's unsubmitted stuff in the target or filter field, but we can probably warn the user somehow.

charliepark:
Summary: This PR makes it harder to accidentally drop firewall targets/filters when creating/editing a firewall.

Context: It is not uncommon that a user trying to set up a new firewall rule will accidentally submit the overall form without having fully submitted sub-elements — the target, port filter, or host filter — within the form. You can imagine a user having filled in the Targets part of the form below, then moving their attention to Filters (and beyond), not knowing that they haven't actually added the target to their firewall rule:
Screenshot 2025-01-06 at 1 59 55 PM

This PR updates the firewall rules form so that if any of the three subforms have data entered-but-not-submitted, the overall form has its submit function disabled, with copy in a tooltip that directs the user to the appropriate spot on the form
Screenshot 2025-01-06 at 2 03 40 PM

Still writing a few tests for it.

Closes #2627


david-crespo:
I kind of like it feeling like a validation error and auto-scrolling to the offending field. In the confirmation version, I think you'd be curious what you left hanging, so you'd hit "no I don't want to continue" and go looking for it, which is the same thing with a lot more steps. Unless we put it right in the modal like "here, you dropped this", which is a weird thing to do, because if you do want to go back and add that thing, you still have to close the confirmation and go find it in the form.

Implementation-wise, I think we can do a lot better than manually tracking the subform dirty states: we already have a canonical spot for that that state, namely the subforms.

```ts
function useSubformsDirtyState() {
  return {
    target: !!useWatch({ control: targetControl, name: 'value' }).trim(),
    host: !!useWatch({ control: hostControl, name: 'value' }).trim(),
    port: !!useWatch({ control: portControl, name: 'portRange' }).trim(),
  }
}
```

This would require some refactoring because the `target` and `host` subforms are currently initialized inside a helper component.

https://github.com/oxidecomputer/console/blob/9de391c8b0a890941d87c91896de70f000071e65/app/forms/firewall-rules-common.tsx#L87-L108

This might be a little tough, though, because moving them up one level to `CommonFields`

https://github.com/oxidecomputer/console/blob/9de391c8b0a890941d87c91896de70f000071e65/app/forms/firewall-rules-common.tsx#L299-L301

isn't enough if what we need is to make the data available up here. Manually tracking the state is a reasonable hack but I'd rather just get the hierarchy right and have access to the source of truth where we need it. 

image

It wouldn't be a big deal to init all the subforms here next to the top level `form` and pass them into `CommonFields` as props.

https://github.com/oxidecomputer/console/blob/9de391c8b0a890941d87c91896de70f000071e65/app/forms/firewall-rules-create.tsx#L106

PR #2639, opened 2025-01-08
Upgrade dev dependencies

charliepark:
Just an update of dev dependencies. Nothing major, as it were.


package.json
```diff
@@ -75,57 +75,57 @@
...
+    "tailwindcss": "^3.4.17",
+    "tsx": "^4.19.2",
+    "type-fest": "^4.31.0",
     "typescript": "~5.7.0",
-    "vite": "^6.0.3",
+    "vite": "^6.0.7",
     "vite-plugin-html": "^3.2.2",
-    "vite-tsconfig-paths": "^5.1.3",
+    "vite-tsconfig-paths": "^5.1.4",
     "vitest": "^3.0.0-beta.2"
```
david-crespo, line 129: this doesn't come up automatically because it's not semver, but it looks like you can bump this to beta 4
https://github.com/vitest-dev/vitest/releases/tag/v3.0.0-beta.4

david-crespo (approved):

PR #2644, opened 2025-01-09
Instance auto restart

### LINKED ISSUE #2376: Cap how long an instance can poll for
If you land on the instance page and it is stuck in a transitional state, I think it could be reasonable to give up after some number of seconds, like 30 or 60. I wouldn't want it to just sit there polling forever if left open in the browser. You can always click the reload button or refresh the whole page to kick it off again.

We could do something fancy like poll every second for 30 seconds, then every 10 seconds for a few more minutes, then every minute for however long, maybe forever.

### LINKED ISSUE #2469: Expose instance auto-restart status in the console
PR oxidecomputer/omicron#6503 implemented automatic restarts of instances in the `Failed` state. This change introduced some additional instance state that should be exposed to users. In particular:

- When a `Failed` instance is automatically restarted, a cooldown timer is started for that instance. If that instance fails again while the cooldown period is still active, it will not be automatically restarted again until the cooldown period has elapsed.
- Some instances may be configured with auto-restart policies that do not permit them to be restarted when they are `Failed`.

New fields were added to the external-API instance message to report state related to automatic restarts. Instances now have an `auto_restart_enabled: boolean` field that indicates if their auto-restart policy permits restarting the instance, and an `auto_restart_cooldown_expiration: string` representing the date and time at which the cooldown period will have completed (allowing the instance to be restarted again). See: https://github.com/oxidecomputer/omicron/blob/45813be40b62167eff75333c410515e8bee24211/openapi/nexus.json#L15094-L15104

This data should probably be exposed to users: if an instance is in the `Failed` state, the user will want to know why it has not yet been automatically restarted, whether it will ever be automatically restarted, and if it will, when that will happen. We probably only need to display this information for instances which are `Failed`. If a `Failed` instance has `auto_restart_enabled` set to `false`, we should tell the user that auto-restart is disabled for that instance. Otherwise, if there is an `auto_restart_cooldown_expiration` timestamp, we should tell the user that the instance will be restarted only after that time. If `auto_restart_enabled` is not false and there is no `auto_restart_cooldown_expiration` timestamp, then the instance will be automatically restarted --- we might want to indicate that as well.

benjaminleonard:
Fixes #2469 
Fixes #2376 

CleanShot 2025-01-09 at 12 03 09

- [x] Validate conditional rendering is correct
- [x] Automatically set `autoRestartEnabled` to match actual API behaviour
- [x] Add icon to design system
- [x] Write docs text
- [x] Add settings items help text
- [x] Show last auto restarted?

I'm sure someone will talk me down from the slow spinning icon when auto restart is enabled.


app/components/DocsPopover.tsx
```diff
@@ -48,8 +48,8 @@ export const DocsPopover = ({ heading, icon, summary, links }: DocsPopoverProps)
         <Info16Icon aria-label="Links to docs" className="shrink-0" />
       </PopoverButton>
       <PopoverPanel
-        // DocsPopoverPanel needed for enter animation
-        className="DocsPopoverPanel z-10 w-96 rounded-lg border bg-raise border-secondary elevation-2"
+        // popover-panel needed for enter animation
+        className="popover-panel z-10 w-96 rounded-lg border bg-raise border-secondary elevation-2"
```
benjaminleonard, line 52: This casing feels like a carry over from radix, and perhaps inconsistent?

app/pages/project/instances/instance/tabs/SettingsTab.tsx
```diff
@@ -0,0 +1,147 @@
...
+            items={[
+              { value: '', label: 'None (Default)' },
+              { value: 'never', label: 'Never' },
+              { value: 'best_effort', label: 'Best effort' },
+            ]}
+            required
+            className="max-w-none"
+          />
+          <FormMeta label="Enabled" helpText="Help">
+            {instance.autoRestartEnabled ? 'True' : 'False'}{' '}
```
benjaminleonard, line 89: With our mock instance, it's possible for the policy to be set to never but `autoRestartEnabled` to be true. In case of the actual API it'll automatically set `autoRestartEnabled` to be false.

https://github.com/oxidecomputer/omicron/blob/f63ed095e744fb8d2383fda6799eb0b2d6dfbd3c/nexus/db-queries/src/db/datastore/instance.rs#L228C26-L239

We might want to modify the mock `InstanceView` handler to match this behaviour.

app/pages/project/instances/instance/tabs/SettingsTab.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const { isDirty } = form.formState
+
+  const onSubmit = form.handleSubmit((values) => {
+    instanceUpdate.mutate({
+      path: { instance: instanceSelector.instance },
+      query: { project: instanceSelector.project },
+      body: {
+        ncpus: instance.ncpus,
+        memory: instance.memory,
+        bootDisk: instance.bootDiskId,
```
benjaminleonard, line 60: This strikes me as an awkward API design. Let's say the user just wants to update the `autoRestartPolicy` and does not include the `bootDisk`, it would then unset the instances boot disk. It also makes updating this in the CLI cumbersome.

Obviously it's outside of the scope of this PR, but thought I'd mention.

app/components/InstanceAutoRestartPopover.tsx
```diff
@@ -0,0 +1,158 @@
...
+            </span>
+            <OpenLink12Icon className="ml-1 translate-y-[1px] text-secondary" />
+          </a>
+        </div>
+      </PopoverPanel>
+    </Popover>
+  )
+}
+
+const AutoRestartIcon12 = ({ className }: { className?: string }) => (
```
benjaminleonard, line 135: todo: Add to design system and import from there

app/pages/project/instances/instance/tabs/SettingsTab.tsx
david-crespo: We've talked about this a lot on the API side (see https://github.com/oxidecomputer/omicron/issues/6406). Most of our PUTs don't work this way, but it makes things simpler and more consistent if they do. Having to specify all the fields every time is of course a downside for clients. The most basic problem with making the fields optional is that if leaving a field out doesn't unset, you still need to be able to unset one if you want, which means you have to have different semantics for "`bootDisk` not present" and `bootDisk: null`, and that's both confusing and not spec compliant.

benjaminleonard: And I suppose having dedicated instance resize and update instance restart policy endpoints adds a bunch of extra code on the API? I was think less optionality and more something that would facilitate (CLI) `instance auto-restart policy enable`.

david-crespo: Yeah, we don't really want to proliferate endpoints like that. My inclination would be to write that command by hand on the CLI side without it being a dedicated endpoint, or better yet come up with a pattern (a flag to `instance update`?) that lets you update one field while telling it to handle the rest for you.

app/pages/project/instances/instance/tabs/SettingsTab.tsx
david-crespo: Yeah, definitely modify the mock handler and link to the API code in a comment.

app/components/InstanceAutoRestartPopover.tsx
```diff
@@ -0,0 +1,158 @@
...
+  policy: InstanceAutoRestartPolicy
+  cooldownExpiration: Date | undefined
+}) => {
+  const instanceSelector = useInstanceSelector()
+  const [now, setNow] = useState(new Date())
+
+  useEffect(() => {
+    const timer = setInterval(() => setNow(new Date()), 1000)
+    return () => clearInterval(timer)
+  }, [])
```
david-crespo, line 55: I'll see about making this less scary

app/pages/project/instances/instance/tabs/SettingsTab.tsx
```diff
@@ -0,0 +1,147 @@
...
+        </SettingsGroup.Header>
+        <SettingsGroup.Body>
+          <ListboxField
+            control={form.control}
+            name="autoRestartPolicy"
+            label="Policy"
+            description="If unconfigured this instance uses the default auto-restart policy, which may or may not allow it to be restarted."
+            placeholder="Default"
+            items={[
+              { value: '', label: 'None (Default)' },
```
hawkw, line 81: I have a slight preference to just say "Default" here, rather than "None (Default)". I don't want to give the incorrect impression that not setting a policy is equivalent to disabling auto-restart when currently, we always default to "Best Effort".

app/pages/project/instances/instance/tabs/SettingsTab.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <SettingsGroup.Header>
+          <SettingsGroup.Title>Auto-restart</SettingsGroup.Title>
+          <p>The auto-restart policy configured for this instance</p>
+        </SettingsGroup.Header>
+        <SettingsGroup.Body>
+          <ListboxField
+            control={form.control}
+            name="autoRestartPolicy"
+            label="Policy"
+            description="If unconfigured this instance uses the default auto-restart policy, which may or may not allow it to be restarted."
```
hawkw, line 78:
```suggestion
            description="If unconfigured, this instance uses the default auto-restart policy, which may or may not allow it to be restarted."
```

hawkw (commented):
This looks lovely! I'm not particularly qualified to review the implementation of the frontend code, but I left a couple very minor suggestions on the text we display in the UI. Thanks for all your work on getting this exposed in the console, it looks great!

app/pages/project/instances/instance/tabs/SettingsTab.tsx
benjaminleonard: Sounds good – I'd included `None` because it makes the API more intuitive to a user that has used the console, but default is clearer.

I suppose the argument for enforcing `auto_restart_policy` not null, and adding a `default` enum on the API is that it means the user is forced to include it at create time, when for most users they shouldn't need to touch this?

app/components/InstanceAutoRestartPopover.tsx
benjaminleonard: I swapped for `useInterval`

david-crespo:
Not enough contrast to see that `failed` is highlighted IMO. (Copy is tweaked, I'm making it shorter and avoiding saying "control plane") I also expected to see a hover state on the link.

image

david-crespo:
The spinning icon is ok, but it doesn't feel quite enough like waiting for me. It feels more like a loading spinner, like something is actually happening.

david-crespo:
If I change the policy to never, hit save, and change back to default, I get a validation error.

image

app/pages/project/instances/instance/tabs/SettingsTab.tsx
```diff
@@ -0,0 +1,166 @@
...
+            )}
+          </FormMeta>
+        </SettingsGroup.Body>
+        <SettingsGroup.Footer>
+          <div>
+            <LearnMore text="Auto-Restart" href={links.sshDocs} />
+          </div>
+          <Button size="sm" type="submit" disabled={!isDirty}>
+            Save
+          </Button>
```
david-crespo, line 138: I think the relationship between the select and the table underneath could be clearer. The fact that the save button is at the bottom might be the problem. It makes it feel like the preview is inside the form and should therefore update live as a preview of what I'm saving. If the save button was next to the input and the state was below, it might be clearer that it would only update after I hit save.

david-crespo:
I opened with the criticisms but overall this is beautiful. My suggestions will just be small changes.

david-crespo:
I have a fix for setting the value to default in the form and I'm working on e2e tests and some other tweaks.

app/pages/project/instances/instance/tabs/SettingsTab.tsx
```diff
@@ -0,0 +1,189 @@
...
+        <SettingsGroup.Header>
+          <SettingsGroup.Title>Auto-restart</SettingsGroup.Title>
+          <p>The auto-restart policy for this instance</p>
+        </SettingsGroup.Header>
+        <SettingsGroup.Body>
+          <ListboxField
+            control={form.control}
+            name="autoRestartPolicy"
+            label="Policy"
+            description="If unconfigured, this instance uses the default auto-restart policy, which may or may not allow it to be restarted."
```
david-crespo, line 103: Given that we present an "unconfigured" state (value `undefined`) as Default in the picker, this copy is confusing because from the user's point of view there is no unconfigured state.

Currently there is no ability to set a default at project or silo level, so in practice `default` also means best effort. Rather than future-proofing the copy, I think I'd rather we tell the user what is the case now, like we did for the instance update request body docs in [omicron#6959](https://github.com/oxidecomputer/omicron/pull/6959/files?diff=unified&w=0#diff-aee01ef600c8eb547421397e5d89c41a67dd6e6ca6033676cac47b7909e1e344R1146-R1151):

```rs
/// Currently, the global default auto-restart policy is "best-effort", so
/// instances with `null` auto-restart policies will be automatically
/// restarted. However, in the future, the default policy may be
/// configurable through other mechanisms, such as on a per-project basis.
/// In that case, any configured default policy will be used if this is
/// `null`.
```

david-crespo: I almost want to eliminate the third option and just present `Never` and either `Best effort` or `Default`. The problem is deciding whether to keep Default/`undefined` or `best_effort` as the non-never value. The practical difference is that if/when we add configurable defaults in the future, an instance set to default will of course inherit that default, while an instance explicitly set to `best_effort` will override that default. It's hard for me to imagine a user caring either way when they don't even know whether we'll ever introduce a higher-level default, and even if they do care, it's hard for me to imagine them making that decision on behalf of their future selves with much confidence.

This is kind of tough, but I think we're stuck with presenting all three options and giving all the info. The existing docs are undercooked, at least in the [instance update guide](https://docs.oxide.computer/guides/managing-instances#_update_instances). 

The [API docs](https://docs.oxide.computer/api/instance_update) are better, but nobody is going to get there from the console unless we link straight to it. Which I guess we could do? We could stand to have fragment links for individual fields on an API response.

image

app/pages/project/instances/instance/tabs/SettingsTab.tsx
benjaminleonard: I think that's fair, though I'd prefer not to move the save button. This is the UX pattern for this settings form block universally, and it'll be a bit strange to do something different.

Perhaps I can try something like a treatment on the the enabled bit whilst the form has been changed. Cooldown expiration and auto-restart timestamp should not be affected by the policy change?

benjaminleonard: CleanShot 2025-02-03 at 11 03 14

Gave this a try. It might read as the policy is disabling the "enabled" field?

The best UX would be for us to infer what the field would be, but that might get a bit dicey. Especially if the behaviour changes in the future.

benjaminleonard: Alternatively, we know based on the policy, it's either: true, false or project default. It's the inferred `autoRestartEnabled` that we're not sure of until the form is saved. So perhaps we hide the inferred value until it's submitted.

CleanShot 2025-02-03 at 11 21 39

david-crespo: Neither of these solves the problem I ran into, which is that I didn't feel confident about whether/when my change is actually applied. Dimmed does look disabled. I don't think we have a pattern like this, but dotted line border or underline could make me think temporary/provisional. Another way would be to put the dropdown in a modal, but that feels very silly considering we have this perfect spot. What about just getting rid of the `enabled` part ? It's redundant with the listbox. Then all you need is a way of visually marking an unsaved change to the listbox. ...and I guess something to do with cooldown expiration while the listbox is changed to never but not saved. Kinda weird but figuring out a pattern here might be handy for inline editable table rows.

app/pages/project/instances/instance/InstancePage.tsx
```diff
@@ -203,15 +220,13 @@ export function InstancePage() {
...
-                  <button type="button">
-                    <Spinner className="ml-2" />
-                  </button>
-                </Tooltip>
-              )}
+              {match(pollReason)
+                .with('transition', () => <PollingSpinner />)
+                .with('cooldown', () => <InstanceAutoRestartPopover instance={instance} />)
+                .with(null, () => null)
+                .exhaustive()}
```
david-crespo, line 229: How did we live without `ts-pattern`

app/api/util.ts
```diff
@@ -99,8 +99,19 @@ const instanceActions = {
...
+  // https://github.com/oxidecomputer/omicron/blob/0c6ab099e/nexus/db-queries/src/db/datastore/instance.rs#L1282-L1288
+  // https://github.com/oxidecomputer/omicron/blob/0c6ab099e/nexus/db-model/src/instance_state.rs#L39-L42
+  resize: ['stopped', 'failed', 'creating'],
+
+  // https://github.com/oxidecomputer/omicron/blob/0c6ab099e/nexus/db-queries/src/db/datastore/instance.rs#L1209-L1215
+  updateBootDisk: ['stopped', 'failed', 'creating'],
+
+  // there are no state restrictions on setting the auto restart policy, so we
+  // don't have a helper for it
+  // https://github.com/oxidecomputer/omicron/blob/0c6ab099e/nexus/db-queries/src/db/datastore/instance.rs#L1050-L1058
```
hawkw, line 114: :+1:

PR #2646, opened 2025-01-09
Updated – Switching from `@react-spring/web` to `motion`

benjaminleonard:
New PR based on #2214 

> These libraries are very similar, but in places I prefer the `framer-motion` API. One such case is that you can provide it an apple style `duration` and `bounce` and still get a spring animation but without needing to contort your brain with friction, mass etc.
> 
> `framer-motion` does have a much bigger bundle but you can keep it down by excluding the features we don't need. The [layout features](https://www.framer.com/motion/layout-animations/) in particular are very powerful but we don't do any big native style transitions to make its inclusion justifiable.


app/main.tsx
```diff
@@ -46,12 +46,15 @@ function render() {
...
     <StrictMode>
       <QueryClientProvider client={queryClient}>
-        <ErrorBoundary>
-          <ConfirmActionModal />
-          <SkipLink id="skip-nav" />
-          <ReduceMotion />
-          <RouterProvider router={router} />
-        </ErrorBoundary>
+        <LazyMotion strict features={domAnimation}>
+          <MotionConfig reducedMotion="user">
```
david-crespo, line 50: this rules

david-crespo:
This all looks good. I noticed that the modal scrim background still fades in. Does that count as motion?

benjaminleonard:
I believe by default it leaves opacity transitions – which from a quick straw poll previously of people with it turned on at the company seems okay to keep.

david-crespo:
The `LazyMotion` thing might not be configured correctly:

```sh
# before
dist/assets/app-yEWnC-Yg.js      1,273.30 kB │ gzip: 392.99 kB │ map: 5,700.71 kB

# after
dist/assets/app-C59FWg1Z.js      1,300.20 kB │ gzip: 400.19 kB │ map: 5,920.83 kB
```

I'll poke at it. Lazy loading might not be that beneficial if it's going to be used on every page anyway, though it's true we don't need it immediately because we never animate on the initial pageload.

david-crespo:
wazaaa. smaller than before this PR

```
dist/assets/motion-features-DEG9_YHT.js       53.12 kB │ gzip:  18.53 kB │ map:   311.32 kB
dist/assets/app-Bg92zneh.js                1,247.03 kB │ gzip: 382.52 kB │ map: 5,622.99 kB
```

david-crespo:
Running into an issue on Vercel that only shows up after c92c32b61c04a081a1554c6cdd2acecea2759eaf. I discovered it with `disable cache` checked, but it seems to happen with it on or off. When I hard refresh, it seems like MSW is not initializing quickly enough, and the API calls are going straight through to Vite, which responds with `index.html` as it does to any request. That response of course fails to parse as JSON, which blows up the whole page, causing it to reload somehow (occasionally it flashes the Something Went Wrong error — that's how I noticed this). At this point everything works, presumably because MSW has spun up.

In the video, the way this manifests is that all the requests for the page are made twice. This does not happen on regular refreshes.

https://github.com/user-attachments/assets/ca6e3c06-b17a-4768-932e-826877360919

PR #2659, opened 2025-01-22
Remove line that was accidentally copied over

charliepark:
I accidentally copied this from the heading block above it.


david-crespo (approved):
PRs targeting this typos

PR #2660, opened 2025-01-22
Update default blockSize to 4096 for new blank disks

### LINKED ISSUE #2653: Set default block size to 4k for new blank disks
From @mkeeter 
> Crucible's per-block metadata is 48 bytes per block, so that's an additional 10% overhead for 512-byte blocks (which are the default)

Most cloud images at build on 512 – so 4k as a default would just be for blank disks.

charliepark:
This sets the default block size to 4096 when creating a new blank disk.

Screenshot 2025-01-22 at 10 27 39 AM

Closes #2653


david-crespo (approved):
we stan a one-liner king

PR #2665, opened 2025-01-23
Visual snapshot tests with Playwright

david-crespo:
@benjaminleonard suggested we try some of this to get more confidence around design system changes.


david-crespo:
All 4 chrome tests failed due to colors being different! What the heck.

### Actual

visual-snapshots-projects-1-actual


### Expected
visual-snapshots-projects-1-expected

### Diff

visual-snapshots-projects-1-diff

benjaminleonard:
Reckon – that's a difference between your local environment and the server that is running the CI? Maybe related to P3 colour.

david-crespo:
Ok, now it's all the browsers, which tbh makes slightly more sense.

david-crespo:
I'm not going to be finishing this any time soon.

PR #2668, opened 2025-01-28
tools: refactor bump omicron script, use worktree

david-crespo:
This started out with me just wanting to use a git worktree to make the console bump in omicron in order to avoid messing with my omicron working directory, but I found the script confusing to refactor, so I took the opportunity to reorganize it a bit. The new `run` function wrapping the main logic shows relatively clearly what the script does.

I also added a positional arg that lets you specify a console commit to bump to instead of always using `HEAD`.


tools/deno/bump-omicron.ts
```diff
@@ -56,29 +55,28 @@ function linkifyGitLog(line: string): string {
...
-if (args.help) {
-  console.info(HELP)
-  Deno.exit()
+  return {
+    dir: tmpDir,
+    [Symbol.asyncDispose]: async function () {
+      console.log('Cleaning up worktree')
+      await $`git worktree remove ${tmpDir}`.cwd(OMICRON_DIR).quiet('stdout')
+    },
+  }
```
david-crespo, line 68: this is really cool
https://www.totaltypescript.com/typescript-5-2-new-keyword-using

PR #2671, opened 2025-01-31
Better error for admin group name misconfig

### LINKED ISSUE #2573: "Something went wrong" warning on login with no roles is too generic
Users logging into a silo for the first time can be met with a generic "something went wrong" warning, which seems to be the defacto warning for misconfigured role assignment.

Opening this issue to request a more clear warning to the user to prevent confusion if they're not yet authorized (i.e. user not added to the group that was assigned to the silo admin group, etc).

david-crespo:
Closes #2573 

Obviously the copy could be shorter, but in theory this error should only show up to operators immediately after IdP setup and virtually never again.

Doc link goes to: https://docs.oxide.computer/guides/operator/faq#_how_do_i_fix_the_something_went_wrong_error, which is added in [v13](https://github.com/oxidecomputer/docs/pull/513), so it won't work until that's merged.

image

### DIFF
--- a/app/components/ErrorBoundary.tsx
+++ b/app/components/ErrorBoundary.tsx
@@ -12,6 +12,7 @@ import { useRouteError } from 'react-router'
 import { apiq } from '~/api'
 import { type ApiError } from '~/api/errors'
 import { Message } from '~/ui/lib/Message'
+import { links } from '~/util/links'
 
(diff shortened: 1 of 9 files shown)


david-crespo:
Amazing (derogatory). It went back to flaking for the original reason, image upload cancel and retry in chrome.

PR #2674, opened 2025-02-04
Improve image upload cleanup logic, reduce e2e flake

david-crespo:
Trying to fix some flakes that came up in #2671 (and periodically for a very long time). I think we were calling cleanup twice more often than I thought.

### DIFF
--- a/app/forms/image-upload.tsx
+++ b/app/forms/image-upload.tsx
@@ -184,12 +184,6 @@ export function Component() {
   const queryClient = useApiQueryClient()
   const { project } = useProjectSelector()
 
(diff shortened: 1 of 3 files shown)


app/forms/image-upload.tsx
```diff
@@ -335,7 +339,6 @@ export function Component() {
       await deleteDiskCleanup.mutateAsync({ path: { disk: disk.current.id } })
       disk.current = null
     }
-    cleaningUp.current = false
```
david-crespo, line 338: Because we turned this back to false when cleanup was done, it did not prevent it from running twice in the case where the first attempt finished before the second started.

PR #2676, opened 2025-02-05
Instance metrics design tweaks

benjaminleonard:


.licenserc.yaml
```diff
@@ -20,5 +20,7 @@ header:
     - 'OMICRON_VERSION'
     - '.husky'
     - 'mockServiceWorker.js'
+    - 'app/util/format-date-range.ts'
+    - 'app/util/format-date-range.test.ts'
```
benjaminleonard, line 24: This lib is MIT licensed. Unsure if there's a better way to handle this – it's so small it feels better to copy it over rather than importing.

app/components/RefetchIntervalPicker.tsx
```diff
@@ -37,9 +37,19 @@ type Props = {
   enabled: boolean
   isLoading: boolean
   fn: () => void
+  showLastFetched?: boolean
+  className?: string
+  isSlim?: boolean
```
benjaminleonard, line 42: This gives us a version that has the dropdown but not the number...perhaps we just scrap the selected interval in the listbox all together here and we can simplify the implementation. I don't think the interval necessarily needs to be visible anywhere.

app/components/RouteTabs.tsx
```diff
@@ -39,20 +39,30 @@ export interface RouteTabsProps {
...
+  fullWidth,
+  sideTabs = false,
+  tabListClassName,
+}: RouteTabsProps) {
   const wrapperClasses = sideTabs
     ? 'ox-side-tabs flex'
     : cn('ox-tabs', { 'full-width': fullWidth })
   const tabListClasses = sideTabs ? 'ox-side-tabs-list' : 'ox-tabs-list'
-  const panelClasses = cn('ox-tabs-panel', { 'flex-grow': sideTabs })
+  const panelClasses = cn('ox-tabs-panel @container', { 'ml-5 flex-grow': sideTabs })
```
benjaminleonard, line 57: Container query lets us move the datepicker onto the next line when we start running out of space. Flex wrap might work here, but eventually I think I'd like to do something cleverer with the layout when it reflows.

app/components/TimeSeriesChart.tsx
```diff
@@ -153,9 +157,22 @@ export default function TimeSeriesChart({
   // re-render on every render of the parent when the data is undefined
   const data = useMemo(() => rawData || [], [rawData])
 
+  if (!data || data.length === 0) {
```
benjaminleonard, line 160: Temporary empty state. We should handle errors and loading differently.

app/ui/lib/DateRangePicker.tsx
```diff
@@ -68,8 +64,8 @@ export function DateRangePicker(props: DateRangePickerProps) {
               : 'border-default ring-accent-secondary'
           )}
         >
-          <div className={cn('relative flex w-[17rem] items-center px-3 text-sans-md')}>
-            {label}
+          <div className="relative flex w-[16rem] items-center px-3 text-sans-md">
+            <div className="truncate">{label}</div>
```
benjaminleonard, line 68: This is useful more generally to stop the date picker text from wrapping

tailwind.config.ts
```diff
@@ -71,5 +72,6 @@ export default {
       addUtilities(colorUtilities)
       addUtilities(elevationUtilities)
     }),
+    containerQueriesPlugin,
```
benjaminleonard, line 75: Not necessary after we upgrade to Tailwind v4.

david-crespo:
Design changes look good. I think I have to veto `little-date` — it was made for a hackathon and probably hasn't been worked on further, so it's not very polished and has some weird behaviors that I don't think fit, like if you select all of January precisely it just says `January 2025` in the label. I'd rather have it say the range and not abbreviate. The bundle size by 1% gz (4kb) over sticking with the react-aria thing.

### little-date edge cases

https://github.com/user-attachments/assets/93c57d2b-ae85-4463-8f3a-d2d514c2a54b

### react-aria `useDateFormatter`

The `react-aria` thing is a little ugly but I think it's more consistent and reliable.

https://github.com/user-attachments/assets/f4fa6c37-8f7f-4e76-86aa-a2b2e7f9cccc

### react-aria wtf (fixed above)

Turns out `h24` here is not right, it gets you 24:10 instead of 00:10. Changing that to `h23` gets us what we want. No idea why that option is even in there.

https://github.com/oxidecomputer/console/blob/93c5abc92b386145d0019246c449f09a9abdc375/app/ui/lib/DateRangePicker.tsx#L36-L40

https://github.com/user-attachments/assets/e7392a1f-e070-456e-9ba1-0add2380b109

2025-02-07-hour-cycle

PR #2678, opened 2025-02-06
Fetch all-ish silos on utilization page

### LINKED ISSUE #2677: Utilization page only retrieves first 100 silos
I only noticed this once the `oxide` silo was pushed out of the result list as our utilization metrics dropped dramatically. This affects both the utilization view at the top of the page as well as the summary view below.

Instead we probably need to follow the pagination to collect all of the results (or use something like OxQL for the summary boxes).

david-crespo:
Closes #2677 

The alternative would be to paginate, but that is annoying, plus it doesn't cover the metrics silo listbox — that still has to fetch all.

### DIFF
--- a/app/pages/system/UtilizationPage.tsx
+++ b/app/pages/system/UtilizationPage.tsx
@@ -10,10 +10,11 @@ import { useIsFetching } from '@tanstack/react-query'
 import { useMemo, useState } from 'react'
 
 import {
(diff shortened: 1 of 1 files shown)


augustuswm (approved):
I think this is sufficient for now. Especially since we are probably the only one that this affects.

PR #2697, opened 2025-02-17
Dynamic chart Y axis width

benjaminleonard:
Seems like a small change but it makes them feel much more resolved in the container.
Comments explain it pretty well, but essentially we can figure out the container width by taking the width of a character – which we know – and adding the padding and tick stroke size. Recharts does not have a way of doing this natively.

It is possible to calculate the width of a string, given the string and the text using the canvas – which is dynamic of course but probably overkill in this instance. We might need to re-evaluate for mobile layouts eventually.

Before:
image

After:
image


david-crespo (approved):
Subtle, yet exquisite

PR #2698, opened 2025-02-17
Metrics error & loading states

benjaminleonard:
Still WIP.

Padding is funky on the non instance metrics page. Error screen flashes if it auto-refreshes and finds it errored again.

CleanShot 2025-02-17 at 18 20 25
CleanShot 2025-02-17 at 17 49 41

### DIFF
--- a/app/components/TimeSeriesChart.tsx
+++ b/app/components/TimeSeriesChart.tsx
@@ -7,7 +7,7 @@
  */
 import cn from 'classnames'
 import { format } from 'date-fns'
-import { useMemo } from 'react'
+import { useMemo, type ReactNode } from 'react'
 import {
   Area,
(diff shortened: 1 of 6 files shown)


benjaminleonard:
Let me poke at it some more, still some padding fixes and potentially that error flash issue. But it's not far off.

benjaminleonard:
Made some small tweaks and think it's ready to merge. Also fixed the lack of a load state on the interval picker. Though we do get lots of spinners at once on first load. Might be worth using something other than a spinner on the skeleton state instead. Not a blocker, I'll give it some thought.

benjaminleonard:
Switched it to animated bars instead. Let me know what you think:

CleanShot 2025-02-25 at 12 20 44

PR #2699, opened 2025-02-17
React 19

david-crespo:
Redoing #2636 without needing to use `overrides` in `package.json` by updating some deps and fixing the react peer deps in `@oxide/design-system` (https://github.com/oxidecomputer/design-system/pull/118) and `@oxide/react-asciidoc` (https://github.com/oxidecomputer/react-asciidoc/pull/32).

Most of our type errors are due to `ReactElement`'s props generic changing from `any` to `unknown`. See https://react.dev/blog/2024/04/25/react-19-upgrade-guide#changes-to-the-reactelement-typescript-type. This is clearly more correct, and forces us to fix some things we were doing unsafely, but it is annoying.

- [x] Get `npm install` to not blow up
- [x] Fix a couple of failing e2es
- [ ] Fix type errors 
- [ ] Finish and merge https://github.com/oxidecomputer/react-asciidoc/pull/32
- [ ] Finish and merge https://github.com/oxidecomputer/design-system/pull/118


patches/@radix-ui+react-focus-guards+1.0.1.patch
```diff
@@ -1,18 +0,0 @@
...
-     element.setAttribute('data-radix-focus-guard', '');
-     element.tabIndex = 0;
--    element.style.cssText = 'outline: none; opacity: 0; position: fixed; pointer-events: none';
-+    element.style.outline = 'none';
-+    element.style.opacity = '0';
-+    element.style.position = 'fixed';
-+    element.style.pointerEvents = 'none';
-     return element;
- }
- const $3db38b7d1fb3fe6a$export$be92b6f5f03c0fe9 = $3db38b7d1fb3fe6a$export$ac5b58043b79449b;
```
david-crespo, line 18: This patch isn't needed anymore because we now depend on the version of this library that incorporated our [patch](https://github.com/radix-ui/primitives/pull/2840). 

```
$ npm ls @radix-ui/react-focus-guards
oxide@0.0.0 /Users/david/oxide/console
└─┬ @radix-ui/react-dialog@1.1.6
  └── @radix-ui/react-focus-guards@1.1.1
```

patches/react-remove-scroll+2.6.3.patch
```diff
@@ -34,35 +41,17 @@ index 08eda83..e48ccd6 100644
...
- import { effectCar } from './medium';
- var nothing = function () {
-@@ -29,8 +28,4 @@ RemoveScroll.defaultProps = {
-     removeScrollBar: true,
-     inert: false,
- };
--RemoveScroll.classNames = {
--    fullWidth: fullWidthClassName,
--    zeroRight: zeroRightClassName,
--};
```
david-crespo, line 67: This existing bit of the patch wasn't necessary. These lines do not set classes, they set a value `RemoveScroll.classNames` that is imported by other things and used for stuff.

david-crespo:
The `Invalid prop data-headlessui-state supplied to React.Fragment` thing is a Headless UI bug that they would love to fix if there was a repro, so I made a repro: https://github.com/tailwindlabs/headlessui/issues/3351#issuecomment-2664183749

app/components/form/fields/DateTimeRangePicker.tsx
```diff
@@ -103,7 +103,7 @@ export function DateTimeRangePicker({
   return (
     <form className="flex">
       <Listbox
-        className="z-10 w-[10rem] border-r border-r-default [&>button]:!rounded-r-none [&>button]:!border-r-0"
+        className="z-10 w-[10rem] border-r border-r-default [&_button]:!rounded-r-none [&_button]:!border-r-0"
```
david-crespo, line 106: This was needed to fix the border between the `Listbox` and the button after adding a wrapping div to work around the Headless bug. We can always change it back if want after they fix it.

image

app/ui/lib/Button.tsx
```diff
@@ -55,80 +55,82 @@ const noop: MouseEventHandler<HTMLButtonElement> = (e) => {
...
+// focus to the button programmatically. There is an example in TlsCertsField
+// in the silo create form: when there are no certs added, the validation error
+// on submit focuses and scrolls to the add TLS cert button. All of that is
+// normal. The confusing part is that when the button is disabled and wrapped
+// in a tooltip, the tooltip component wants to add its own ref to the button
+// so it can figure out where to place the tooltip. In order to make both refs
+// work at the same time (so that, for example, in theory, a button could be
+// simultaneously disabled with a tooltip *and* be focused programmatically [I
+// tested this]), we merge the two refs inside Tooltip, using child.props.ref to
+// get the original ref on the button.
```
david-crespo, line 75: important comment added

Issue #1855, opened 2023-12-13
Allow real links in more actions menu


augustuswm:
We received as feedback from an external user that when using the resource menu to access the serial console, it would be very helpful to be able to right-click > open in new tab.

![image](https://github.com/oxidecomputer/console/assets/1377730/5d235e5e-a911-4567-830d-2aa9f4171a1b)

I have not looked at how these are set up, but it seems like (for those that are appropriate) the menu item should be a real link as opposed to a div.

david-crespo:
Ah yeah, that’s interesting. Obviously most of those aren’t links, so we’d have to figure out how to mix buttons and links.

PR #2700, opened 2025-02-18
Metrics more actions

benjaminleonard:
Adding for parity with the Figma designs:
image

---

Splits the copy code modal and the button so that the modal can be called without a regular button.

It might be simpler to just remove the button part completely. Will defer to you guys.

The docs link thing is placeholder. There's two potential improvements that probably should be made.

1. We adapt the actions menu to take links (both internal and external` that use the `Dropdown.LinkItem`. This has the benefit of making the other links in these menus openable with a command click
2. The way we link to docs is potentially brittle. Want some way of validating that these sections do in fact exist. But a hash docs link won't 404.

### DIFF
--- a/app/util/links.ts
+++ b/app/util/links.ts
@@ -33,6 +33,9 @@ export const links = {
   keyConceptsProjectsDocs:
     'https://docs.oxide.computer/guides/key-entities-and-concepts#_projects',
   oxqlDocs: 'https://docs.oxide.computer/guides/operator/system-metrics#_oxql_quickstart',
+  // TODO: update URL once https://github.com/oxidecomputer/docs/pull/426 is merged
+  oxqlSchemaDocs: (metric: string) =>
+    `https://docs-git-timeseries-guide-oxidecomputer.vercel.app/guides/operator/available-metric-data#_${metric.replace(':', '')}`,
   projectsDocs: 'https://docs.oxide.computer/guides/onboarding-projects',
   quickStart: 'https://docs.oxide.computer/guides/quickstart',
   routersDocs:
--- a/app/components/CopyCode.tsx
+++ b/app/components/CopyCode.tsx
@@ -5,6 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
+import { motion as m } from 'framer-motion'
 import { useState, type ReactNode } from 'react'
 
 import { Success12Icon } from '@oxide/design-system/icons/react'
@@ -15,7 +16,6 @@ import { useTimeout } from '~/ui/lib/use-timeout'
 
 type CopyCodeProps = {
   code: string
-  modalButtonText: string
   copyButtonText: string
   modalTitle: string
   footer?: ReactNode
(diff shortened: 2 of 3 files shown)


david-crespo:
1 is #1855, which I've been meaning to get to. I don't think it's particularly hard.

app/util/links.ts
```diff
@@ -33,6 +33,9 @@ export const links = {
   keyConceptsProjectsDocs:
     'https://docs.oxide.computer/guides/key-entities-and-concepts#_projects',
   oxqlDocs: 'https://docs.oxide.computer/guides/operator/system-metrics#_oxql_quickstart',
+  // TODO: update URL once https://github.com/oxidecomputer/docs/pull/426 is merged
+  oxqlSchemaDocs: (metric: string) =>
+    `https://docs-git-timeseries-guide-oxidecomputer.vercel.app/guides/operator/available-metric-data#_${metric.replace(':', '')}`,
```
david-crespo, line 38: Couple things about this. I like the idea of linking to a doc, though I'm not sure the schema is that helpful in understanding the graph.

I'm not that worried about us breaking these links. We can pretty easily test them directly if we want: have playwright click the link and go to the site and see whether the right thing is in the viewport. Speaking of which, it seems like the site is kind of janky around that, possibly a timing thing due to the page being so long. If I open this link in a new tab, I see it briefly at the right part of the page before jump back up to the top, presumably when something else loads or hydrates or whatever. If I hit enter in the URL bar again, it jumps to the right place.

https://docs-git-timeseries-guide-oxidecomputer.vercel.app/guides/operator/available-metric-data#_virtual_machinevcpu_usage

Related to that, it seems to cut off the heading. Other fragment links on other pages don't seem to have this problem. It could be because they render a little differently because they have the `[discrete]` tag on them in the asciidoc to keep them out of the sidebar ToC.

image


image

david-crespo: Eh, I take it back about it not being helpful — seeing that description of what this value is is pretty helpful. I do think we should consider tweaking the copy and maybe the visual hierarchy to make it easier to parse all that info. Metadata (a name I put in there) is not that great, it's actually almost the opposite! Fields are the metadata, and that table actually describes the data. I will work on it in the docs site PR.

david-crespo:
Trying to think of more specific copy than "docs" that really says "what even is this data". I think it could use an external link icon. I can make that part of the solution in #2343 — distinguish three ways between non-links, local links, and external links. The first two would look the same I think, latter would have the icon.

image

david-crespo:
Merging and we can do the followup work in the main PR.

benjaminleonard:
The discrete link thing I'm planning to fix. Basically it's missing the anchor tag with the offset applied. Need to dig into the way discrete should behave. We only really care that it's not in the ToC.

app/util/links.ts
benjaminleonard: Fixed the discrete thing and hopefully the link to the sections works better now. There was a hydration error due to asciidoc shenanigans.

Issue #2695, opened 2025-02-16
Fix HMR


david-crespo:
HMR doesn’t work at all. Totally my fault: https://github.com/oxidecomputer/console/pull/2545 exports components and loaders from the same file, which breaks HMR. Those route files are all the top level files and HMR is broken for any changes below them, so it’s broken for all changes. This is annoying because it takes a few seconds to see a change instead of like 50ms, and it refreshes the page, losing any client-side state.

That change was meant to prepare us for RR v7 in framework mode + SPA mode, but I’m not sure how feasible that is for us. So we need to determine whether we plan to move to that, and if not, switch all the route files back to sticking loader on `PageComponent.loader`.

david-crespo:
RRv7 in framework may or may not fix this. As their [docs](https://github.com/remix-run/react-router/blob/fbe796a4c5f257d710b2f779235e0cbcaef25396/docs/explanation/hot-module-replacement.md#supported-exports) point out, the Vite plugin removes exports like `loader` and `action` from the module. But it doesn't say anything about `clientLoader`. So this may require [`unstable_splitRouteModules`](https://github.com/remix-run/react-router/blob/fbe796a4c5f257d710b2f779235e0cbcaef25396/docs/explanation/code-splitting.md) to split the `clientLoader` off from the component export.

PR #2702, opened 2025-02-19
chore: Start converting to RRv7 framework mode

david-crespo:
This is the beginning of both #2019 and #2695.

Following this helpful guide https://reactrouter.com/upgrading/router-provider, we

1. Create a `convert` function to convert from framework mode route modules with a default export and a `clientLoader` to what library mode wants, namely a named `Component` export and a `loader`.
2. Convert existing `<Route>` to use `lazy={() => import('./path/to/route')}`

For this PR we can leave it at that, but eventually we will

3. Convert all the route modules to work that way
4. Shuffle files around and convert `routes.tsx` to a framework mode `route.ts` and flip on framework mode by using the RR Vite plugin.

In this PR we also make a couple of other tweaks:

* Allow previously prohibited default exports in the lint config, but only in `app/pages` and `app/layouts`
* Set [`output.experimentalMinChunkSize`](https://rollupjs.org/configuration-options/#output-experimentalminchunksize) in the Rollup config so we don't get a ton of 300 byte JS chunks.

#### With `output.experimentalMinChunkSize`

image

#### Without `output.experimentalMinChunkSize`

image


app/pages/project/instances/instance/SerialConsolePage.tsx
```diff
@@ -20,13 +20,12 @@ import { PrevArrow12Icon } from '@oxide/design-system/icons/react'
...
 import { EquivalentCliCommand } from '~/components/CopyCode'
 import { InstanceStateBadge } from '~/components/StateBadge'
+import { Terminal } from '~/components/Terminal'
 import { getInstanceSelector, useInstanceSelector } from '~/hooks/use-params'
 import { Badge, type BadgeColor } from '~/ui/lib/Badge'
 import { Spinner } from '~/ui/lib/Spinner'
 import { pb } from '~/util/path-builder'
 
-const Terminal = lazy(() => import('~/components/Terminal'))
-
```
david-crespo, line 29: Nice example of what is happening here: we get this for free through the route being split.

david-crespo:
This can be merged as-is, we can convert the rest of the routes gradually.

app/layouts/ProjectLayoutBase.tsx
```diff
@@ -0,0 +1,123 @@
...
+          </NavLinkItem>
+          <NavLinkItem to={pb.projectAccess(projectSelector)}>
+            <Access16Icon title="Access" /> Access
+          </NavLinkItem>
+        </Sidebar.Nav>
+      </Sidebar>
+      {overrideContentPane || <ContentPane />}
+    </PageContainer>
+  )
+}
```
david-crespo, line 123: This is just `ProjectLayout` moved in its entirety. The diff doesn’t show as a rename because the original file is still there, it’s just very short now.

PR #2707, opened 2025-02-22
Slight refactor to pluralization util; use with vCPUs

charliepark:
It's rare that someone would have just 1 vCPU, but we should handle it when they do.

Before:
Screenshot 2025-02-21 at 5 37 31 PM

After:
Screenshot 2025-02-21 at 5 37 41 PM

I also noticed that we weren't consistent in using `vCPU` or `vCPUs` (these had all said `vCPU`, as in `4 vCPU`), so I pulled the updated pluralization in to the instances overview page:
Screenshot 2025-02-21 at 5 45 53 PM

### DIFF
--- a/app/util/str.spec.tsx
+++ b/app/util/str.spec.tsx
@@ -60,8 +60,9 @@ it('commaSeries', () => {
   expect(commaSeries([], 'or')).toBe('')
   expect(commaSeries(['a'], 'or')).toBe('a')
   expect(commaSeries(['a', 'b'], 'or')).toBe('a or b')
-  expect(commaSeries(['a', 'b'], 'or')).toBe('a or b')
+  expect(commaSeries(['a', 'b'], 'and')).toBe('a and b')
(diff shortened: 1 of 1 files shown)


app/util/str.ts
```diff
@@ -10,7 +10,10 @@ import React, { type ReactElement, type ReactNode } from 'react'
 
 export const capitalize = (s: string) => s && s.charAt(0).toUpperCase() + s.slice(1)
 
-export const pluralize = (s: string, n: number) => `${n} ${s}${n === 1 ? '' : 's'}`
+export const pluralize = (word: string, count: number, stringOnly = false) => {
+  const pluralizedWord = count === 1 ? word : `${word}s`
+  return stringOnly ? pluralizedWord : `${count} ${pluralizedWord}`
+}
```
david-crespo, line 16: you could just drop the count. there are no other uses of this helper

david-crespo (approved):

PR #2710, opened 2025-02-27
Remove restrictions on name field entries

### LINKED ISSUE #2672: Don't restrict typing in name fields to valid characters
Added in #2512, tweaked in #2549.

Issues raised by @taspelund and @cbiffle in [chat](https://matrix.to/#/!rUuGIIGuojsCiHirYz:oxide.computer/$Hoa5OBaQdD9CTrSNyxapaYaIrMxR3nYXnxIG_xWwYZI?via=oxide.computer&via=unix.house&via=matrix.org):

1. No way to know what the rules are if you can't submit something invalid and see the error (see #1792 re: tooltip)
1. Very confusing that you can type `-` sometimes (after other characters) but not other times (when field is empty)
1. Pasting something with invalid characters filters those out but leave the rest, so you can end up with a different name than you thought if you're not paying close attention

charliepark:
Removes a handful of helper functions that would restrict the characters entered into the Name field. Folks on the team raised a few concerns (outlined in #2672) about modifying the user's entered info, possibly without their noticing.

This does not remove the validation on the text, so they'll still get warnings if they try to submit non-conforming strings.

Closes #2672


test/e2e/project-create.e2e.ts
```diff
@@ -31,8 +31,6 @@ test.describe('Project create', () => {
 
   test('shows field-level validation error and does not POST', async ({ page }) => {
     const input = page.getByRole('textbox', { name: 'Name' })
-    await input.pressSequentially('no sPoNgEbOb_CaSe or spaces')
-    await expect(input).toHaveValue('no-spongebob-case-or-spaces')
```
david-crespo, line 35: Would be good to add a couple of types of bad input here and assert about the errors. I think we have different messages depending on what's wrong. There was one case removed when the normalization was added, but might as well do 2 or 3.

https://github.com/oxidecomputer/console/pull/2512/files#diff-0b899f6668600830b3fd43d75fa1fabec22eb556702ad569b281c8a9e7b8c47f

david-crespo (approved):

PR #2715, opened 2025-02-28
chore: get rid of `forwardRef` thanks to React 19

david-crespo:
The diff is about 1/10 as long with [whitespace hidden](https://github.com/oxidecomputer/console/pull/2715/files?w=1).

### DIFF
--- a/app/components/form/fields/RadioField.tsx
+++ b/app/components/form/fields/RadioField.tsx
@@ -6,7 +6,7 @@
  * Copyright Oxide Computer Company
  */
 import cn from 'classnames'
-import { useId, type default as React } from 'react'
+import { useId } from 'react'
 import {
   useController,
   type Control,
--- a/app/util/classed.ts
+++ b/app/util/classed.ts
@@ -6,7 +6,7 @@
  * Copyright Oxide Computer Company
  */
 import cn from 'classnames'
-import React, { forwardRef, type JSX } from 'react'
+import React, { type JSX } from 'react'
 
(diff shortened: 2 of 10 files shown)


app/components/form/fields/RadioField.tsx
```diff
@@ -6,7 +6,7 @@
  * Copyright Oxide Computer Company
  */
 import cn from 'classnames'
-import { useId, type default as React } from 'react'
+import { useId } from 'react'
```
david-crespo, line 9: I had changed these to this ugly `type default` thing in #2713 to satisfy oxlint complains about type imports, but it turns out that when `React` is used as just a type, it's actually global already and we don't have import it at all.

app/util/classed.ts
```diff
@@ -6,22 +6,16 @@
...
-    const Comp = forwardRef(
-      ({ className, children, ...rest }: JSX.IntrinsicElements[T], ref) =>
-        React.createElement(
-          tag,
-          { className: cn(strings[0], className), ...rest, ref },
-          children
-        )
-    )
+    const Comp = ({ className, ...rest }: JSX.IntrinsicElements[T]) =>
+      React.createElement(tag, { className: cn(strings[0], className), ...rest })
```
david-crespo, line 18: This one is my favorite. I changed the `children` thing from a positional arg to a prop and as far as I can tell that's totally fine. I [confirmed](https://github.com/facebook/react/blob/56c7d1070aabe42b955ea0006477ea1ca02dd0c5/packages/react/src/jsx/ReactJSXElement.js#L641-L801) it in the React source.

PR #2720, opened 2025-03-01
Reduced motion loading bar

david-crespo:
Realized the loading bar motion could be annoying when I myself got annoyed at it. Here is an attempt at a pulsing version for `prefers-reduced-motion` that is still a little in-your-face but at least it doesn't move. Kinda nice! [Diff with indentation changes hidden](https://github.com/oxidecomputer/console/pull/2720/files?w=1) is better.

https://github.com/user-attachments/assets/d3dd4966-da3c-4df9-a4e0-2dfccca3eb6f

### DIFF
--- a/app/ui/styles/components/loading-bar.css
+++ b/app/ui/styles/components/loading-bar.css
@@ -11,48 +11,88 @@
   width: 0%;
 }
 
-.global-loading-bar.loading {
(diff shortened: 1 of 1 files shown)


benjaminleonard:
Looks good

PR #2723, opened 2025-03-03
Update properties table

### LINKED ISSUE #2722: Remove badge from label of properties table
This was always a little bit confusing. New designs use regular `text-mono-sm` instead.

Image

benjaminleonard:
Fixes #2722 

Bonus fix of this joined two column properties table is we automatically get the label columns matching width when it wraps on smaller screens.

Inline grid stuff maybe isn't the most readable, but doing it in a mix of externally in the CSS I always find confusing. Perhaps fixable with some utility classes.

image

Slightly narrower so it doesn't take so much space, especially on the side panes.

Pretty much ready to review, I need to poke around all the usages to check for any regressions.


app/ui/lib/PropertiesTable.tsx
```diff
@@ -35,7 +39,13 @@ export function PropertiesTable({ className, children }: PropertiesTableProps) {
...
       className={cn(
         className,
-        'properties-table grid min-w-min basis-6/12 divide-y rounded-lg border border-default children:pl-3 children:pr-6 children:border-secondary'
+        'properties-table min-w-min basis-6/12 rounded-lg border border-default',
+        'children:border-t children:pl-3 children:pr-6 children:border-secondary [&>*:nth-child(-n+2)]:!border-t-0',
+        'grid grid-cols-[minmax(min-content,1fr)_3fr]',
+        {
+          'lg+:grid-cols-[minmax(min-content,1fr)_3fr_minmax(min-content,1fr)_3fr] lg+:[&>*:nth-child(-n+4)]:!border-t-0 lg+:[&>*:nth-child(4n-2)]:border-r':
+            columns === 2,
+        }
```
david-crespo, line 48: I'll take your word for it

Issue #2721, opened 2025-03-01
Test flake due to async route modules


david-crespo:
Lots of test flake, but interestingly it didn’t really kick off until near the end of the conversion, so hopefully it’s due to something specific.

It only seems to happen in one browser per run, every time. And it affects different tests every time too. I don't get it.

Firefox: https://github.com/oxidecomputer/console/actions/runs/13598150882/job/38019429785
Safari: https://github.com/oxidecomputer/console/actions/runs/13635164967/job/38112041124?pr=2724
Firefox: https://github.com/oxidecomputer/console/actions/runs/13597704894/job/38018130878
Chrome: https://github.com/oxidecomputer/console/actions/runs/13609558382/job/38045006012?pr=2720

Ok finally had two browsers fail in one run: https://github.com/oxidecomputer/console/actions/runs/13635221035

david-crespo:
#2729 seems to improve things a lot. Will close if it improves things to a tolerable degree.

PR #2725, opened 2025-03-03
chore: Playwright 1.51 beta

david-crespo:
Most likely will not improve #2721 but I'm curious.


david-crespo:
Still failed, as expected. One more data point.

Issue #2704, opened 2025-02-19
Test flake: firewall rule form targets table


david-crespo:
Never seen this before. It seems like the enter key press in the combobox passed through to submit the form as a whole, which threw everything off.

db94deab5d9a53e859230c6b66f3139718a13591

https://github.com/oxidecomputer/console/actions/runs/13424227374/job/37503809801

I get failures about 1 or 2 out of 10 runs with:

```
npx playwright test --project=firefox -g 'form targets table' --repeat-each 10
```

<img width="1453" alt="Image" src="https://github.com/user-attachments/assets/dea77fc2-119d-43b7-a7ed-a6aadcfc861c" />

david-crespo:
#2705 seems to have improved it but not fixed it.

https://github.com/oxidecomputer/console/actions/runs/13443453825/job/37563220513

david-crespo:
Added a second sleep in f9931e03 and I think I got it. Will close in a bit if we stop seeing the failure. I failed to notice that the failures where about the _other_ spot in the test where we press enter twice in succession.

david-crespo:
I think we got it.

david-crespo:
Not quite. Got one here.
https://github.com/oxidecomputer/console/actions/runs/14012768662/job/39234630632?pr=2761

PR #2727, opened 2025-03-03
chore: react router 7.2

david-crespo:


david-crespo:
Saw an e2e flake but it predates the route modules business: #2704

PR #2742, opened 2025-03-04
Updated badges with integrated transitional spinners

benjaminleonard:
CleanShot 2025-03-04 at 15 49 52


david-crespo:
Looks great. I kinda miss the tooltip on the badge. It's moving, it demands to be hovered on. I'm like a dog seeing a squirrel.

I forgot about this one, which is still there:

image

benjaminleonard:
> Looks great. I kinda miss the tooltip on the badge. It's moving, it demands to be hovered on. I'm like a dog seeing a squirrel.

I was thinking we'd switch to the dropdown auto refresh. Maybe we can put it in the dropdown so if anyone ever clicks it open because they are curious about the frequency of the refresh it's displayed there.

app/ui/styles/components/spinner.css
```diff
@@ -13,6 +13,11 @@
   animation: rotate 5s linear infinite;
 }
 
+.spinner .path,
+.spinner .bg {
+  stroke: currentColor;
```
benjaminleonard, line 18: When a button is disabled this makes it look dull because the text colour of the button becomes something like `text-accent-disabled`. Need to investigate.

benjaminleonard:
Whilst I'm im there – handling disk and snapshot states.

CleanShot 2025-03-06 at 11 40 14

`importing_from_url`, `import_ready` and `importing_from_bulk_writes` are pretty goofy. Is it worth stripping out the `_` characters.

image

test/e2e/instance-create.e2e.ts
```diff
@@ -509,25 +509,25 @@ test('attaching additional disks allows for combobox filtering', async ({ page }
   await attachExistingDiskButton.click()
   await selectADisk.click()
   // several disks should be shown
-  await expect(page.getByRole('option', { name: 'disk-0001' })).toBeVisible()
-  await expect(page.getByRole('option', { name: 'disk-0002' })).toBeVisible()
-  await expect(page.getByRole('option', { name: 'disk-1000' })).toBeVisible()
+  await expect(page.getByRole('option', { name: 'disk-0005' })).toBeVisible()
```
benjaminleonard, line 512: After seeding the disks with a mix of states, not all of the previous items were visible. Now the random numbers are deterministic, and so the same disks should be available each time – the only way I can see this breaking is if we fiddle with the `Rando` numbers.

david-crespo:
Definitely kill the underscores. I'm not sure all of those should get spinners — they're not all transitional in the sense of intended to be temporary and automatically moved to another state like with the instance states. A disk can be in importing from bulk writes forever.

benjaminleonard:
This is ready for review – I think switching from a refresh button to an interval picker is best done in a follow up PR. It requires some changes to the interval picker component.

PR #2747, opened 2025-03-05
Prototype: Stop instance disk modal

benjaminleonard:
Disabling buttons (with a tooltip) is a pattern I'm not particularly fond of – outside of the lesser accessibility, it adds friction in a situation we can help the user get what they want done.

Rough implementation of an idea I had around letting the user stop the instance inline within the disk attach modal. Can be abstracted to any similar situation that requires something be done first.

The polling to see if the instance state has change is less than ideal perhaps, and might require some more thinking.

CleanShot 2025-03-05 at 12 25 06

Interested to hear what you think? Maybe the consensus is that the added friction to stop an instance is necessary (especially since we have a confirm modal elsewhere) because of the consequences of stopping a VM accidentally. And the user still needs to start the instance after attaching.

### DIFF
--- a/app/components/StopInstancePrompt.tsx
+++ b/app/components/StopInstancePrompt.tsx
@@ -1,0 +1,110 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+import { useEffect, useState, type ReactNode } from 'react'
+
+import {
+  instanceTransitioning,
+  useApiMutation,
+  useApiQuery,
+  useApiQueryClient,
+  type Instance,
+} from '@oxide/api'
+
+import { HL } from '~/components/HL'
+import { addToast } from '~/stores/toast'
+import { Button } from '~/ui/lib/Button'
+import { Message } from '~/ui/lib/Message'
+
+const POLL_INTERVAL_FAST = 2000 // 2 seconds
+
+type StopInstancePromptProps = {
+  instance: Instance
+  children: ReactNode
+}
+
+export function StopInstancePrompt({ instance, children }: StopInstancePromptProps) {
+  const queryClient = useApiQueryClient()
+  const [isStoppingInstance, setIsStoppingInstance] = useState(false)
+
+  const { data } = useApiQuery(
+    'instanceView',
+    {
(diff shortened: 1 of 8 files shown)


app/forms/disk-attach.tsx
```diff
@@ -25,6 +26,7 @@ type AttachDiskProps = {
   diskNamesToExclude?: string[]
   loading?: boolean
   submitError?: ApiError | null
+  instance: Instance
```
david-crespo, line 29: Requiring the instance isn't compatible with the use of this on the instance create form, but we can figure out how to deal with that.

david-crespo:
I like it as an interaction. I simplified things a bit, which makes it less of a question of whether it's worth the complexity.

app/components/StopInstancePrompt.tsx
```diff
@@ -0,0 +1,64 @@
...
+  instance: Instance
+  children: ReactNode
+}
+
+export function StopInstancePrompt({ instance, children }: StopInstancePromptProps) {
+  const isStoppingInstance = instance.runState === 'stopping'
+
+  const stopInstance = useApiMutation('instanceStop', {
+    onSuccess: () => {
+      // trigger polling by the top level InstancePage one
```
benjaminleonard, line 27: Oh this is very clever

app/ui/styles/components/button.css
```diff
@@ -44,6 +44,10 @@
   @apply hidden;
 }
 
+.btn-notice.ox-button:after {
```
benjaminleonard, line 47: I want to clean up button styles for these mini coloured ghost icons

app/ui/styles/components/spinner.css
```diff
@@ -13,6 +13,11 @@
   animation: rotate 5s linear infinite;
 }
 
+.spinner .path,
```
benjaminleonard, line 16: This is borrowed from #2742

app/components/StopInstancePrompt.tsx
david-crespo: The only downside is that you have a delay in the loading state because it doesn't change to loading while the initial stop request goes through. I tried this

```ts
const isStoppingInstance = stopInstance.isPending || instance.runState === 'stopping'
```

and it does respond immediately, but it flashes back to off briefly in between the stop request completing and the invalidation running and flipping the instance to 'stopping'. I can't think of a fix other than adding back in the explicit state thing you had. Those things make me nervous because the logic has to be bulletproof to avoid accidentally getting stuck in a loading state, but I'm sure we can do it.

PR #2748, opened 2025-03-06
add toPrefix to NavLinks in Sidebar to better highlight isActive

charliepark:
This fixes an issue that we were seeing with tabbed pages, where the sidebar navigation wasn't working as expected for tabbed pages.

Note how the "Inventory" link is "active" for the Sleds tab, but not the Disks tab:
Screenshot 2025-03-06 at 2 45 37 PM
Screenshot 2025-03-06 at 2 45 42 PM

This PR fixes NavLinks in the Sidebar by adding a `toPrefix`, which is essentially the first part of a URL path. If a link is given a `toPrefix` prop, that more generic string will be used when the component is determining if the tab should be rendered as "active" or not.

Here it is, fixed:
navlinks

### DIFF
--- a/app/components/RouteTabs.tsx
+++ b/app/components/RouteTabs.tsx
@@ -84,11 +84,11 @@ export interface TabProps {
    * able to link directly to the first sidebar tab, but we of course also want
    * this tab to appear active for all the sidebar tabs. See instance metrics.
    */
-  toPrefix?: string
+  activePrefix?: string
   children: ReactNode
 }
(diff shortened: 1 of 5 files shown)


david-crespo:
Looks good. What do you think about renaming `toPrefix` to `activePrefix` (both here and on the existing RouteTab one)? I came up with `toPrefix` but in hindsight it would be good to say what it’s for right in the name.

david-crespo (approved):

PR #2751, opened 2025-03-11
Allow links in row actions and more actions dropdowns

### LINKED ISSUE #1855: Allow real links in more actions menu

### LINKED ISSUE #2174: IP pool list: disable pool delete when there are ranges in pool

david-crespo:
Closes #1855 
Closes #2174

It's more noisy than it needs to be because I also cleaned up some duplicated logic that didn't need to exist — notice there are net deletes despite the increase in functionality. I think I would like to convert `RowActions` to taking elements directly as children as well instead of using these action config objects we have to memoize, but that would be too much churn for this PR: we have row actions in like 20 files.

### DIFF
--- a/app/pages/project/instances/InstancePage.tsx
+++ b/app/pages/project/instances/InstancePage.tsx
@@ -6,7 +6,7 @@
  * Copyright Oxide Computer Company
  */
 import { filesize } from 'filesize'
-import { useId, useMemo, useState } from 'react'
+import { useId, useState } from 'react'
 import { useForm } from 'react-hook-form'
 import { Link, useNavigate, type LoaderFunctionArgs } from 'react-router'
 
@@ -27,6 +27,7 @@ import {
   instanceCan,
   instanceTransitioning,
 } from '~/api/util'
+import { CopyIdItem } from '~/components/CopyIdItem'
 import { ExternalIps } from '~/components/ExternalIps'
 import { NumberField } from '~/components/form/fields/NumberField'
 import { HL } from '~/components/HL'
@@ -44,6 +45,7 @@ import {
 import { addToast } from '~/stores/toast'
 import { EmptyCell } from '~/table/cells/EmptyCell'
 import { Button } from '~/ui/lib/Button'
+import * as DropdownMenu from '~/ui/lib/DropdownMenu'
 import { Message } from '~/ui/lib/Message'
 import { Modal } from '~/ui/lib/Modal'
 import { PageHeader, PageTitle } from '~/ui/lib/PageHeader'
@@ -179,19 +181,6 @@ export default function InstancePage() {
     { enabled: !!primaryVpcId }
   )
 
-  const allMenuActions = useMemo(
-    () => [
-      {
-        label: 'Copy ID',
-        onActivate() {
-          window.navigator.clipboard.writeText(instance.id || '')
-        },
-      },
-      ...makeMenuActions(instance),
-    ],
-    [instance, makeMenuActions]
-  )
-
   const memory = filesize(instance.memory, { output: 'object', base: 2 })
 
   return (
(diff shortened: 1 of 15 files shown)


app/pages/project/instances/actions.tsx
```diff
@@ -153,9 +153,7 @@ export const useMakeInstanceActions = (
         },
         {
           label: 'View serial console',
-          onActivate() {
-            navigate(pb.serialConsole({ project, instance: instance.name }))
-          },
+          to: pb.serialConsole({ project, instance: instance.name }),
```
david-crespo, line 156: This is the narrow change asked for on behalf of an external user in #1855. This allows serial console to be opened in a new tab from both instance list row actions and instance detail "more actions".

app/pages/project/vpcs/RouterPage.tsx
```diff
@@ -197,7 +186,9 @@ export default function RouterPage() {
             summary="Routers are collections of routes that direct traffic between VPCs and their subnets."
             links={[docLinks.routers]}
           />
-          <MoreActionsMenu label="Router actions" actions={actions} />
+          <MoreActionsMenu label="Router actions">
+            <CopyIdItem id={routerData.id} />
+          </MoreActionsMenu>
```
david-crespo, line 191: net negative lines speak for themselves I think

app/pages/project/vpcs/VpcPage.tsx
```diff
@@ -46,33 +46,23 @@ export default function VpcPage() {
...
+            <DropdownMenu.LinkItem to={pb.vpcEdit(vpcSelector)}>Edit</DropdownMenu.LinkItem>
+            <DropdownMenu.Item
+              label="Delete"
+              onSelect={confirmDelete({
+                doDelete: () => deleteVpc({ path: { vpc: vpcName }, query: { project } }),
+                label: vpcName,
+              })}
+              className="destructive"
+            />
+          </MoreActionsMenu>
```
david-crespo, line 65: This shows why the children approach is nice. There's no config object where you then elsewhere have to decide what it renders. You just render those things directly.

david-crespo:
Claude caught a real bug. A bargain at 2.8 cents.

image

app/pages/project/instances/InstancePage.tsx
```diff
@@ -215,7 +204,28 @@ export default function InstancePage() {
...
+                <DropdownMenu.Item
+                  key={action.label}
+                  label={action.label}
+                  onSelect={action.onActivate}
+                  disabled={action.disabled}
+                  className={action.className}
+                />
+              )
+            )}
+          </MoreActionsMenu>
```
david-crespo, line 228: This is the one spot where we are adding some noise, because this is the one spot where we were passing the same set of action config objects to `RowActions` in one place (instance list) and `MoreActionsMenu` in another (here, instance detail). Because `MoreActionsMenu` has been converted to take children directly, we have to do the conversion to elements here. Maybe this should be extracted since it's the same as what RowActions does internally....

app/table/columns/action-col.tsx
```diff
@@ -66,7 +78,6 @@ type RowActionsProps = {
 export const RowActions = ({ id, copyIdLabel = 'Copy ID', actions }: RowActionsProps) => {
   return (
     <DropdownMenu.Root>
-      {/* TODO: This name should not suck; future us, make it so! */}
```
david-crespo, line 69: This comment is from May 2022 #883 and refers to the aria label "Row actions", which tbh I don't see anything wrong with. That's what they are.

PR #2753, opened 2025-03-12
chore: Delete non-streaming mock serial console

david-crespo:
It is very cool, so I could be persuaded to keep it. Currently we're not using it — it is used in the response from the mock non-streaming endpoint, and we use the streaming one. Here's what it would look like if we hooked it up. I kinda think it's too realistic — makes it seem like the VM is real.

https://github.com/user-attachments/assets/671df32f-aaa5-48bf-8c05-44866ad84f3e


david-crespo:
Weird e2e flakes. They don't seem related.

PR #2754, opened 2025-03-13
Page loading skeleton

### LINKED ISSUE #1569: Top-level loading state for initial pageload
On initial pageload, we just show a blank screen while the loaders run.

2023-06-09-no-loading-state

We could show the loading bar we use for client-side navs, that would be something, but it would probably be better if we had a page-side skeleton state. The only problem there is that if we have a page in the console that does not have that layout (for example device auth or login), we would want a different skeleton state for those pages. For that reason, I think the skeleton state should be configured at the layout level. React Router makes this very easy, I think there's a `fallbackElement` (or something) right on the `<Route>`.

david-crespo:
Figured out how to do it and it is good. Closes #1569. In the video a show that we avoid showing the skeleton on login and device token verify pages that do not have the grid lines.

https://github.com/user-attachments/assets/a6d427d3-92b5-447f-a8c4-c52494d07b32

### DIFF
--- a/app/routes.tsx
+++ b/app/routes.tsx
@@ -9,9 +9,9 @@ import type { ReactElement } from 'react'
 import {
   createRoutesFromElements,
   Navigate,
+  redirect,
   Route,
   type LoaderFunctionArgs,
-  type redirect,
 } from 'react-router'
 
 import { NotFound } from './components/ErrorPage'
@@ -207,7 +207,7 @@ export const routes = createRoutesFromElements(
         </Route>
       </Route>
 
-      <Route index element={<Navigate to={pb.projects()} replace />} />
+      <Route index loader={() => redirect(pb.projects())} />
 
       <Route lazy={() => import('./layouts/SiloLayout').then(convert)}>
         <Route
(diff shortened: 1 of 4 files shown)


david-crespo:
Ok I just noticed on the preview that in MSW banner mode, we need the banner on the loading state, otherwise the grid doesn't line up when the page pops in.

app/routes.tsx
```diff
@@ -192,7 +207,7 @@ export const routes = createRoutesFromElements(
         </Route>
       </Route>
 
-      <Route index element={<Navigate to={pb.projects()} replace />} />
+      <Route index loader={() => redirect(pb.projects())} />
```
david-crespo, line 210: Without this, when loading the root (an extremely common thing to do) you'd see the skeleton flash and then it would go blank wile the loaders for the next route load. I think doing it this way is fine as long we never link to `/`, which we don't. We'll have to see if there's any weird behavior, though. I think another workaround would be to leave the element thing in place and give this route a loader that just calls the loader from the projects page, that way the loading time is spent with the fallback up.

PR #2760, opened 2025-03-21
Add Anti-Affinity to the UI: Phase 1 — read-only

charliepark:
This PR is based off of this Omicron PR — https://github.com/oxidecomputer/omicron/pull/7658 — so won't be mergeable quite yet. But this gets the UI ready for Anti-Affinity. (https://github.com/oxidecomputer/omicron/pull/7858 is also relevant)

This PR is read-only, but we'll build off of it for Create/Update/Delete in subsequent PRs.
Screenshot 2025-03-20 at 5 16 54 PM
Screenshot 2025-03-20 at 5 17 09 PM


mock-api/msw/handlers.ts
```diff
@@ -1601,26 +1603,80 @@ export const handlers = makeHandlers({
...
+    return paginated(query, antiAffinityGroups)
+  },
+  antiAffinityGroupView: ({ path, query }) =>
+    lookup.antiAffinityGroup({ ...path, ...query }),
+  antiAffinityGroupMemberList: ({ path, query }) => {
+    const project = lookup.project({ ...query })
+    const antiAffinityGroup = lookup.antiAffinityGroup({
+      project: project.id,
+      antiAffinityGroup: path.antiAffinityGroup,
+    })
```
david-crespo, line 1642: The group lookup handles the project lookup internally:

```suggestion
    const antiAffinityGroup = lookup.antiAffinityGroup({ ...path, ...query })
```

mock-api/msw/handlers.ts
```diff
@@ -1601,26 +1603,80 @@ export const handlers = makeHandlers({
...
+        const { id, name, run_state } = lookup.instance({
+          instance: i.anti_affinity_group_member.value.id,
+        })
+        return { type: 'instance', value: { id, name, run_state } }
+      })
+    return { items: members }
+  },
+  antiAffinityGroupMemberInstanceDelete: ({ path, query }) => {
+    const project = lookup.project({ ...query })
+    const instance = lookup.instance({ ...query, instance: path.instance })
```
david-crespo, line 1655: This instance lookup made me realize: in the mock instance delete handler, we have to go and make sure we are also deleting it from any affinity groups.

mock-api/msw/db.ts
```diff
@@ -436,6 +472,12 @@ type DiskBulkImport = {
 }
 
 const initDb = {
+  affinityGroups: [...mock.affinityGroups],
+  affinityGroupMembers: [...mock.affinityGroupMembers],
+  affinityGroupMemberLists: [...mock.affinityGroupMemberLists],
+  antiAffinityGroups: [...mock.antiAffinityGroups],
+  antiAffinityGroupMembers: [...mock.antiAffinityGroupMembers],
```
david-crespo, line 479: Looks like this and `affinityGroupMembers` aren't used — they could be removed from here, and probably removed altogether.

app/ui/lib/Badge.tsx
```diff
@@ -62,3 +62,13 @@ export const Badge = ({
     </span>
   )
 }
+
+export const AffinityGroupPolicyBadge = (props: { policy: string; className?: string }) => (
+  <Badge
+    color="neutral"
+    variant={props.policy === 'allow' ? 'default' : 'solid'}
```
david-crespo, line 69: use `ts-pattern` for an exhaustive match on the policy values. also maybe move this into the file where it's used, or if it's going to be shared, maybe `app/components`

app/routes.tsx
```diff
@@ -485,6 +485,18 @@ export const routes = createRoutesFromElements(
...
+            lazy={() => import('./pages/project/affinity/AffinityPage.tsx').then(convert)}
+          />
+          <Route path="affinity" handle={{ crumb: 'Affinity' }}>
+            <Route
+              path=":antiAffinityGroup"
+              lazy={() =>
+                import('./pages/project/affinity/AntiAffinityGroupPage.tsx').then(convert)
+              }
+            />
+          </Route>
```
david-crespo, line 499: This way expresses the situation more clearly and lets you share the crumb (remember to take the handle out of `AffinityPage.tsx`)

```tsx
<Route path="affinity" handle={{ crumb: 'Affinity' }}>
  <Route
    index
    lazy={() => import('./pages/project/affinity/AffinityPage.tsx').then(convert)}
  />
  <Route
    path=":antiAffinityGroup"
    lazy={() =>
      import('./pages/project/affinity/AntiAffinityGroupPage.tsx').then(convert)
    }
  />
</Route>
```

mock-api/affinity-group.ts
```diff
@@ -0,0 +1,198 @@
...
+  },
+  {
+    affinity_group_id: affinityGroup.id,
+    affinity_group_member: affinityGroupMemberStartingInstance,
+  },
+  {
+    affinity_group_id: sandboxAffinityGroup.id,
+    affinity_group_member: affinityGroupMemberInstance,
+  },
+]
```
david-crespo, line 107: You can cut a bunch of lines in this file by deleting `affinityGroupMemberInstance` and friends and changing the join list to only use the instance ID.

```ts
type DbAffinityGroupMember = {
  affinity_group_id: string
  affinity_group_member: { type: 'instance'; id: string }
}
```

mock-api/affinity-group.ts
```diff
@@ -0,0 +1,131 @@
...
+    policy: 'fail',
+    project_id: project.id,
+    time_created: '2025-04-13T00:00:00.000Z',
+    time_modified: '2025-04-16T00:00:00.000Z',
+  },
+]
+
+type DbAffinityGroupMember = {
+  affinity_group_id: string
+  affinity_group_member: { type: 'instance'; id: TypedUuidForInstanceKind }
```
david-crespo, line 54: I think it's better to avoid the typed UUID types because they give the impression something fancier is happening that isn't — they're just aliases for `string`.

mock-api/msw/handlers.ts
```diff
@@ -1601,26 +1610,76 @@ export const handlers = makeHandlers({
...
+    const affinityGroups = db.affinityGroups.filter((i) => i.project_id === project.id)
+    return paginated(query, affinityGroups)
+  },
+  affinityGroupView: ({ path, query }) => lookup.affinityGroup({ ...path, ...query }),
+  affinityGroupMemberList: ({ path, query }) => {
+    const project = lookup.project({ ...query })
+    const affinityGroup = lookup.affinityGroup({
+      project: project.id,
+      affinityGroup: path.affinityGroup,
+    })
```
david-crespo, line 1624: this one can also be

```suggestion
    const affinityGroup = lookup.affinityGroup({ ...path, ...query })
```

app/pages/project/affinity/AntiAffinityGroupPage.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const { antiAffinityGroup, project } = useAntiAffinityGroupSelector()
+  const { data: group } = usePrefetchedQuery(
+    antiAffinityGroupView({ antiAffinityGroup, project })
+  )
+  const { id, name, description, policy, timeCreated } = group
+  const { data: members } = usePrefetchedQuery(
+    memberList({ antiAffinityGroup, project }).optionsFn()
+  )
+  // TODO: Run an ALL_ISH query to get total number of members
+  const membersCount = members?.items.length ?? 0
```
david-crespo, line 89: `members` is always defined because of the prefetch, so no need for `?.` or fallback. That got me interested in [this typescript-eslint rule](https://typescript-eslint.io/rules/no-unnecessary-condition), which does error on this line, but it also turns up 50 other spots and some of them are non-trivial to fix.

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -0,0 +1,168 @@
...
+    </>
+  )
+}
+
+export const AntiAffinityGroupEmptyState = () => (
+  <TableEmptyBox>
+    <EmptyMessage
+      icon={<Affinity24Icon />}
+      title="No anti-affinity groups"
+      body="Create a new anti-affinity group to see it here"
```
david-crespo, line 131:
```suggestion
      body="Create an anti-affinity group to see it here"
```

app/pages/project/affinity/AntiAffinityGroupPage.tsx
```diff
@@ -0,0 +1,134 @@
...
+  ])
+  return null
+}
+
+const AntiAffinityGroupMemberEmptyState = () => (
+  <TableEmptyBox>
+    <EmptyMessage
+      icon={<Affinity24Icon />}
+      title="No anti-affinity group members"
+      body="Add a new anti-affinity group member to see it here"
```
david-crespo, line 71: How about "Add an instance to the group to see it here" ?

app/util/path-builder.ts
```diff
@@ -97,6 +98,18 @@ export const pb = {
...
+    `${pb.affinityGroups(params)}/${params.affinityGroup}`,
+  affinityGroupNew: (params: PP.Project) =>
+    `${pb.affinityGroups(params)}/affinity-groups-new`,
+
+  antiAffinityGroups: (params: PP.Project) => affinityBase(params),
+  antiAffinityGroup: (params: PP.AntiAffinityGroup) =>
+    `${affinityBase(params)}/${params.antiAffinityGroup}`,
+  antiAffinityGroupNew: (params: PP.Project) =>
+    `${affinityBase(params)}/anti-affinity-groups-new`,
+
```
david-crespo, line 112: I'd get rid of the paths that don't exist in the routes file yet and get rid of `affinityBase` since there are no longer any tabs, so the base/real route distinction no longer applies. You can add the create form path in the next PR when you add the route for it.

```ts
affinity: (params: PP.Project) => `${projectBase(params)}/affinity`,
antiAffinityGroup: (params: PP.AntiAffinityGroup) => `${pb.affinity(params)}/${params.antiAffinityGroup}`,
```

app/pages/project/affinity/AntiAffinityGroupPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+    memberList({ antiAffinityGroup, project }).optionsFn()
+  )
+  const membersCount = members.items.length
+  const staticCols = [
+    colHelper.accessor('value.name', {
+      header: 'Name',
+      cell: makeLinkCell((instance) => pb.instance({ project, instance })),
+    }),
+    colHelper.accessor('value.runState', Columns.instanceState),
+  ]
```
david-crespo, line 95: These aren't static — "static" is used in these variable names to refer to columns that are defined outside of render and therefore don't need to be memoized. These depend on `project`, so they can't be made static. The state column is static, but it's not worth pulling it out if you're going to have to memoize the full set anyway.

Short version: memoize this and rename it to `columns`.

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -0,0 +1,162 @@
...
+    }),
+    colHelper.accessor('policy', {
+      cell: (info) => <AffinityGroupPolicyBadge policy={info.getValue()} />,
+    }),
+    colHelper.accessor('name', {
+      header: 'members',
+      cell: (info) => <AffinityGroupMembersCell antiAffinityGroup={info.getValue()} />,
+    }),
+    colHelper.accessor('timeCreated', Columns.timeCreated),
+  ]
```
david-crespo, line 103: Same deal as the other one, though in this case because you have 5 static columns, I think it is worth it to pull out the true `staticCols` at top level and then spread that into the list here inside the `useMemo`.

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -0,0 +1,162 @@
...
+    <Link
+      key={instance}
+      to={pb.instance({ project, instance })}
+      className="link-with-underline"
+    >
+      {instance}
+    </Link>
+  ))
+  if (instances.length > instancesToShow.length) {
+    links.push(<>…</>)
```
david-crespo, line 159: This could be an actual link to the list page, like we do with IPs. Not a big deal either way. This is so similar to to `ExternalIps`, but I can see why you didn't share the code — the abstraction would be worse than the duplication.

app/pages/project/affinity/AntiAffinityGroupPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const AntiAffinityGroupMemberEmptyState = () => (
+  <TableEmptyBox>
+    <EmptyMessage
+      icon={<Affinity24Icon />}
+      title="No anti-affinity group members"
+      body="Add an instance to the group to see it here"
+      buttonText="Add anti-affinity group member"
+      // TODO: this should open the AntiAffinityGroup edit modal
+      buttonTo={pb.antiAffinityGroupNew(useProjectSelector())}
```
david-crespo, line 74: For this PR to be self-contained I would just comment out the button props.

benjaminleonard:
Small detail, but its missing the dividing line. Aesthetically it does a similar job to the tab bar, like on the instance page where it separates the page content from the page header.

image

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -0,0 +1,162 @@
...
+    .then((data) => {
+      // Preload the anti-affinity group details
+      data.items.forEach((antiAffinityGroup) => {
+        queryClient.fetchQuery(
+          memberList({ antiAffinityGroup: antiAffinityGroup.name, project })
+        )
+      })
+    })
+  return null
+}
```
david-crespo, line 60: This is not quite right because the `fetchQuery` is not being awaited. This explains why you were (I'm guessing) stuck using `useApiQuery` in the component — the fetches are kicked off by the loader, but the loader doesn't wait for them to be done. The way to make this wait would be:

```ts
const groups = await queryClient.fetchQuery(
  antiAffinityGroupList({ project }).optionsFn()
)
await Promise.all(
  groups.items.map(({ name }) =>
    queryClient.fetchQuery(memberList({ antiAffinityGroup: name, project }))
  )
)
```

(I prefer await instead of the `then` here because there's only one fetch to wait for before kicking off the requests that depend on it — in the other spots where we use `then` in a loader it's because there's a top-level `Promise.all` and I'm trying to jam everything in it.)

However, with 16 instances in a group, it would probably make the pageload time too long if we wait for them all. A fun thing about this implementation is that you can turn off waiting by just removing the `await` from the `Promise.all`. Putting a little timer in there confirms that we're talking a few hundred ms without waiting for the members vs. double that if there are just a few members (fewer than 6 would mean they all run in parallel without any queuing by the browser).

```ts
export async function clientLoader({ params }: LoaderFunctionArgs) {
  const start = performance.now()
  const { project } = getProjectSelector(params)
  const groups = await queryClient.fetchQuery(
    antiAffinityGroupList({ project }).optionsFn()
  )
  /* await */ Promise.all(
    groups.items.map(({ name }) =>
      queryClient.fetchQuery(memberList({ antiAffinityGroup: name, project }))
    )
  )
  console.log(performance.now() - start, 'ms')
  return null
}
```

The downside to not waiting is that the members column pops in a bit and the whole table shifts. I have a slight preference for a faster initial load, but I could see the argument the other way. We could also get sneaky and decide whether to wait or not based on the number of members. Like if there are under 6 we could wait, but if there are more, don't bother:

```ts
const memberFetches = groups.items.map(({ name }) =>
  queryClient.prefetchQuery(memberList({ antiAffinityGroup: name, project }))
)
if (groups.items.length < 6) await Promise.all(memberFetches)
```

david-crespo: Oh, one more thing. Generally when we are going to let something load async, we don't bother doing anything in the loader for it. The VPC list is one example, where we fetch the firewall rules for each VPC async as a regular `useQuery`. 

https://github.com/oxidecomputer/console/blob/179f347f7a1b8293befd34c4313e927a3a76c7e6/app/pages/project/vpcs/VpcsPage.tsx#L57-L73

However! Kicking off a fetch in the loader but not waiting for is actually a cool idea because it starts the fetch earlier. How much earlier? When I stick `console.time()` and `console.timeLog()` to see how long it is between the loader ending and the render happening:


```diff
diff --git a/app/pages/project/affinity/AffinityPage.tsx b/app/pages/project/affinity/AffinityPage.tsx
index 11d4bb76cc..df76942ad3 100644
--- a/app/pages/project/affinity/AffinityPage.tsx
+++ b/app/pages/project/affinity/AffinityPage.tsx
@@ -48,14 +48,7 @@
   const { project } = getProjectSelector(params)
   await queryClient
     .fetchQuery(antiAffinityGroupList({ project }).optionsFn())
-    .then((data) => {
-      // Preload the anti-affinity group details
-      data.items.forEach((antiAffinityGroup) => {
-        queryClient.fetchQuery(
-          memberList({ antiAffinityGroup: antiAffinityGroup.name, project })
-        )
-      })
-    })
+  console.time()
   return null
 }
 
@@ -136,6 +129,7 @@
   antiAffinityGroup: string
 }) => {
   const { project } = useProjectSelector()
+  console.timeLog()
   const { data: members } = useApiQuery('antiAffinityGroupMemberList', {
     path: { antiAffinityGroup },
     query: { project, limit: 2 },
```

...the results are interesting! If I turn off the latency randomization in the mock API, I consistently get 50-70ms from the end of the loader to the render, so that's the time we would save by kicking off the fetches a bit earlier. Not nothing but also not a ton. However, if I leave the randomization on and even make it a bit wider of a range, we sometimes have a much bigger gap because other loaders running in parallel to this one might take longer than this one, giving us more time in which nothing is happening, in which we could have been fetching the groups. So overall I think kicking off the fetches in the loader even when we're not waiting for them is probably a good idea and unlikely to have downsides.

That is to say, I think this one is the way to go:

```ts
// kick off the fetches but only conditionally wait for them
const memberFetches = groups.items.map(({ name }) =>
  queryClient.prefetchQuery(memberList({ antiAffinityGroup: name, project }))
)
if (groups.items.length < 6) await Promise.all(memberFetches)
```

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -0,0 +1,162 @@
...
+export const AffinityGroupMembersCell = ({
+  antiAffinityGroup,
+}: {
+  antiAffinityGroup: string
+}) => {
+  const { project } = useProjectSelector()
+  const { data: members } = useApiQuery('antiAffinityGroupMemberList', {
+    path: { antiAffinityGroup },
+    query: { project, limit: 2 },
+  })
```
david-crespo, line 142: Check this out. You can still use the options helper with non-prefetched `useQuery`.

```suggestion
  const { data: members } = useQuery(memberList({ project, antiAffinityGroup }))
```

david-crespo: Better to do it yourself — you need to add the import for `useQuery` anyway, and who knows whether what I wrote will typecheck or lint

david-crespo (approved):
Let's party. We'll do the e2e tests once we have new/edit.

PR #2761, opened 2025-03-21
Descriptions for CPU metrics

### LINKED ISSUE #2714: Add help copy explaining instance metrics
We added several new metrics, including CPU utilization charts that incorporate different states — running, emulating, waiting, idling. We should add some help copy to explain them, and link to relevant documentation for more.

david-crespo:
Closes #2714. See notes in that issue about CPU states. Confirmed in Q&A channel that none of my choices here are egregious.

image

### DIFF
--- a/app/pages/project/instances/CpuMetricsTab.tsx
+++ b/app/pages/project/instances/CpuMetricsTab.tsx
@@ -22,6 +22,15 @@ import { Listbox } from '~/ui/lib/Listbox'
 
 import { useMetricsContext } from './common'
 
+type CpuChartType = OxqlVcpuState | 'all'
+
+const descriptions: Record<OxqlVcpuState, string | undefined> = {
+  run: 'Executing guest instructions',
+  idle: 'CPU has no work to do',
+  emulation: 'Handling guest operations in the host (like I/O)',
+  waiting: 'Ready but waiting, usually due to contention',
+}
+
 export default function CpuMetricsTab() {
   const { project, instance } = useInstanceSelector()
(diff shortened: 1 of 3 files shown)


david-crespo:
These descriptions are all kind of redundant, and the references to "the disk" are confusing in the case of the default "all disks". If we remove "on the disk" it becomes even more similar to the title of the chart. So I lean toward removing them altogether, which I did in cfbcd1052359461a1f32555c5c961d16881f0e30.

Note also the use of "on the link" in the network ones, which should be "on the interface". I decided they are similarly redundant, don't think they add anything.

### Before

image

image

### After

image

image

app/pages/project/instances/CpuMetricsTab.tsx
```diff
@@ -22,6 +22,15 @@ import { Listbox } from '~/ui/lib/Listbox'
 
 import { useMetricsContext } from './common'
 
+type CpuChartType = OxqlVcpuState | 'all'
+
+const descriptions: Record<OxqlVcpuState, string | undefined> = {
+  run: 'Executing guest instructions',
+  idle: 'CPU has no work to do',
```
david-crespo, line 29: This one is a little casual, but the alternatives I've come up with are too formal, like  "Not executing any instructions"

david-crespo: Claude's ideas are bad. GPT-4.5's dramatically more expensive ideas are much better.

image

david-crespo: Went with "Not executing instructions"

Issue #2737, opened 2025-03-03
OxQL metrics: explicit "no data for this time range" state


david-crespo:
Possibly would cover #2733 if there's nothing special we need to do there in the data handling.

Issue #2733, opened 2025-03-03
OxQL metrics: stuck in loading when exactly one data point


david-crespo:
I started an instance and went to metrics. Because my instance had just started and the time range for metrics ended at `now`, all my graphs had one data point. Unfortunately, we always drop the first data point because it is the cumulative value up until now, and then the chart component treats an empty list of points as a loading state.

I think in this case we probably want to keep the former but change the latter. I'm not sure we can usefully display the cumulative value, though maybe we can get cute and do what we kinda used to do, start with that value and sort of synthetically fill out the data with a bunch of identical values using it. (We would have to make sure we don't project that value into the future.)

https://github.com/oxidecomputer/console/blob/1fe4a7ec4cb36ac5b91b2f237c9addb95a29ab86/app/components/oxql-metrics/util.ts?plain=1#L140-L145

https://github.com/oxidecomputer/console/blob/1fe4a7ec4cb36ac5b91b2f237c9addb95a29ab86/app/components/TimeSeriesChart.tsx?plain=1#L223-L229

![Image](https://github.com/user-attachments/assets/caca1c10-c3b8-4f9d-886d-71c09cdb782a)

![Image](https://github.com/user-attachments/assets/f4c42d58-8c08-43b8-bd49-597ef9f44585)

Issue #2732, opened 2025-03-03
OxQL metrics: don't cut off the part of the time range with no data


david-crespo:
I started an instance and messed with it a bit to test the charts. I made the time range 3 hours, including time before the instance existed and in the future. The time before comes back with no data, while the time in the future comes back with nulls. We treat these differently: we display the nulls but not the nothing.

I'm not 100% sure whether we should always display the full range asked for — on one hand I was pretty confused about why my graph started at a time other than the one I put in the picker. On the other hand, if you ask for a big range and there's only data for a little bit of it, you get a pretty useless graph. On the third and hopefully final hand, in that situation the user can simply change the time range to zoom in on the relevant slice. (This rules out bad ideas I had like a checkbox about whether to show the full range or only spots where we have actual data. That's hard to explain and, more importantly, covered by the existing controls.)

<img width="1188" alt="Image" src="https://github.com/user-attachments/assets/9cac14ed-785e-4072-8b01-06e771296833" />

david-crespo:
By coincidence, Geoffrey Litt [wrote](https://x.com/geoffreylitt/status/1896682818362843159) the following a few minutes before I wrote the above about redundant one-off controls:

> along the same lines - one of the worst misconceptions in product design is that a microwave needs to have a button for every thing you could possibly cook: "popcorn", "chicken", "potato", "frozen vegetable" bla bla bla.
> 
> you really don't! you can just have a time (and power) button. people will figure out how to cook stuff.
> 
> good designs expose systematic structure; they lean on their users' ability to understand this structure and apply it to new situations. we were born for this.
> 
> bad designs paper over the structure with superficial labels that hide the underlying system, inhibiting their users' ability to actually build a clear model in their heads.

david-crespo:
Taking out of 14 but bot sure we should prioritize this for 15 yet.

PR #2762, opened 2025-03-22
Convert old charts to new style

david-crespo:
Looking at #2732, #2733, and #2737, I realized we'd benefit from simplifying the chart component, and one source of complexity was the fact that the non-OxQL system and silo utilization charts were styled differently. So here I bring them up to date with the new style and eliminate the props we were using to differentiate.

## Before

image

## After

image

### DIFF
--- a/app/components/SystemMetric.tsx
+++ b/app/components/SystemMetric.tsx
@@ -14,9 +14,7 @@ import {
   type SystemMetricName,
 } from '@oxide/api'
 
-import { Spinner } from '~/ui/lib/Spinner'
-
-import { TimeSeriesChart } from './TimeSeriesChart'
+import { ChartContainer, ChartHeader, TimeSeriesChart } from './TimeSeriesChart'
 
 // The difference between system metric and silo metric is
 //   1. different endpoints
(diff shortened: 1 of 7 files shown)


app/components/TimeSeriesChart.tsx
```diff
@@ -110,14 +112,11 @@ type TimeSeriesChartProps = {
   className?: string
   data: ChartDatum[] | undefined
   title: string
-  width: number
-  height: number
```
david-crespo, line 114: We were passing height and width at every callsite but it had no effect, probably because the `ResponsiveContainer` was overriding it.

app/components/SystemMetric.tsx
```diff
@@ -93,25 +91,17 @@ export function SiloMetric({
   // in the tooltip. could be just once on the end of the x-axis like GCP
 
   return (
-    <div>
-      <h2 className="flex items-center gap-1.5 px-3 text-mono-sm text-default">
-        {title} {unit && <span className="text-tertiary">({unit})</span>}{' '}
-        {(inRange.isPending || beforeStart.isPending) && <Spinner />}
```
david-crespo, line 99: The loading state is handled inside the chart. The logic needs to be improved, but it does show up for these charts.

PR #2764, opened 2025-03-24
Tighten spacing and more consistent with other layouts

benjaminleonard:
Making more consistent with other chart layouts. Also means we can simplify the interval picker.

image

### DIFF
--- a/app/components/RefetchIntervalPicker.tsx
+++ b/app/components/RefetchIntervalPicker.tsx
@@ -6,14 +6,13 @@
  * Copyright Oxide Computer Company
  */
 import cn from 'classnames'
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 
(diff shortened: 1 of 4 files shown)


benjaminleonard:
Similar to the instance metrics layout, I'd like to have the inputs go full width at a certain size but this will be much easier when we upgrade to tailwind 4 since we can use the container queries (we're using a plugin for it currently but it's missing a feature I'm looking for).

david-crespo:
Very dangerous to tell me that upgrading to Tailwind v4 is a product imperative

benjaminleonard:
> Very dangerous to tell me that upgrading to Tailwind v4 is a product imperative

It sounds fun until you realise we need to refactor the token system to accommodate (it's probably not as bad as I'm thinking)

david-crespo:
You forget that the token JSON is my sworn enemy

PR #2767, opened 2025-03-27
Instance affinity groups on settings tab

david-crespo:
Just a quick start. I actually really like the tooltips Claude Code put in the column header but we'll need to think about a) the copy, and b) spreading them around consistently so it doesn't feel haphazard.

image

<details>
<summary>Screenshot of initial version on its own tab</summary>

image

</details>

### DIFF
--- a/app/pages/project/instances/AutoRestartCard.tsx
+++ b/app/pages/project/instances/AutoRestartCard.tsx
@@ -1,0 +1,165 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+
+import { formatDistanceToNow } from 'date-fns'
+import { type ReactNode } from 'react'
+import { useForm } from 'react-hook-form'
+import { match } from 'ts-pattern'
+
+import {
+  apiQueryClient,
+  instanceAutoRestartingSoon,
+  useApiMutation,
+  usePrefetchedApiQuery,
+} from '~/api'
+import { ListboxField } from '~/components/form/fields/ListboxField'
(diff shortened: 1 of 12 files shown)


benjaminleonard:
I suggest either group or name for the first column label.

Do you think an entire tab gives it undue prominence. Especially since most instances are unlikely to be part of one. There are consequences to over-encouraging users to use affinity groups with live upgrade and I think the shape of the UI affects the ways in which users will consider the frequency to use it.

Then again I'm not a fan of conditionally rendering UI, because it's confusing for the use when it's not there.

Is there some commonality with auto-restart where we move them onto the same tab. Or just shift it into settings. The auto-restart one would need to be full width and the input partially widthed to accomodate.

david-crespo:
I'll fix the column header. Yeah, I could see putting it on settings. Good idea. I don't think giving it its own tab is necessarily bad either — we would have to make the empty state make clear that you don't need it unless you know you need it.

david-crespo:
That's not bad! Have to constrain the width to match the other card, but that's not too bad. Truncating the description is probably fine.

image

Had to add a prop to `DescriptionCell` to let me truncate more than the default.


image

Screenshot 2025-03-27 at 11 10 21 AM

david-crespo:
Yeah, maybe!

image

app/pages/project/instances/AutoRestartCard.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+const FormMeta = ({ label, tip, children }: FormMetaProps) => (
+  <div>
+    <div className="mb-2 flex items-center gap-1 border-b pb-2 text-sans-md border-secondary">
+      <div>{label}</div>
+      {tip && <TipIcon>{tip}</TipIcon>}
+    </div>
+    {children}
+  </div>
+)
```
david-crespo, line 165: Unfortunately git can't tell this is almost entirely a move. The only change is taking off `width="medium"` and putting `max-w-lg` on the `ListboxField`.

Issue #2416, opened 2024-09-05
Feedback on copy ID buttons


david-crespo:
A toast is probably too much (and it would be better if the effect was local to the button). As it stands, it feels like "... did that work?" The hard thing is that the button disappears when you click it, since it's in a little menu. I wonder if it would feel weird to delay the close (somehow) and show a little checkmark animation inside the button. I guess if that's too weird we can just do a toast.

https://github.com/user-attachments/assets/1d5ed64f-bef1-4338-9d82-4ecc585d1e22

benjaminleonard:
The delay _might_ work, but I can see it adding to a feeling of sluggishness.

I've seen this as a pattern, that's toast-esque but lower impact.

![image](https://github.com/user-attachments/assets/9b5eb675-888a-4ab9-956b-de47bbd3e08b)

We could also add a tooltip thats attached to the more info menu.

![image](https://github.com/user-attachments/assets/c29e9c01-a9d9-4374-bef7-9f1f32bba026)

We'd need to think how we tie it into the other copy interactions.

PR #2768, opened 2025-03-27
Delete affinity groups / remove affinity group members

charliepark:
This PR handles the deleting of affinity groups and the removal of affinity group members. Creation / editing flows will add a good bit more as we add modals to handle those operations, so I'm inclined to keep this one scoped to just the delete actions.

Screenshot 2025-03-27 at 4 36 20 PM
Screenshot 2025-03-27 at 10 59 42 PM
Screenshot 2025-03-27 at 4 36 37 PM
Screenshot 2025-03-27 at 4 45 42 PM

### DIFF
--- a/app/pages/project/affinity/AffinityPage.tsx
+++ b/app/pages/project/affinity/AffinityPage.tsx
@@ -127,16 +127,10 @@ export default function AffinityPage() {
               }),
             modalTitle: 'Delete anti-affinity group',
             modalContent: (
-              <>
-                <p>
-                  Are you sure you want to delete the anti-affinity group{' '}
-                  <HL>{antiAffinityGroup.name}</HL>?
-                </p>
-                <p>
-                  Future placement of the affinity group’s members will not attempt to
-                  satisfy the affinity rules.
-                </p>
-              </>
+              <p>
+                Are you sure you want to delete the anti-affinity group{' '}
+                <HL>{antiAffinityGroup.name}</HL>?
+              </p>
             ),
             errorTitle: `Error removing ${antiAffinityGroup.name}`,
           })
--- a/app/pages/project/affinity/AffinityPage.tsx
+++ b/app/pages/project/affinity/AffinityPage.tsx
@@ -24,7 +24,7 @@ import { Affinity24Icon } from '@oxide/design-system/icons/react'
 
 import { HL } from '~/components/HL'
 import { getProjectSelector, useProjectSelector } from '~/hooks/use-params'
-import { confirmDelete } from '~/stores/confirm-delete'
+import { confirmAction } from '~/stores/confirm-action'
 import { addToast } from '~/stores/toast'
 import { EmptyCell, SkeletonCell } from '~/table/cells/EmptyCell'
 import { makeLinkCell } from '~/table/cells/LinkCell'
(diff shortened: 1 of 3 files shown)


david-crespo:
> I can't say I love the words in the confirmation modals, but presume the specific phrases are important.

nah

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -98,8 +103,54 @@ const staticCols = [
...
+            modalContent: (
+              <>
+                <p>
+                  Are you sure you want to delete the anti-affinity group{' '}
+                  <HL>{antiAffinityGroup.name}</HL>?
+                </p>
+                <p>
+                  Future placement of the affinity group’s members will not attempt to
+                  satisfy the affinity rules.
+                </p>
```
david-crespo, line 141: I think we could probably just leave the line off, but if we keep it, the direction I'd go is something like:

> Member instances will no longer be kept apart in future placements, including starts or migrations.

One downside of this phrasing is that they could conceivably be in another group that also keeps them apart. So technically we can't say they won't be kept apart, only that this group is no longer keeping them apart.

> Member instances will no longer be kept apart in future placements unless they are in another anti-affinity group.

This is a bit silly. The use of "placements" I think is ok assuming we will have copy elsewhere (and docs, of course) explaining it, for example in the add instance form and/or the group create form.

app/pages/project/affinity/AntiAffinityGroupPage.tsx
```diff
@@ -84,15 +89,71 @@ export default function AntiAffinityPage() {
...
+    }
+  )
+
+  const makeActions = useCallback(
+    (antiAffinityGroupMember: AntiAffinityGroupMember): MenuAction[] => [
+      {
+        label: 'Copy instance ID',
+        onActivate() {
+          navigator.clipboard.writeText(antiAffinityGroupMember.value.id)
+          addToast('ID copied to clipboard')
```
david-crespo, line 110: We don't do the toast on any of the other copy ID row actions, as far as I can tell. I think you're right that it would be better to have some feedback, but the toast is not ideal because it's so far away. See #2416

mock-api/msw/handlers.ts
```diff
@@ -1637,6 +1637,16 @@ export const handlers = makeHandlers({
...
   antiAffinityGroupView: ({ path, query }) =>
     lookup.antiAffinityGroup({ ...path, ...query }),
+  antiAffinityGroupDelete: ({ path, query }) => {
+    const antiAffinityGroup = lookup.antiAffinityGroup({ ...path, ...query })
+    db.antiAffinityGroups = db.antiAffinityGroups.filter(
+      (i) =>
+        !(
+          i.name === antiAffinityGroup.name && i.project_id === antiAffinityGroup.project_id
+        )
+    )
```
david-crespo, line 1647: use the ID

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -98,8 +103,54 @@ const staticCols = [
 
 export default function AffinityPage() {
   const { project } = useProjectSelector()
-  const columns = useMemo(
-    () => [
+  const { data } = usePrefetchedQuery(antiAffinityGroupList({ project }).optionsFn())
```
david-crespo, line 106: Small thing, but because we're not passing these calls to `useQueryTable` because there's no pagination, you can use `apiq` instead of `getListQFn` to make this helper, and then you can drop the `.optionsFn()` everywhere.

david-crespo: I made the change since it's trivial.

PR #2772, opened 2025-04-01
Surf it, scroll it, pause it, click it

benjaminleonard:
How does this ~~3~~ 38 line CSS PR somehow add sound effects to all the buttons? Beats me.

### DIFF
--- a/app/components/DocsPopover.tsx
+++ b/app/components/DocsPopover.tsx
@@ -45,8 +45,10 @@ type DocsPopoverProps = {
 export const DocsPopover = ({ heading, icon, summary, links }: DocsPopoverProps) => {
   return (
     <Popover>
-      <PopoverButton className={cn(buttonStyle({ size: 'sm', variant: 'ghost' }), 'w-8')}>
-        <Info16Icon aria-label="Links to docs" className="shrink-0" />
+      <PopoverButton>
+        <div className={cn(buttonStyle({ size: 'sm', variant: 'ghost' }), 'w-8')}>
+          <Info16Icon aria-label="Links to docs" className="shrink-0" />
+        </div>
       </PopoverButton>
       <PopoverPanel
         // popover-panel needed for enter animation
--- a/app/ui/styles/components/button.css
+++ b/app/ui/styles/components/button.css
@@ -10,7 +10,7 @@
   @apply relative;
 }
 .active-clicked:active:not(.visually-disabled) {
-  @apply translate-y-px;
+  @apply motion-safe:translate-y-px;
 }
 .ox-button:after {
   content: '';
--- a/app/ui/styles/components/button.css
+++ b/app/ui/styles/components/button.css
@@ -9,7 +9,7 @@
 .ox-button {
   @apply relative;
 }
-.ox-button:active {
+.ox-button:active:not(.visually-disabled) {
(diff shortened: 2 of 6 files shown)


david-crespo:
Very cool. I like it. Not sure how I feel about the dropdown buttons getting it, feels slightly weird. Also the docs popover button is missing it.

https://github.com/user-attachments/assets/75b1f317-0b21-4103-832e-5f431bfceed2

app/ui/styles/components/button.css
```diff
@@ -9,6 +9,9 @@
 .ox-button {
   @apply relative;
 }
+.active-clicked:active:not(.visually-disabled) {
+  @apply translate-y-px;
+}
```
david-crespo, line 14: It's a tiny thing but how about a `motion-safe:`?

david-crespo:
Definitely better without moving the menu. Another little issue: when I tap on my trackpad, sometimes it's fast enough to not move perceptibly, but other times I get this twitch. Interestingly the video doesn't really capture it, maybe because it's double size? It looks more jerky IRL.

https://github.com/user-attachments/assets/54d42036-32e3-4777-9b79-7beb3b1a5b8f

david-crespo:
This works pretty well to do the opposite — make sure it doesn't move in case of super short presses. I don't think we can make it work the other with just `:active`.

```css
.active-clicked {
   @apply transition-transform duration-[50ms] ease-[cubic-bezier(0.65,0,0.35,1)];
}
```

benjaminleonard:
That felt a little weirder to me, like it was occasionally completely broken.

This might be a bit of a browser limitation. I tried a combination of delays / durations / animations to no real benefit. The problem I think is that the application of active and removal is too quick to transition.

We could in theory do with js but that kills the simplicity of it, and probably unnecessary overhead. You might just be an exceptionally quick clicker.

david-crespo:
I think it's actually the way the tap clicks work on my MBP trackpad. When I tap to click rather than fully pressing, it's not about how fast I do it — it does not appear possible to do it slowly enough to see the full animation. If I tap slower it stops being a click at all.

app/components/DocsPopover.tsx
```diff
@@ -46,11 +46,10 @@ export const DocsPopover = ({ heading, icon, summary, links }: DocsPopoverProps)
...
     <Popover>
-      <PopoverButton
-        className={cn(buttonStyle({ size: 'sm', variant: 'ghost' }), 'w-8')}
-        title={title}
-      >
-        <Info16Icon aria-hidden className="shrink-0" />
+      <PopoverButton title={title}>
+        <div className={cn(buttonStyle({ size: 'sm', variant: 'ghost' }), 'w-8')}>
+          <Info16Icon aria-hidden className="shrink-0" />
+        </div>
```
david-crespo, line 52: This one doesn't work, not sure why. When I move `buttonStyle` to the `PopoverButton`, I can get it to move when I check `:active` in the dev tools, but that doesn't happen when I actually click it.

benjaminleonard: This works for me, can you still reproduce?

david-crespo: Turns out it's only broken in Firefox. I think we can live with that.

Issue #2771, opened 2025-03-31
Create / Edit punchlist for Anti-affinity groups


charliepark:
- [ ] create anti-affinity group button and modal
- [ ] add member instance to anti-affinity group
  - [ ] from group page
  - [ ] from instance's Settings tab
- [ ] remove instance from anti-affinity group
  - [ ] from group page
  - [ ] from instance's Settings tab
- [ ] edit group (name, description)

- Note: Will be tricky explaining `policy`
- ~Note: Any time we need to list a combobox of instances that are available for adding to an anti-affinity group, we'll also need to check all affinity (note: affinity, not anti-affinity) groups, and to make sure that any instances already in an affinity group, if used in this anti-affinity group, isn't available for use.~ Turns out this is not actually true for now.

Issue #1834, opened 2023-12-05
Display resource IDs


karencfv:
It would be nice to be able to see the ID of a resource. Often times after I've copied the ID of several resources I want to make sure I have the right ones or whatever and it'd be nice to be able to see it directly on the console. Otherwise I have to copy and paste again, and that feels a bit off. Especially if I want to verify several IDs.

Additionally, it's not super obvious that you can copy the ID from the 3 little dots on the "resource list" page the first time you do it. I spent a bit of time looking for the ID on the individual resource's page itself.

leftwo:
+1 on this, yes. I want it as well :).  Thanks for filing.

benjaminleonard:
Are you looking to have this in the table views?

karencfv:
Thanks @benjaminleonard :) That'd be nice yeah! In both the resource's list and view pages

PR #2775, opened 2025-04-02
Add create and edit functionality to anti-affinity groups

charliepark:
This PR adds some functionality to anti-affinity groups, mainly in the "adding and removing instances" vein. Also, editing an anti-affinity group's name and description.

Once this PR is merged, the following items from https://github.com/oxidecomputer/console/issues/2771 will be functional.
- [X] create anti-affinity group button and modal
- [x] add member instance to anti-affinity group
  - [X] from group page
  - ❌ from instance's Settings tab
- [x] remove instance from anti-affinity group
  - [X] from group page
  - ❌ from instance's Settings tab
- [X] edit group (name, description)

Screenshot 2025-04-02 at 4 53 54 PM
Screenshot 2025-04-02 at 4 54 01 PM
Screenshot 2025-04-02 at 4 54 39 PM
Screenshot 2025-04-02 at 4 54 50 PM


app/forms/anti-affinity-group-create.tsx
```diff
@@ -0,0 +1,106 @@
...
+export const handle = titleCrumb('New anti-affinity group')
+
+export async function clientLoader({ params }: LoaderFunctionArgs) {
+  const { project } = getProjectSelector(params)
+  await Promise.all([
+    queryClient.prefetchQuery(antiAffinityGroupList({ project })),
+    queryClient.prefetchQuery(affinityGroupList({ project })),
+  ])
+  return null
+}
```
david-crespo, line 39: Waiting for these in the loader means there's a loading bar when you click the button to open the form. Probably better not to prefetch (or to kick off prefetch in the loader but not `await` it) so that the form opens instantly. The only annoying thing is you probably have to disable submit until these requests come back. The loading time shouldn't be noticeable because nobody is filling the form in under a second (except Playwright).

app/table/columns/common.tsx
```diff
@@ -21,6 +23,15 @@ function dateCell(info: Info<Date>) {
   return <DateTime date={info.getValue()} />
 }
 
+function idCell(info: Info<string>) {
+  return (
+    <span className="flex max-w-full items-center gap-0.5">
+      <Truncate text={info.getValue()} maxLength={14} position="middle" />
```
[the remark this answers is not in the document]
david-crespo: Might not need to truncate it — you have to be on a pretty narrow view for it to not have room. I'd say yes to getting rid of the copy ID row action if we have it right there. I find the ID a little jarring to see there but I probably just need to get used to it and we need to add it in a lot more places. I get the argument for having it and not hiding it in the copy ID row action.

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -64,13 +58,26 @@ export async function clientLoader({ params }: LoaderFunctionArgs) {
...
+              to={pb.antiAffinityGroupEdit({ project, antiAffinityGroup: name })}
+            >
+              Edit
+            </DropdownMenu.LinkItem>
+          </MoreActionsMenu>
+        )}
+      </div>
+    </PageHeader>
+  )
+}
```
david-crespo, line 80: I don't think it's worth sharing this header between the two call sites and having this name logic. I'd just inline the appropriate thing at each all site.

mock-api/msw/handlers.ts
```diff
@@ -1667,6 +1699,35 @@ export const handlers = makeHandlers({
...
+        name: instance.name,
+        run_state: instance.run_state,
+      },
+    }
+    db.antiAffinityGroupMemberLists.push({
+      anti_affinity_group_id: antiAffinityGroup.id,
+      anti_affinity_group_member: { type: 'instance', id: instance.id },
+    })
+    return json(newMember, { status: 201 })
+  },
```
david-crespo, line 1730: I think instances can only be added to a group while stopped ([omicron source](https://github.com/oxidecomputer/omicron/blob/1a1f2c6865b7a31ef223e0bd3520ca628416dd35/nexus/db-queries/src/db/datastore/affinity.rs#L823-L876)). We should enforce that here.

In the UI I'm not so sure. It's a little weird to not see instances in the picker list that you expect to be there, just because they're running. Maybe we have to add copy to the add instance modal explaining that only stopped instances will show up in the list.

david-crespo (approved):

app/pages/project/affinity/AntiAffinityGroupPage.tsx
```diff
@@ -172,11 +202,25 @@ export default function AntiAffinityPage() {
         <CardBlock.Header
           title="Members"
           description="Instances in this anti-affinity group"
-        />
+        >
+          <CreateButton
```
benjaminleonard, line 206: Let's use a regular button without the add roundel. We're not currently using it inline in these card blocks.

mock-api/msw/handlers.ts
benjaminleonard: We could show a message and disable the submit if it's running. The slight complicating factor is that at some point we might allow adding running instances, should the policy allow it – so we don't want to incept too strongly the current limitation.

app/pages/project/affinity/AffinityPage.tsx
```diff
@@ -95,12 +80,16 @@ const staticCols = [
     header: 'members',
     cell: (info) => <AffinityGroupMembersCell antiAffinityGroup={info.getValue()} />,
   }),
+  colHelper.accessor('id', Columns.id),
```
david-crespo, line 83: I don't think having the ID column here and in almost no other table makes sense. The ID is on the detail page and can be copied in the row actions. I think we should address this everywhere at once later. This is #1834.

app/layouts/ProjectLayoutBase.tsx
```diff
@@ -115,10 +115,10 @@ export function ProjectLayoutBase({ overrideContentPane }: ProjectLayoutProps) {
             <IpGlobal16Icon /> Floating IPs
           </NavLinkItem>
           <NavLinkItem to={pb.affinity(projectSelector)}>
-            <Affinity16Icon title="Affinity" /> Affinity
+            <Affinity16Icon /> Affinity
           </NavLinkItem>
           <NavLinkItem to={pb.projectAccess(projectSelector)}>
-            <Access16Icon title="Access" /> Access
+            <Access16Icon /> Access
```
david-crespo, line 121: Got rid of the titles here — I think they may actually make accessibility worse by duplicating the text.

image

david-crespo:
It didn't feel right with anti-affinity in the titles. I changed it back to affinity.

image

app/pages/project/affinity/AntiAffinityGroupPage.tsx
```diff
@@ -152,9 +156,35 @@ export default function AntiAffinityPage() {
     getCoreRowModel: getCoreRowModel(),
   })
 
+  const disabledReason = () => {
+    // https://github.com/oxidecomputer/omicron/blob/main/nexus/db-queries/src/db/datastore/affinity.rs#L66
```
david-crespo, line 160: link to a commit (press `y` to get a canonical URL) so the line number is guaranteed to stay correct

app/forms/anti-affinity-group-member-add.tsx
```diff
@@ -0,0 +1,88 @@
...
+            path: {
+              antiAffinityGroup,
+              instance: form.getValues('instance'),
+            },
+            query: { project },
+          })
+        }
+        actionText="Add to group"
+      />
+    </Modal>
```
david-crespo, line 86: Using a `<form>` and going through the form hook's `handleSubmit` is better because it makes stuff work like pressing enter in the combobox to submit the modal. Example:

https://github.com/oxidecomputer/console/blob/ab06a337f81df76cd5fcf619089e3bd4d055270c/app/pages/SiloImagesPage.tsx#L162-L205

app/forms/anti-affinity-group-member-add.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const form = useForm<AntiAffinityGroupMemberInstanceAddPathParams>({
+    defaultValues: {
+      instance: '',
+    },
+  })
+
+  const onDismiss = () => {
+    setIsModalOpen(false)
+    form.reset()
```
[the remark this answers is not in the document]
david-crespo: Oh, interesting. I tried that without the reset and I didn’t see the state persisting. I will try again. Maybe I didn’t save the file or something.

david-crespo:
Actually going with this. I was worried that it wasn't clear enough when you get to the page (or even before, since the sidebar is affected too) that what we're talking about is groups. "New group" in the button wasn't enough for me.

image

image

app/forms/anti-affinity-group-member-add.tsx
david-crespo: As we figured out in chat, `Combobox` was explicitly eating the enter keypress when the combobox options are closed (precisely the situation we care about here). In 37a673383c7d1de68f13429432ca08b93c3472df I tweaked the logic in `Combobox` to only do that when `onEnter` is passed, i.e., only when we're using the combobox in a subform (only example currently is firewall rule form) where we want the subform to submit on enter rather than the whole form.

david-crespo:
Feels a little silly to have three slightly different ways of putting it, but on the other hand that might actually help.

Screenshot 2025-04-07 at 4 19 45 PM

Screenshot 2025-04-07 at 4 19 51 PM

Screenshot 2025-04-07 at 4 19 58 PM

david-crespo (approved):
yeehaw

https://github.com/user-attachments/assets/d64f5e61-099e-46bf-92d5-2807d52c02bd

PR #2780, opened 2025-04-04
Move error into picker

### LINKED ISSUE #2778: Funky error display on system metric time picker
Looks bad whether focused or unfocused.

Image

Image

benjaminleonard:
Fixes #2778

image

### DIFF
--- a/app/components/form/fields/DateTimeRangePicker.tsx
+++ b/app/components/form/fields/DateTimeRangePicker.tsx
@@ -116,7 +116,7 @@ export function DateTimeRangePicker({
   return (
     <form className="flex">
       <Listbox
-        className="z-10 w-[10rem] border-r border-r-default [&_button]:!rounded-r-none [&_button]:!border-r-0"
+        className="z-10 w-[10rem] [&_button]:!rounded-r-none [&_button]:!border-r-0"
         name="preset"
         selected={preset}
         aria-label="Choose a time range preset"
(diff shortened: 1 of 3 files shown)


david-crespo:
Nice! Non-focused border doesn't go all the way around:

image

image

image

david-crespo:
Green dot in red day.

image

benjaminleonard:
> Nice! Non-focused border doesn't go all the way around:

It actually doesnt go round in either, the ring is tricking you. Will take a look at both.

david-crespo:
Out of scope for this PR because it's about the picker state logic, but I find this interaction weird. I make the range invalid by setting end time before start time, which makes sense, but then the way we keep the invalid range around is kind of odd: if I click two different days, then we get a valid range because of course `4/3 10pm - 4/4 9pm` is a valid range. But then I click the same day twice and it becomes invalid again because `4/3 10pm - 4/3 9pm` is invalid. I understand the logic, and maybe it's just consistent and there isn't a better way, but I feel like maybe we could be more aggressive in resetting the times, _or_ we could make clearer visually that the start time is associated with the start date, maybe by displaying the date above the times. Right now it feels a little disjointed because you have to look at both the input at the bottom of the popover and the thing at the top to really understand what you've chosen. The times at the bottom look like a time range within a single day, which is only true when the start and end days are the same.

https://github.com/user-attachments/assets/f3b74f6d-7c56-43a5-9d1a-47a9d8c0f813

benjaminleonard:
> Out of scope for this PR because it's about the picker state logic, but I find this interaction weird. 

I think we could make a general issue for datepicker nits now we've sat with the current implementation for a while.

benjaminleonard:
Today dot and focused border should be fixed now

david-crespo (approved):
fantastic, huge improvement

PR #2788, opened 2025-04-08
Add instance to anti-affinity group in create form

benjaminleonard:
Configuration might not be the best place in the advanced accordion to put it, but not sure if it justifies its own section. Added to the `instanceCreate` handler and a e2e test.

image


app/forms/instance-create.tsx
```diff
@@ -169,6 +173,9 @@ export async function clientLoader({ params }: LoaderFunctionArgs) {
     apiQueryClient.prefetchQuery('currentUserSshKeyList', {}),
     apiQueryClient.prefetchQuery('projectIpPoolList', { query: { limit: ALL_ISH } }),
     apiQueryClient.prefetchQuery('floatingIpList', { query: { project, limit: ALL_ISH } }),
+    apiQueryClient.prefetchQuery('antiAffinityGroupList', {
```
benjaminleonard, line 176: Number of groups is likely to be low. Curious if there is a threshold where its better to move this stuff after the page is loaded.

david-crespo:
What if the combobox was inline instead of in a modal?

benjaminleonard:
Worth a try, but we would need to change the floating IPs one also

benjaminleonard:
There's a few other things on that, e.g. do we do that for attach disks also? Do we stack the listbox and button (like firewall rules)? For now I'd say we defer that, since I've been trying to figure out a minitable and integrated input pattern for the webhooks work and also solve confusion around fields not being added in the firewall rules modal.

david-crespo:
This needs to be updated because we no longer have the advanced and configuration accordions. I'm also not sure we need it considering we've gone this long without it. But it was good as-is, maybe we can just figure out how to make it not stick out too much.

PR #2790, opened 2025-04-08
Disable add-instance-to-anti-affinity-group when instance not stopped

charliepark:
An instance can only be added to an anti-affinity group if that instance is stopped.

We tried a few treatments to the "add an instance to this anti-affinity group" modal, but they were all rather awkward.
- Not including running instances made it look like instances were getting lost.
- Adding badges to indicate an instances `runState` got really noisy.

We decided to list all of the instances not currently in the anti-affinity group, and to disable the submit button if a non-stopped instance were selected.

add_instance

### DIFF
--- a/app/api/util.ts
+++ b/app/api/util.ts
@@ -132,9 +132,9 @@ const instanceActions = {
   // https://github.com/oxidecomputer/omicron/blob/6dd9802/nexus/src/app/instance.rs#L1520-L1522
   serialConsole: ['running', 'rebooting', 'migrating', 'repairing'],
 
-  // https://github.com/oxidecomputer/omicron/blob/5e27bde/nexus/src/app/affinity.rs#L357 checks to see that there's no VMM
-  // TODO: determine whether the intent is only `stopped` or also `failed`
-  addToAntiAffinityGroup: ['stopped'],
+  // check to see that there's no VMM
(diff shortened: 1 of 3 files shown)


app/api/util.ts
```diff
@@ -132,9 +132,9 @@ const instanceActions = {
   // https://github.com/oxidecomputer/omicron/blob/6dd9802/nexus/src/app/instance.rs#L1520-L1522
   serialConsole: ['running', 'rebooting', 'migrating', 'repairing'],
 
-  // https://github.com/oxidecomputer/omicron/blob/5e27bde/nexus/src/app/affinity.rs#L357 checks to see that there's no VMM
-  // TODO: determine whether the intent is only `stopped` or also `failed`
-  addToAntiAffinityGroup: ['stopped'],
+  // check to see that there's no VMM
+  // https://github.com/oxidecomputer/omicron/blob/c496683/nexus/db-queries/src/db/datastore/affinity.rs#L1025-L1034
+  addToAffinityGroup: ['stopped'],
```
david-crespo, line 137: renamed because this will be the same for both kinds of group

david-crespo (approved):
Merging for now because I am like 90% confident the `stopped` thing is correct even though we don't have confirmation yet

PR #2792, opened 2025-04-08
Prevent adding instance to A-A group when not stopped

charliepark:
This sets it so the user can't add the instance to an anti-affinity group when the instance is not stopped.


david-crespo:
Thanks!

PR #2799, opened 2025-04-09
Add text-raise class to FieldLabels

### LINKED ISSUE #2787: Fix inconsistent input title colour
Some are default and some are raise:

Image

charliepark:
This adds the `text-raise` class to the FieldLabel component, which is used for many/most of the input labels.

Turns this …
Screenshot 2025-04-09 at 4 36 16 PM

to this …
Screenshot 2025-04-09 at 4 36 26 PM

Closes #2787, I believe, but Ben let me know if I misunderstood that.

### DIFF
--- a/app/forms/idp/shared.tsx
+++ b/app/forms/idp/shared.tsx
@@ -41,7 +41,7 @@ export function MetadataSourceField({
   } = useController({ control, name: 'idpMetadataSource' })
   return (
     <fieldset>
-      <legend id="metadata-source-legend" className="mb-2 text-sans-md">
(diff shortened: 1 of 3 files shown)


benjaminleonard:
Delicioso

benjaminleonard (approved):
Nice, might be worth also checking if there are inputs that aren't using this that should be?

david-crespo:
I poked around all the weirder and more bespoke forms (firewall rules, IAM stuff) and the only things I had to fix were:

image

image

PR #2801, opened 2025-04-10
Debounce `startTime` and `endTime` in date range picker

david-crespo:
I noticed on dogfood it's too easy to overload OxQL by holding the up arrow in the time picker. Input wasn't debounced, so we were firing a request for every value of the input. I couldn't believe how easy this was to do. 400ms seems quite responsive while fully solving the problem.

### Before

image

https://github.com/user-attachments/assets/0c045c78-466d-4426-a7da-a4e60e9da88b

### After

https://github.com/user-attachments/assets/05f1bfcf-ef54-48a2-b05d-98c5b81b050b


benjaminleonard (approved):
Feels like a nice balance, nice.

PR #2803, opened 2025-04-14
Thinner small rows

benjaminleonard:
Easter present for @david-crespo. Slightly thinner column cells. I tried `h-10` but that was a bridge too far.

In contrast the taller rows could do with a design pass. In some places at least these could be converted to smaller rows.

image

### DIFF
--- a/app/pages/system/networking/IpPoolsPage.tsx
+++ b/app/pages/system/networking/IpPoolsPage.tsx
@@ -112,6 +112,7 @@ export default function IpPoolsPage() {
   const { table, query } = useQueryTable({
     query: ipPoolList(),
     columns,
+    rowHeight: 'large',
     emptyState: <EmptyState />,
   })
   const { data: pools } = query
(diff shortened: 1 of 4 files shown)


app/pages/system/networking/IpPoolsPage.tsx
```diff
@@ -112,6 +112,7 @@ export default function IpPoolsPage() {
   const { table, query } = useQueryTable({
     query: ipPoolList(),
     columns,
+    rowHeight: 'large',
```
david-crespo, line 115: The irony is there are no IPv6 ranges, so we don't strictly need to accommodate them in the rows. I'm going to take a quick look at making this sneakier while making sure it behaves reasonably if IPv6 shows up without us noticing.

david-crespo:
## Without IPv6

Guaranteed to be true... for now.

image


## With IPv6

Looks a little silly but is not broken. I could live with myself if we accidentally enabled IPv6 ranges and forgot to update this.

image

benjaminleonard:
Or...

image

Bad, but less bad. Until I fix it.

PR #2804, opened 2025-04-14
minor: fix label and title on docs popover button

david-crespo:
It occurred to me it would be nice to get a tooltip on the docs popover button on hover. Instead of saying "Links to docs" I made it use the same title we had in the popover.

image

### DIFF
--- a/app/components/DocsPopover.tsx
+++ b/app/components/DocsPopover.tsx
@@ -36,17 +36,21 @@ export const DocsPopoverLink = ({ href, linkText }: DocsPopoverLinkProps) => (
 )
 
 type DocsPopoverProps = {
-  heading: React.ReactNode
+  heading: string
   icon: JSX.Element
   links: Array<DocsPopoverLinkProps>
   summary: string
 }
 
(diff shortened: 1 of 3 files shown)


app/components/DocsPopover.tsx
```diff
@@ -36,17 +36,21 @@ export const DocsPopoverLink = ({ href, linkText }: DocsPopoverLinkProps) => (
...
+  const title = `Learn about ${heading}`
   return (
     <Popover>
-      <PopoverButton className={cn(buttonStyle({ size: 'sm', variant: 'ghost' }), 'w-8')}>
-        <Info16Icon aria-label="Links to docs" className="shrink-0" />
+      <PopoverButton
+        className={cn(buttonStyle({ size: 'sm', variant: 'ghost' }), 'w-8')}
+        title={title}
+      >
+        <Info16Icon aria-hidden className="shrink-0" />
```
david-crespo, line 53: Confirmed this is sufficient for a11y and, indeed, likely preferable. There's no need to have the label on the icon.

image

PR #2808, opened 2025-04-23
Fix target for checkbox on instance create form

### LINKED ISSUE #2807: Fix unclickable gap between checkbox and label
This post reminded me, we have the same issue: https://blog.jim-nielsen.com/2024/sanding-ui/

charliepark:
On the instance-create table, there's a checkbox that has a gap between the checkbox and the label, mostly because we were re-creating something that already works. This update uses the default checkbox / label approach, making it work like other places in the app.
click_target

It is possible I have missed something, so will poke around a bit more to see if there are more broken checkbox/labels.

Fixes #2807

### DIFF
--- a/app/forms/instance-create.tsx
+++ b/app/forms/instance-create.tsx
@@ -737,24 +737,21 @@ const AdvancedAccordion = ({
               it is deleted
             </TipIcon>
           </h2>
-          <div className="flex items-start gap-2.5">
-            <Checkbox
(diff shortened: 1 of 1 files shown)


david-crespo:
Were there other cases @benjaminleonard? It does seem like this is the only one where we used `<label>` like that.

david-crespo (approved):

PR #2811, opened 2025-04-29
Add component for empty MiniTable

### LINKED ISSUE #2661: Minitable empty state
Currently the mini table is not rendered if it has no items.

An empty state here would be a bit clearer, and would also help mitigate #2627 by showing the user that it has not yet been selected.

Image

I tried versions with and without the icon. The icon is nice, but I think it's inclusion would cause extra layout shift when the first item is added which might be distracting.

charliepark:
This adds an empty state for MiniTables. One hope is that this resolves (or at least mitigates) #2627.

Screenshot 2025-04-29 at 12 09 27 PM

The no-whitespace-diffs view is a bit easier to glance at: https://github.com/oxidecomputer/console/pull/2811/files?diff=split&w=1

Closes #2661

### DIFF
--- a/app/components/form/fields/DisksTableField.tsx
+++ b/app/components/form/fields/DisksTableField.tsx
@@ -13,6 +13,7 @@ import type { DiskCreate } from '@oxide/api'
 import { AttachDiskModalForm } from '~/forms/disk-attach'
 import { CreateDiskSideModalForm } from '~/forms/disk-create'
 import type { InstanceCreateInput } from '~/forms/instance-create'
+import { EmptyCell } from '~/table/cells/EmptyCell'
 import { Badge } from '~/ui/lib/Badge'
 import { Button } from '~/ui/lib/Button'
 import * as MiniTable from '~/ui/lib/MiniTable'
@@ -71,7 +72,7 @@ export function DisksTableField({
                   </MiniTable.Cell>
                   <MiniTable.Cell>
                     {item.type === 'attach' ? (
-                      '—'
+                      <EmptyCell />
                     ) : (
                       <>
                         <span>{bytesToGiB(item.size)}</span>
--- a/app/ui/styles/components/mini-table.css
+++ b/app/ui/styles/components/mini-table.css
@@ -54,10 +54,7 @@
 
   /* the delete button */
   & td:last-child > div > button {
-    @apply -m-2 flex items-center justify-center p-2 text-tertiary;
-  }
-  & td:last-child > div:has(button:hover, button:focus) {
-    @apply text-default;
+    @apply -m-2 flex items-center justify-center p-2 text-tertiary hover:text-secondary focus:text-secondary;
   }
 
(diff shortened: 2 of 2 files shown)


benjaminleonard:
I know @david-crespo likes components to be more declarative, but I wonder if the current mini table implementation leaves room for accidental differences to sneak in.

I say that here because we could pass items to the mini table and let it handle the empty state within the component automatically.

We're also going to have two versions with these webhook changes, one that is pretty much the same as it is now and one with integrated inputs. In this case would we keep the existing component and create a wrapper component that takes props directly instead?

app/ui/lib/MiniTable.tsx
```diff
@@ -36,6 +37,37 @@ export const Cell = ({ children }: Children) => {
...
+  <Row>
+    <td colSpan={props.colSpan}>
+      <div className="!m-0 !w-full !flex-col !border-none !bg-transparent !py-14">
+        <EmptyMessage title={props.title} body={props.body} />
+      </div>
+    </td>
+  </Row>
+)
+
+export const InputCell = ({
```
charliepark, line 50: Not used in this branch, but will be used in follow-up

app/components/form/fields/DisksTableField.tsx
```diff
@@ -67,15 +67,15 @@ export function DisksTableField({
                     <Truncate text={item.name} maxLength={35} />
                   </MiniTable.Cell>
                   <MiniTable.Cell>
-                    <Badge variant="solid">{item.type}</Badge>
+                    <Badge>{item.type}</Badge>
                   </MiniTable.Cell>
                   <MiniTable.Cell>
                     {item.type === 'attach' ? (
                       '—'
```
benjaminleonard: Should use the `EmptyCell` component here I think for consistency.

app/components/form/fields/DisksTableField.tsx
```diff
@@ -67,15 +67,15 @@ export function DisksTableField({
...
                     ) : (
                       <>
                         <span>{bytesToGiB(item.size)}</span>
```
benjaminleonard, line 78: Not a change we need to make now, but should we make a size component that ensures the item and the unit are handled consistently / spacing and text colour.

app/ui/styles/components/mini-table.css
```diff
@@ -11,57 +11,67 @@
...
+  /* last cell's div (the div for the delete button) */
   & td:last-child > div {
-    @apply w-12 justify-center pl-0 pr-0;
+    @apply flex w-8 items-center justify-center border-none px-5;
   }
+
+  /* the delete button */
   & td:last-child > div > button {
-    @apply -mx-3 -my-3 flex items-center justify-center px-3 py-3;
+    @apply -m-2 flex items-center justify-center p-2 text-tertiary;
```
benjaminleonard: Can we add `hover:text-secondary`

PR #2815, opened 2025-05-02
Full page form spacing

benjaminleonard:
One small thing, one small but open for discussion.

1. Tighten divider spacing, side modal and full page form are now the same
2. Removing the highest preset, if the user wants these values they can do so in the custom tab and this wraps awkwardly

### DIFF
--- a/app/ui/styles/components/form.css
+++ b/app/ui/styles/components/form.css
@@ -2,7 +2,7 @@
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, you can obtain one at https://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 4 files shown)


app/ui/styles/components/form.css
```diff
@@ -22,10 +22,7 @@
   @apply space-y-6;
 }
 
-.ox-form.is-side-modal .ox-divider {
+.ox-form .ox-divider,
+.ox-form .ox-divider:not([hidden]) {
```
benjaminleonard, line 26: I have this to avoid `!my-10` since the `space-y-...` applies it and is therefore more specific. Maybe `!` would be better? Unsure

david-crespo (approved):
2025-05-02-no-space-only-text

PR #2818, opened 2025-05-02
Bump omicron to latest dogfood deploy commit

david-crespo:


david-crespo:
Wow! Three e2e failures in a row on Firefox. [First](https://github.com/oxidecomputer/console/actions/runs/14802804216/job/41565274548?pr=2818) was one I don't think I've seen before: `test/e2e/ip-pools.e2e.ts:319:1 › no ranges means no utilization bar`.

2 and 3 were `test/e2e/firewall-rules.e2e.ts:151:1 › firewall rule form targets table`, which we see all the time, see #2704.

PR #2821, opened 2025-05-06
Regen API client with proper nullable fields

### LINKED ISSUE #2820: Runtime error due to unhandled API `null` auto restart policy
The whole instance page blows up when an instance is marked failed with an unset auto restart policy. The error is thrown by this ts-pattern `match` call, which blows up to tell you you have an unhandled case.

https://github.com/oxidecomputer/console/blob/f097ec960c43378f1469850d9679a1026babb1a8/app/components/InstanceAutoRestartPopover.tsx#L65-L73

Strictly speaking this is only a manifestation of the real bug in the TS client generator, which does not generate `| null` on types that can in fact be `null`. This means that in the console we do not handle the possibility of `null` adequately.

https://github.com/oxidecomputer/console/blob/f097ec960c43378f1469850d9679a1026babb1a8/app/api/__generated__/Api.ts#L1927-L1927

david-crespo:
Closes #2820 

Note that while there are many lines of changes to the generated types and zod validators (the latter only used in the mock API), the only runtime app change is the change from `undefined` to `P.nullish` in the offending `match` in the auto restart popover.

### DIFF
--- a/app/components/InstanceAutoRestartPopover.tsx
+++ b/app/components/InstanceAutoRestartPopover.tsx
@@ -9,7 +9,7 @@ import { CloseButton, Popover, PopoverButton, PopoverPanel } from '@headlessui/r
 import { formatDistanceToNow } from 'date-fns'
 import { type ReactNode } from 'react'
 import { Link } from 'react-router'
-import { match } from 'ts-pattern'
+import { match, P } from 'ts-pattern'
 
 import {
   AutoRestart12Icon,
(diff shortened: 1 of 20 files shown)


app/components/InstanceAutoRestartPopover.tsx
```diff
@@ -69,7 +69,7 @@ export const InstanceAutoRestartPopover = ({ instance }: { instance: Instance })
                 </Badge>
               ))
               .with('best_effort', () => <Badge>best effort</Badge>)
-              .with(undefined, () => <Badge color="neutral">Default</Badge>)
+              .with(P.nullish, () => <Badge color="neutral">Default</Badge>)
```
david-crespo, line 72: Jerk!

david-crespo: Confirmed a test fails with a `null` policy without the fix, pass with the fix.

Screenshot 2025-05-06 at 5 56 36 PM

PR #2825, opened 2025-05-15
Bump max instance RAM to 1024 GiB

### LINKED ISSUE #2824: Bump up maximum VM memory size to 1024 GiB
The maximum is raised to 1024 GiB with the fixes in https://github.com/oxidecomputer/omicron/issues/7918. Once the related PRs have landed, please change the web console validations to match with the new maximum size.

david-crespo:
Closes #2824 

Turns out we don't have any tests exercising these limits, and while testing it I noticed that we do something really annoying on instance create if you type in too big of a RAM amount. Unlike in the resize modal, where we correctly show an error on blur, on instance create we just silently change the value to the max.

image

### DIFF
--- a/app/api/util.ts
+++ b/app/api/util.ts
@@ -23,13 +23,13 @@ import type {
   VpcFirewallRuleUpdate,
 } from './__generated__/Api'
 
-// API limits encoded in https://github.com/oxidecomputer/omicron/blob/main/nexus/src/app/mod.rs
+// API limits encoded in https://github.com/oxidecomputer/omicron/blob/b7af5f8e/nexus/src/app/mod.rs
 
 export const MAX_NICS_PER_INSTANCE = 8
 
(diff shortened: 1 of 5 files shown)


david-crespo:
> The minValue and maxValue props can be used to limit the entered value to a specific range. The value will be clamped when the user blurs the input field.

https://react-spectrum.adobe.com/react-aria/useNumberField.html#minimum-and-maximum-values

The difference between instance create and resize is that we are not passing `max` in the resize one (relying solely on the explicit validation function to validate on blur) so it is not doing the clamping. The downside here is that the up button doesn't get disabled when you reach the max, which is cute. If forced to choose, I'd rather have validate on blur than disabled button, but I'll see if we can get both.

david-crespo:
Tolerable IMO even though the up button (and up arrow key) isn't disabled. At these numbers you're probably not using them anyway.

image

PR #2826, opened 2025-05-16
Tailwind v4

benjaminleonard:
From https://github.com/oxidecomputer/design-system/pull/133

We can hold on this and revisit until we have fewer PRs since it touches so many files, I just wanted to find any issues with the design system upgrade first. It did surface a couple small issues which I've fixed.

The lack of a z-index utility is a little clunky here. And still a couple things to fix. There's some strange table behaviour and I think I need to check the complex variants are behaving as they should.


app/components/CapacityBars.tsx
```diff
@@ -23,7 +23,7 @@ export const CapacityBars = ({
   allocatedLabel: string
 }) => {
   return (
-    <div className="mb-12 flex min-w-min flex-col gap-3 lg+:flex-row">
+    <div className="1000:flex-row mb-12 flex min-w-min flex-col gap-3">
```
benjaminleonard, line 26: These breakpoints I've added to the design system, previously we had a different style for this syntax each time. The width syntax here I find much easier since I don't have to remember how wide `sm` is.

benjaminleonard:
@david-crespo Let me know if it's worth pushing this to fix all the remaining bits or leave for later

PR #2828, opened 2025-05-19
Zod v4

david-crespo:
https://zod.dev/v4

Was curious how the Zod v4 migration would look, so I tried it and it turned out to be easy enough that I finished it. This will remain a draft for now because we need to upgrade [oxide.ts](https://github.com/oxidecomputer/oxide.ts) to do automatically what I've done here by hand in the generated code:

- `z.string().uuid()` -> `z.uuid()`
- `z.string().ip()` -> `z.ipv4()` or a union of v4 and v6
- `z.record(X)` -> `z.record(z.string(), X)` (they got rid of the single-arg shorthand)
- Put `.nullable()` before `.default(null)` so `default` knows it can take `null`

### DIFF
--- a/mock-api/snapshot.ts
+++ b/mock-api/snapshot.ts
@@ -77,7 +77,7 @@ export const snapshots: Json<Snapshot>[] = [
     state: 'ready',
   },
   {
-    id: '7fc6ca11-452e-d3e4-9e1c-752ff615abea',
+    id: '75832119-7d9b-4e91-a7c6-76b996b69aaa',
     name: 'snapshot-heavy',
     description: '',
     project_id: project.id,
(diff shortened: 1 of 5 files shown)


mock-api/snapshot.ts
```diff
@@ -77,7 +77,7 @@ export const snapshots: Json<Snapshot>[] = [
     state: 'ready',
   },
   {
-    id: '7fc6ca11-452e-d3e4-9e1c-752ff615abea',
```
david-crespo, line 80: Turns out this was an invalid UUID, correctly rejected by Zod because of the `d` in `d3e4`, which is supposed to be a 4 for UUID v4.

david-crespo:
Will reopen when we get to this.

PR #2833, opened 2025-06-03
List and delete my access tokens

david-crespo:
Built on top of https://github.com/oxidecomputer/omicron/pull/8231, so that needs to merge and then we need to bump the API here to a main commit.

---

My tokens page is very straightforward — all you can do is list and delete, and those operations are extremely simple.

image

Extra messaging in confirmation explaining what happens when you kill a token:

image

Tooltip on `id` column explaining that it is not the token itself.

image

### Moved to #2834

~~Access page is a little harder — any silo viewer can see this value, and that's basically everyone, but only silo admins can edit it. We don't have a good way to detect who is a silo admin. For fleet viewer, we hit an endpoint and see if it 403s. I was thinking maybe we should add `fleet_role` and `silo_role` fields to the `/v1/me` response — that would make this easy. Not sure if there's good stopgap for v15, or **maybe we should just take this page out and leave it to the CLI for now**.~~

image

### DIFF
--- a/app/pages/settings/AccessTokensPage.tsx
+++ b/app/pages/settings/AccessTokensPage.tsx
@@ -8,13 +8,7 @@
 import { createColumnHelper } from '@tanstack/react-table'
 import { useCallback, useMemo } from 'react'
 
-import {
-  getListQFn,
-  queryClient,
-  useApiMutation,
-  useApiQueryClient,
-  type DeviceAccessToken,
-} from '@oxide/api'
+import { getListQFn, queryClient, useApiMutation, type DeviceAccessToken } from '@oxide/api'
 import { AccessToken16Icon, AccessToken24Icon } from '@oxide/design-system/icons/react'
 
 import { DocsPopover } from '~/components/DocsPopover'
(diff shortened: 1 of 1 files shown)


benjaminleonard:
I like we're best leaving out the access setting part. In part because of the permissions issue, but also that design will change, whereas the table is likely to stay as it is.

app/layouts/SiloLayout.tsx
```diff
@@ -68,7 +68,7 @@ export default function SiloLayout() {
             <Metrics16Icon /> Utilization
           </NavLinkItem>
           <NavLinkItem to={pb.siloAccess()}>
-            <Access16Icon /> Access
+            <Access16Icon /> Silo Access
```
david-crespo, line 71: I kept these renames from the silo access page changes because they're easy wins.

david-crespo:
[I've had it with these monkey-fighting flakes on this monday-to-friday plane](https://www.youtube.com/watch?v=z4t6zNZ-b0A)

PR #2838, opened 2025-06-16
Add sled policy data to sled page

### LINKED ISSUE #2837: Add sled policy data to view page
We have the `kind`, `provision_policy`, and `state` on the sled inventory list view (https://github.com/oxidecomputer/console/issues/2333) . It would be very helpful to have this data replicated on the individual sled view as well. When changing the sled state I am often looking at the sled view page, and it would be nice to be able to verify the state change on that page instead of going back to the list page.

charliepark:
This PR adds policy data (already present on the sleds overview page) to the individual sled view page.

Before:
image

After:
Screenshot 2025-06-16 at 10 50 41 AM

This does lose the `Location | Coming soon` line, but I suspect that was added as a means of balancing the rows.
It also adjusts the order of cells in a way that makes more sense on a larger window (like having the two "Usable _____" cells in the bottom of the right-hand column), but as most of our layouts are optimized for desktops over mobile, this seemed like it made more sense.

Closes #2837

### DIFF
--- a/app/pages/system/inventory/SledsTab.tsx
+++ b/app/pages/system/inventory/SledsTab.tsx
@@ -20,7 +20,7 @@ import { ProvisionPolicyBadge, SledKindBadge, SledStateBadge } from './sled/Sled
 const sledList = getListQFn('sledList', {})
 
 export async function clientLoader() {
-  await queryClient.prefetchQuery(sledList.optionsFn())
+  await queryClient.fetchQuery(sledList.optionsFn())
   return null
 }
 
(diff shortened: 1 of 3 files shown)


david-crespo:
We have this second layer of header in the list view to explain what "kind" means and to group kind and provision policy like they are in the response. We can't quite do that here, but calling kind "policy kind" might help. On the other hand it might look weird to have "policy kind" and "provision policy". Maybe @benjaminleonard will have an idea.

image

https://github.com/oxidecomputer/console/blob/e71c8282bb38d881eee39bc4f1a3fa6ad78443b4/app/api/__generated__/Api.ts#L3632-L3644

Issue #2831, opened 2025-05-23
Empty IP column in gateways table is misleading


david-crespo:
When you create a gateway and attach a pool, you will see an empty IP address column and it makes you think there should be something there, when in fact attaching the pool will usually be sufficient.

https://docs.oxide.computer/guides/configuring-guest-networking#internet-gateway

![Image](https://github.com/user-attachments/assets/ca0c4498-2c83-41e8-b50e-03dedd4b7162)

Details TBD, but I think we can take the IP column out of the table and let it show up only in the side modal, where we already have this helpful message. Another approach would be to leave it there but add a tooltip (either in the column header or on the "—" itself) with the same helpful message.

<img width="474" alt="Image" src="https://github.com/user-attachments/assets/6145d777-1d34-4dde-9040-ab7640ef440a" />

david-crespo:
Discussed this in chat [here](https://matrix.to/#/!oSMXTSrgambaNFEemN:oxide.computer/$-RrKkIPcHBrEt1gi9-hMfsKJkVZI8S5oNEq3_yrnckM?via=oxide.computer&via=unix.house&via=matrix.org).

PR #2839, opened 2025-06-16
Move IP Address column in Internet Gateways table

charliepark:
This moves the `Attached IP Address` column to come after the `Attached IP Pool` column, and adds a header with explanatory text.
Screenshot 2025-06-16 at 2 47 55 PM

Relates to #2831, but, per Crespo's comments below, does not fully address the root issue.

### DIFF
--- a/app/pages/project/vpcs/VpcGatewaysTab.tsx
+++ b/app/pages/project/vpcs/VpcGatewaysTab.tsx
@@ -69,17 +69,15 @@ export async function clientLoader({ params }: LoaderFunctionArgs) {
 
   await Promise.all([
     ...gateways.items.flatMap((gateway: InternetGateway) => [
-      queryClient.prefetchQuery(
+      queryClient.fetchQuery(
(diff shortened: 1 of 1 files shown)


app/pages/project/vpcs/VpcGatewaysTab.tsx
```diff
@@ -68,17 +69,15 @@ export async function clientLoader({ params }: LoaderFunctionArgs) {
 
   await Promise.all([
     ...gateways.items.flatMap((gateway: InternetGateway) => [
-      queryClient.prefetchQuery(
+      queryClient.fetchQuery(
```
charliepark, line 72: General prefetchQuery → fetchQuery migration discussed in 2616

david-crespo (approved):
Approving this as a small net improvement but I'm still really unsatisfied with the presentation, so maybe we shouldn't close #2831, or maybe we should make a new issue for general clarification on this page. It's still hard to guess from the console or the docs how any of this works (example below from my chat with Kyle of the background I consider missing, linked in #2831). I might feel better about the console if the docs were better.

image

PR #2841, opened 2025-06-27
Add data minitable component

benjaminleonard:
Proposal for an intermediary component that handles the most common `MiniTable` pattern.

The key things this is solving for:

- **Consistency**: All tables using this component will have the same structure and behaviour
- **Maintainability**: Changes to table behaviour only need to be made in one place 

Thoughts @charliepark @david-crespo 

This works as it did on the instance create page, if you think it's a worthwhile addition I can convert the others.

image

The other motivation being that as we add the input minitable component the logic is likely going to get a bit more complicated.

### DIFF
--- a/app/components/form/fields/DisksTableField.tsx
+++ b/app/components/form/fields/DisksTableField.tsx
@@ -13,7 +13,6 @@ import type { DiskCreate } from '@oxide/api'
 import { AttachDiskModalForm } from '~/forms/disk-attach'
 import { CreateDiskSideModalForm } from '~/forms/disk-create'
 import type { InstanceCreateInput } from '~/forms/instance-create'
-import { EmptyCell } from '~/table/cells/EmptyCell'
 import { sizeCellInner } from '~/table/columns/common'
 import { Badge } from '~/ui/lib/Badge'
 import { Button } from '~/ui/lib/Button'
@@ -22,7 +21,7 @@ import { Truncate } from '~/ui/lib/Truncate'
 
 export type DiskTableItem =
   | (DiskCreate & { type: 'create' })
-  | { name: string; type: 'attach' }
+  | { name: string; type: 'attach'; size: number }
 
 /**
  * Designed less for reuse, more to encapsulate logic that would otherwise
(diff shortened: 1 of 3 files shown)


david-crespo:
I think this is good, just looking through and making some tweaks. I don't think all the accessibility stuff is necessarily needed since it's a `<table>` and the browser + screenreaders know what to do with that. 

For example, I took off the `rowLabel` and tested the table with voiceover, and when you go through the cells it tells you what's in each, using the column header and the cell value together. It may be marginally better to have it on the row, or honestly it may not and it may throw screenreaders off since it's an unusual thing to do.

2025-06-27-table-voiceover

benjaminleonard:
Even simpler, fantastic. And that tracks, perhaps aria labels are more useful if you're doing something unexpected?

app/components/form/fields/NetworkInterfaceField.tsx
```diff
@@ -76,38 +76,29 @@ export function NetworkInterfaceField({
...
+                    type: 'create',
+                    params: value.params.filter((i) => i.name !== item.name),
+                  })
+                }
+                removeLabel={(item) => `remove network interface ${item.name}`}
+                // empty state not used because the table is hidden when there are none
+                emptyState={{
+                  title: 'No network interfaces',
+                  body: 'Add a network interface to see it here',
+                }}
```
david-crespo: The fact that this thing about empty state not being used is true for 4 of them suggests we should make it optional, though it's error-prone, because if we accidentally leave it out, it'll look bad when empty. Another way to do it would be to make it so that if emptyState is left out, have the table hide itself when the list is empty. That's kind of neat.

david-crespo: Well, the other thing we could do is get rid of that behavior and let the empty state do its thing. But it looks kind of huge.

image

benjaminleonard: I'll prod at that behaviour but what you did works for now and not a blocker

PR #2842, opened 2025-06-29
Show size of attached disks in mini table

### LINKED ISSUE #1662: Instance creation: Disk size is not shown for existing disks to be attached
attach-disks-missing-size
The disk size information should be available for existing disks and displayed in the disk section shown above.

charliepark:
This adds in a fix for #1662, where we aren't showing the size of the _attached_ disks in the disks minitable.

As this isn't necessary for the `data-minitable` branch, I'm happy to wait until that branch is merged and I can merge this to main, but if you want to include it in the branch, it's available.

Screenshot 2025-06-29 at 9 11 24 AM

Closes #1662.

### DIFF
--- a/app/components/form/fields/DisksTableField.tsx
+++ b/app/components/form/fields/DisksTableField.tsx
@@ -8,12 +8,11 @@
 import { useState } from 'react'
 import { useController, type Control } from 'react-hook-form'
 
-import type { Disk, DiskCreate } from '@oxide/api'
+import type { DiskCreate } from '@oxide/api'
 
 import { AttachDiskModalForm } from '~/forms/disk-attach'
 import { CreateDiskSideModalForm } from '~/forms/disk-create'
 import type { InstanceCreateInput } from '~/forms/instance-create'
-import { EmptyCell } from '~/table/cells/EmptyCell'
 import { sizeCellInner } from '~/table/columns/common'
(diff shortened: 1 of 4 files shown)


app/components/form/fields/DisksTableField.tsx
```diff
@@ -99,8 +100,11 @@ export function DisksTableField({
       {showDiskAttach && (
         <AttachDiskModalForm
           onDismiss={() => setShowDiskAttach(false)}
-          onSubmit={(values) => {
-            onChange([...items, { type: 'attach', ...values }])
+          onSubmit={({ name }: { name: string }) => {
+            onChange([
+              ...items,
+              { name, type: 'attach', size: allDisks.find((d) => d.name === name)?.size },
+            ])
```
david-crespo: Unfortunately `onChange` has type `(...event: any[]) => void`, which masks a kind of type error here: `size` has type `number` on `DiskTableItem`, but here it is `number | undefined`. You can see this by adding `satisfies DiskTableItem`, which is annoying but the best way I can think of to deal with this.

image

Rather than pulling the size here from the name, it can be done inside `AttachDiskModalForm` because you know that you have the disk on hand there, otherwise it would not show up in the combobox. I made the change already in order to test it out, so I'll just push that.

PR #2844, opened 2025-07-02
Vite 7 + friends

david-crespo:
https://vite.dev/blog/announcing-vite7

Next step is trying out [`rolldown-vite`](https://github.com/vitejs/rolldown-vite), which is very close to ready.

Draft because I'm seeing test failures locally.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
     "esModuleInterop": true,
     "importHelpers": true,
     "incremental": true,
+    "isolatedModules": true,
     "jsx": "react-jsx",
     "module": "es2020",
     "moduleResolution": "bundler",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -24,7 +24,7 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es2020",
+    "target": "es2022",
     "types": ["node", "vite/client", "vitest/importMeta"]
   },
   "exclude": ["tools/deno"]
--- a/mock-api/json-type.ts
+++ b/mock-api/json-type.ts
@@ -6,4 +6,4 @@
  * Copyright Oxide Computer Company
  */
 
-export { Json } from '~/api/__generated__/msw-handlers'
+export { type Json } from '~/api/__generated__/msw-handlers'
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -24,7 +24,7 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es2020",
+    "target": "es2022",
```
david-crespo, line 28: The need for this has to be related to the following, but I don't really get it because I don't see how Vite's target should affect what TS is checking, which would be the source code.

> ### Default Browser Target Changed to Baseline Widely Available
> 
> [Baseline](https://web-platform-dx.github.io/web-features/) gives us clear information about which web platform features work across their core browser set today. Baseline Widely Available indicates the feature is well-established and works across many devices and browser versions, being available across browsers for at least 30 months.
> 
> In Vite 7, the default browser target is changing from 'modules' to a new default: 'baseline-widely-available'. The set of browsers will be updated on each major to match the list of minimum browser versions compatible with Baseline Widely available features. The default browser value of build.target is changing in Vite 7.0:
> 
>     Chrome 87 → 107
>     Edge 88 → 107
>     Firefox 78 → 104
>     Safari 14.0 → 16.0
> 
> This change adds predictability to the default browser target for future releases.

https://vite.dev/blog/announcing-vite7.html#default-browser-target-changed-to-baseline-widely-available

https://github.com/vitejs/vite/pull/20007

david-crespo:
Getting a weird set of e2e failures locally but only sometimes. And it seems to be similarly intermittent on CI.

image


image

The only clue is that when it fails we seem to be getting this message. But if you run it again a second time a) it does not output the message, and b) the tests pass.

```
[WebServer] (!) Failed to run dependency scan. Skipping dependency pre-bundling. Error: failed to resolve rollupOptions.input value: "index.html".
[WebServer]     at resolvePath (file:///Users/runner/work/console/console/node_modules/vite/dist/node/chunks/dep-Bsx9IwL8.js:10766:29)
[WebServer]     at async Promise.all (index 0)
[WebServer]     at async computeEntries (file:///Users/runner/work/console/console/node_modules/vite/dist/node/chunks/dep-Bsx9IwL8.js:10771:44)
[WebServer]     at async scan (file:///Users/runner/work/console/console/node_modules/vite/dist/node/chunks/dep-Bsx9IwL8.js:10701:19)
[WebServer]     at async file:///Users/runner/work/console/console/node_modules/vite/dist/node/chunks/dep-Bsx9IwL8.js:34671:15
```

PR #2849, opened 2025-07-08
Audit Log

benjaminleonard:
Still very rough, but has been a helpful exercise in working through some of the design due to the sheer amount of information and differing layout from other pages.

CleanShot 2025-07-08 at 11 47 02

Stubbing out based on https://github.com/oxidecomputer/omicron/pull/7339.

Uses [Tanstack Virtual](https://tanstack.com/virtual/latest). On testing with > 500 lines it starts to get a bit chunky especially if you're interacting with the page (e.g. opening the row).

Hoping that silo name and actor display name can be plumbed through so those are hard-coded for now.

Still needs:
- [ ] Error state
- [ ] Loading/placeholder state
- [ ] Copy JSON to clipboard
- [ ] Equivalent CLI/API command
- [ ] Fix giant footer spacing

### DIFF
--- a/app/ui/lib/Listbox.tsx
+++ b/app/ui/lib/Listbox.tsx
@@ -101,7 +101,7 @@ export const Listbox = <Value extends string = string>({
               id={id}
               name={name}
               className={cn(
-                `flex h-11 items-center justify-between rounded border text-sans-md`,
+                `flex h-10 items-center justify-between rounded border text-sans-md`,
                 hasError
                   ? 'focus-error border-error-secondary hover:border-error'
                   : 'border-default hover:border-hover',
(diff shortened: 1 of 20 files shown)


app/ui/lib/Listbox.tsx
```diff
@@ -101,7 +101,7 @@ export const Listbox = <Value extends string = string>({
               id={id}
               name={name}
               className={cn(
-                `flex h-11 items-center justify-between rounded border text-sans-md`,
+                `flex h-10 items-center justify-between rounded border text-sans-md`,
```
benjaminleonard, line 104: Sneaking in my continued densification

PR #2850, opened 2025-07-09
Update protocol filters

### LINKED ISSUE #2846: Firewall rules should allow setting/viewing multiple ICMP filters
`VpcFirewallRuleProtocol`s can now carry more information than a protocol alone. Today this is only expressed in ICMP type/code filters, but in future this will include ICMPv6 and probably a generic selector over the IP protocol code. We should allow interaction with a list of filters, rather than the three checkboxes we have today.

https://github.com/oxidecomputer/omicron/blob/d84882702969e42be311f1b4274ab9a4ab9d28a3/openapi/nexus.json#L26196-L26251

The move from representing each as a string to a kind+value will also impact how these are serialised/deserialised.

charliepark:
There's a little bit of design input still coming on this for the mini table and badges in the Firewall Rules tab, but I'm getting to the point where I'm fiddling, so it's time to open up the draft PR.

The crux of this is a change logged in issue #2846, where the API now uses a more sophisticated object for protocol filters, rather than the earlier strings of `TCP`, `UDP`, and `ICMP`. Also, users can add multiple ICMP filters. We still continue to use `TCP` and `UDP`, but `ICMP` protocol filters now have additional information, including an optional ICMP **type** and optional **code**.

For more on ICMP types and codes, [this list of IANA code fields](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml) might be useful.

There are two main changes to the UI that a user will see with this PR.

First, the Firewall Rules tab has a new treatment for the protocol filters:
Screenshot 2025-07-09 at 7 26 12 AM
Note how the code can have either an individual number or a range.

Second, the Protocol Filters section of the Firewall Rules form has a new step-through form process, and a new mini-table:
Screenshot 2025-07-08 at 11 39 57 PM
Screenshot 2025-07-08 at 11 33 16 PM

I am open to any thoughts / tweaks / etc.

Closes #2846


app/forms/firewall-rules-common.tsx
```diff
@@ -251,17 +257,268 @@ const availableItems = (
...
+                name="icmpType"
+                control={protocolForm.control}
+                description="Select ICMP type (leave blank for all)"
+                placeholder=""
+                allowArbitraryValues
+                items={icmpTypeItems}
+                validate={icmpTypeValidation}
+              />
+
+              {selectedIcmpType !== undefined && selectedIcmpType !== 0 && (
```
david-crespo: What is this `!== 0` about?

charliepark: This is a bit of legacy code from when we were being a bit more prescriptive with the ICMP codes, and the idea was to only show the `code` combobox when the ICMP type had associated codes. Some types don't have those codes, and `0` is one of (but not the only) one. I think this was an early experiment in tailoring the UI based on which `type` had been selected. But the benefit of a consistent UI is probably better than being clever.

It also has a bug in it, in that `selectedIcmpType` is now a string, so this wouldn't ever actually have guarded the UI in the current state.

But I also agree with your other comment, about inline inputs in a MiniTable being an attractive solution to explore. So I'll remove the existing `&& selectedIcmpType !== 0` bit and then look more into the editable MiniTable cell.

charliepark: Actually, I think the editable MiniTable cell should wait until one or two other open loops get closed, and can/should exist as a separate push.

david-crespo: Yeah, I think so. That could be pretty complicated.

app/forms/firewall-rules-common.tsx
```diff
@@ -251,17 +258,269 @@ const availableItems = (
...
+        const parsedIcmpType =
+          typeof values.icmpType === 'string'
+            ? parseInt(values.icmpType, 10)
+            : values.icmpType
+
+        // Validation is now handled by the form field, but add safety check
+        invariant(
+          !isNaN(parsedIcmpType) && parsedIcmpType >= 0 && parsedIcmpType <= 255,
+          'ICMP type validation failed: value must be between 0 and 255'
+        )
```
david-crespo: In the unlikely event this gets tripped (assuming form field validation matches this logic exactly), I don't think we want to throw an error here, which is what `invariant` does. I think noop-ing with an early return is better.

app/forms/firewall-rules-common.tsx
```diff
@@ -251,17 +258,269 @@ const availableItems = (
...
+  if (isNaN(parsed) || parsed < 0 || parsed > 255) {
+    return `ICMP type must be a number between 0 and 255`
+  }
+  return undefined
+}
+
+const icmpCodeValidation = (value: string | undefined) => {
+  if (!value || value.trim() === '') return undefined // allow empty
+
+  const trimmedValue = normalizeDashes(value.trim())
```
david-crespo: I thought maybe using this as the `transform` means that when we get the value here, it is already normalized, in which case we don't need to do it here.

charliepark: ah good point

app/forms/firewall-rules-common.tsx
```diff
@@ -251,17 +258,269 @@ const availableItems = (
...
+                  control={protocolForm.control}
+                  description={
+                    <>
+                      Enter a code (0) or range (e.g. 1&ndash;3). Leave blank for all
+                      traffic of type {selectedIcmpType}.
+                    </>
+                  }
+                  placeholder=""
+                  validate={icmpCodeValidation}
+                  transform={normalizeDashes}
```
david-crespo, line 492: Interesting idea, works well. I think it feels ok because there isn't really a risk of copy-paste surprises like there was with names.

app/forms/firewall-rules-common.tsx
```diff
@@ -251,17 +258,269 @@ const availableItems = (
...
+
+    if (start < 0 || start > 255) {
+      return 'ICMP code range start must be between 0 and 255'
+    }
+    if (end < 0 || end > 255) {
+      return 'ICMP code range end must be between 0 and 255'
+    }
+    if (start > end) {
+      return 'ICMP code range start must be less than or equal to range end'
+    }
```
david-crespo, line 396: We should make sure the API doesn't barf on `3-3`

app/forms/firewall-rules-common.tsx
charliepark: Agree that it's slightly different than the copy/paste issue with names and dashes/spaces. I figured this would ease confusion, since we use en-dashes in the help text above the input and I could see some dash-savvy user trying it out that way.

app/forms/firewall-rules-common.tsx
charliepark: Chatted with Kyle on it. Learned a few things.

Internally, even if `3` were entered in as the code, it would get stored as `{ first: 3, last: 3 }`, so the range of `3-3` isn't problematic on its own.

But in looking into it, Kyle noticed that neither port filters nor icmp codes have an internal check preventing a `first` being greater than a `last`, so `3-2` wouldn't actually trip up the API. Obviously, this isn't a good state to end up in, so he's going to fix that in Omicron tomorrow. So our validations can stay as they are, and should match Omicron soon.

benjaminleonard:
Yeah, also realising its not worth to go too deep on this because it's all changing to mini-tables anyway. I'd say lets just squeeze in _a_ version that works and then move onto the mini-table input.

david-crespo:
This all looks pretty good to me. I found one UX issue that we might decide we can just live with: when you type in an invalid value in the ICMP type field and then blur rather than hitting enter or clicking the `Custom` entry, you end up with text in the field, but the actual value is still considered blank, so when you submit, it just goes through as if you hadn't typed, rather than showing a validation error.

I say maybe we can tolerate this because this is how the arbitrary values combobox works. We could have it clear input on blur if there's text typed that hasn't been committed to the underlying state (we wouldn't only do this when the underlying state is unset — I think we'd want it to work the same way if you've already selected a value and _then_ type in some nonsense), but that would only help the UX when the user blurs separately from the submit click. Most of the time I would expect the submit click itself to be the event that blurs, in which case there's no point clearing on blur because it's already going to disappear when the submit goes through and the protocol filter subform resets.

I guess the best UX I can picture is that if you submit the form when there is stuff typed in but not committed to the field state, we would treat the blur/submit as an attempt to set whatever is typed in as the state so that we can trigger a validation error and prevent the surprising thing happening. Switching to the inline-edit minitable avoids in this probably when adding the filter, but we have the same problem upon submission of the full form: you have something invalid typed in the field but it wasn't set as form state so it wasn't validated, and when you submit the form treats it like nothing was there. This is arguably even worse than the current situation because at least now, you might notice the entry in the filters table is missing a type.

https://github.com/user-attachments/assets/f72eb542-b086-4d3d-b7e3-9963d7e484fb

benjaminleonard:
> ### Edit 2
> If instead of clearing on blur we attempt to set the typed-in value as the actual value (analogous to pressing enter) we would immediately get the desired validation error, and this is perfectly fine for both deliberate-click-elsewhere blur and submit-click blur. So maybe that's the best thing.

Going to make an issue for more general listbox/combobox improvements but this seems like the best solution right now.

david-crespo:
I noticed that the other arbitrary values combobox on this page (the only one in the whole app, it turns out) does not have this problem because it uses `onInputChange` to update the form value on every character typed. That is better than `onBlur`. In e3a7710ce0441abf3ef89ce9c2b82359da749cfb I did the same for the ICMP type field and added a test (and confirmed it fails when I remove `onInputChange`, and a test also fails when I remove `onInputChange` from the other combobox). I think that is fine for now, though while investigating I decided our Combobox implementation is way too complicated internally, especially if we're doing this `onInputChange` thing every time we use `allowArbitraryValues`.

I'll have to take another quick look over, but most likely this PR is good to go.

david-crespo:
Welp, that test failure is not a flake.

PR #2853, opened 2025-07-11
bump oxlint to 1.6.0

david-crespo:
Not very interesting. I ran into the oxlint site and remembered they had gone 1.0.


app/components/MswBanner.tsx
```diff
@@ -59,7 +59,7 @@ export function MswBanner({ disableButton }: Props) {
         >
           Learn more <NextArrow12Icon />
         </button>
-      </label>
+      </aside>
```
david-crespo, line 62: `jsx-a11y` got mad that this label didn't wrap a control. it was correct, it shouldn't be a label

PR #2855, opened 2025-07-12
Bump Omicron and add new IP Pool column to networking table

### LINKED ISSUE #2852: Add IP pool to instance external IPs table
We had it for floating IPs, but not ephemeral. https://github.com/oxidecomputer/omicron/pull/8564 adds it for ephemeral, so should be able to add a column and fetch the pool by its ID. Could be async after the page loads like we do sometimes, or we could get it done in the loader, like we do other times.

Image

charliepark:
This is not tied to https://github.com/oxidecomputer/console/pull/2850, but as I worked on that PR, I wanted to be sure to pull in the "verify range values" changes from https://github.com/oxidecomputer/omicron/pull/8578 (which, in hindsight, likely didn't actually affect anything in console).

When I bumped Omicron to pick Kyle's change up, I also picked up the changes in https://github.com/oxidecomputer/omicron/pull/8564, which required some updates in console to resolve. To make sure that things were working right, I took care of https://github.com/oxidecomputer/console/issues/2852.

This will close https://github.com/oxidecomputer/console/issues/2852 when merged to main, though this PR is currently set to merge it back to the branch in https://github.com/oxidecomputer/console/pull/2850 to minimize the diff in the PR.

Screenshot 2025-07-11 at 5 00 37 PM

Closes #2852


mock-api/msw/handlers.ts
```diff
@@ -737,9 +741,12 @@ export const handlers = makeHandlers({
   instanceEphemeralIpAttach({ path, query: projectParams, body }) {
     const instance = lookup.instance({ ...path, ...projectParams })
     const { pool } = body
-    const firstAvailableAddress = getIpFromPool(pool)
+    const poolObj = resolveIpPool(pool, 'for ephemeral IP attachment')
+    const firstAvailableAddress = getIpFromPool(poolObj.name)
+
```
david-crespo, line 746: `getIpFromPool` calls `resolvePool` already, so you should be able to do this by passing `pool` directly to `getIpFromPool`. I don't think the `context` bit adds much anyway, so maybe you could drop it and inline `resolvePool` into `getIpFromPool`.

mock-api/msw/db.ts
```diff
@@ -55,10 +55,24 @@ function ensureNoParentSelectors(
...
+    )
+  }
+}
+
 export const getIpFromPool = (poolName: string | undefined | null) => {
-  const pool = lookup.ipPool({ pool: poolName || undefined })
+  const pool = resolveIpPool(poolName)
   const ipPoolRange = db.ipPoolRanges.find((range) => range.ip_pool_id === pool.id)
-  if (!ipPoolRange) throw notFoundErr(`IP range for pool '${poolName}'`)
+  if (!ipPoolRange) throw notFoundErr(`IP range for pool '${poolName || 'default'}'`)
```
david-crespo: `'default'` is not necessarily the name of the default pool so maybe

```suggestion
  if (!ipPoolRange) {
    const poolLabel = poolName ? `pool '${poolName}'`: 'default pool'
    throw notFoundErr(`IP range for ${poolLabel}`)
  }
```

mock-api/msw/db.ts
```diff
@@ -55,10 +55,24 @@ function ensureNoParentSelectors(
...
+    return lookup.siloDefaultIpPool({ silo: defaultSilo.id })
+  } catch (_error) {
+    const contextMsg = context ? ` ${context}` : ''
+    throw new Error(
+      `No IP pool specified${contextMsg} and no default IP pool configured for silo. Please specify a pool.`
+    )
+  }
+}
+
 export const getIpFromPool = (poolName: string | undefined | null) => {
```
david-crespo: I see this was already like this, but I'm realizing this is not always a name so it might be helpful to rename `poolName`. In the lookup functions we call it `id` even when it might be a name, which is also kind of shitty, so maybe just go maximalist: `poolNameOrId`.

mock-api/msw/db.ts
```diff
@@ -55,10 +55,24 @@ function ensureNoParentSelectors(
...
+  }
+  try {
+    return lookup.siloDefaultIpPool({ silo: defaultSilo.id })
+  } catch (_error) {
+    const contextMsg = context ? ` ${context}` : ''
+    throw new Error(
+      `No IP pool specified${contextMsg} and no default IP pool configured for silo. Please specify a pool.`
+    )
+  }
+}
```
david-crespo: Probably not worth catching and rethrowing just to tweak the error message. These are not real errors — the guiding principle is that we only need them to do just enough to let us assert things in the e2e test.

app/table/cells/IpPoolCell.tsx
```diff
@@ -5,14 +5,20 @@
...
-  const pool = useApiQuery('projectIpPoolView', { path: { pool: ipPoolId } }).data
-  if (!pool) return <EmptyCell />
+  const { data: result } = useApiQueryErrorsAllowed('projectIpPoolView', {
+    path: { pool: ipPoolId },
+  })
+  if (!result) return <SkeletonCell />
+  // this should essentially never happen, but it's probably better than blowing
+  // up the whole page if the pool is not found
+  if (result.type === 'error') return <EmptyCell />
+  const pool = result.data
```
david-crespo, line 21: I made a little tweak here to add a loading state and avoid blowing up the whole page on an (unlikely) error. I think in the other places we were using this, we are fetching the pool by ID in the loader, so we didn't notice there was no loading state. We could do that on the networking tab, but there are already so many requests in the loader, I lean toward keeping it async.

mock-api/msw/handlers.ts
charliepark: We'll need an actual pool object (instead of just the name) so we can get the pool's ID for the `ip_pool_id` on line 749. Agree that the `context` aid is minimal, though. Cleaning that up.

charliepark: I have a bit more cleanup on this that cuts down on double-`resolvePool` calls, though.

david-crespo: I missed that we need the pool object. Keep in mind perf doesn't matter at all here — for me it's 100% about clarity

david-crespo (approved):

david-crespo:
I was concerned about adding noise to the diff in #2850 while we're still reviewing that one, but it's actually very small so I'm going to merge this.

PR #2856, opened 2025-07-14
CI with smaller mac runner

david-crespo:
I have been informed that the large Mac runners are powered by gerbils that refuse to eat anything but hundred dollar bills.

https://docs.github.com/en/billing/managing-billing-for-your-products/about-billing-for-github-actions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,7 +51,7 @@
         "use-debounce": "^10.0.4",
         "uuid": "^10.0.0",
         "zod": "^3.24.3",
-        "zustand": "^5.0.3"
+        "zustand": "^5.0.6"
       },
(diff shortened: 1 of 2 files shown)


david-crespo:
Until @augustuswm can get me a bigger but not too big linux runner, we can at least switch back to the standard mac one, which costs about half as much, though it takes nearly twice as long, so it might be kind of a wash.

david-crespo:
Haha well. If the smaller runner has a bunch of flakes slowing down the tests, it's actually cheaper to use the bigger one (assuming I am reading the prices right and it's double the price).

### macos-15-xlarge

Total: 15 minutes

image


### macos-15

Total: 34 minutes

image

PR #2860, opened 2025-07-23
Audit log

david-crespo:
https://console-git-audit-log-oxidecomputer.vercel.app/system/audit-log

Copied from [#2849](<https://github.com/oxidecomputer/console/issues/2849>), which I accidentally merged and couldn't reopen even after I fixed `main`.

---

Still very rough, but has been a helpful exercise in working through some of the design due to the sheer amount of information and differing layout from other pages.

https://github.com/user-attachments/assets/84cbb9f6-d22b-4e64-9b1d-e39d74d22902

Stubbing out based on [https://github.com/oxidecomputer/omicron/pull/7339](<https://github.com/oxidecomputer/omicron/pull/7339>).

Uses [Tanstack Virtual](<https://tanstack.com/virtual/latest>). On testing with > 500 lines without virtualisation it starts to get a bit chunky especially if you're interacting with the page (e.g. opening the row).

Hoping that silo name and actor display name can be plumbed through so those are hard-coded for now.

Still needs:

- [X] Error state
- [X] Loading/placeholder state
- [X] Copy JSON to clipboard
- [ ] Equivalent CLI/API command
- [X] Fix giant footer spacing
- [X] Arrow key selected item navigation
- [X] Hide overflowing columns
- [X] Improved focus visible look
- [ ] Timestamp hover
- [X] Fix gradient on light mode
- [X] Syntax highlighting


david-crespo:
Can't repro locally, but this happened twice in CI so it's probably not a fluke.

image

app/pages/system/AuditLog.tsx
```diff
@@ -22,11 +22,25 @@ import { Badge } from '~/ui/lib/Badge'
...
 import { PageHeader, PageTitle } from '~/ui/lib/PageHeader'
 import { Spinner } from '~/ui/lib/Spinner'
+import { classed } from '~/util/classed'
 import { toSyslogDateString, toSyslogTimeString } from '~/util/date'
 import { docLinks } from '~/util/links'
 
 export const handle = { crumb: 'Audit Log' }
 
+// todo
+// might want to still render the items in case of error
```
benjaminleonard: Probably, we could just add a banner and a refresh. What type of error could we expect here? If you were no longer authorised you'd be logged out? Just a general server error retrieving?

app/pages/system/AuditLog.tsx
```diff
@@ -22,11 +22,25 @@ import { Badge } from '~/ui/lib/Badge'
...
+const ErrorState = () => {
+  return <div>Error State</div>
+}
+
+// todo
```
benjaminleonard: I think I'll do a skeleton state that is essentially a bunch of empty items to avoid too much layout shift when they load

app/pages/system/AuditLog.tsx
david-crespo: Errors should be pretty unlikely if you can get to the page at all, so I’d say keep it very generic.

benjaminleonard:
@david-crespo What's the path to getting this merged, other than the merge conflicts?

david-crespo:
Claude one-shotted the main merge. It looks pretty good. Still need to check it in more detail and re-review everything. I see some tearing or pop-in when I scroll the list fast, maybe due to virtualization. I don't remember how any of this works.

david-crespo:
This is looking quite good to me. Is it weird that the side thing is under the table headers? It almost makes it looks like they should apply to the modal contents.

image

Something else I notice is that some of the time when I click the load more button and I'm scrolled all the way down, the new data pops in and then my window jumps to the bottom to meet it. Sometimes it happens, other times not. I'll have to see if I can figure out what is going on there.

benjaminleonard:
Yeah I tweaked to move the selected item popover over the header row. We lose the `pt-3` since we want to match the height of the popover header and the header row. But I think it works better this way.

benjaminleonard:
I've also removed the auto scroll on click, it's just if you use the arrow keys or the up/down within the popover. Less jumping around for the user.

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+import { toLocaleDateString, toSyslogDateString, toSyslogTimeString } from '~/util/date'
+import { docLinks } from '~/util/links'
+import { deterRandom } from '~/util/math'
+
+export const handle = { crumb: 'Audit Log' }
+
+/**
+ * Convert API response JSON from the camel-cased version we get out of the TS
+ * client back into snake-case, which is what we get from the API. This is truly
+ * stupid but I can't think of a better way.
```
fakemonster: can we reuse the [snakeify function](https://github.com/oxidecomputer/console/blob/audit-log/app/api/__generated__/util.ts#L64) _ish_? seeing as it really necessitates both snakeify and the dateReplacer in http-client. i think logically we'd expect identical behavior from a "reserializer" here and the regular serializer

app/hooks/use-window-size.ts
```diff
@@ -0,0 +1,40 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+import { useEffect, useState } from 'react'
+
+export const useWindowSize = () => {
```
fakemonster: imo we don't need this yet; breakpoints (maybe container queries) should be enough. if we _really_ do, then: i'm not sure if we're trying to future-proof for future server-rendering, but if not, this doesn't need all the window-guarding. inversely, the way AuditLog is designed, that zero flash doesn't really matter, but that _is_ something i think is worth future-proofing against. you could get by with

```tsx
const getSize = () => ({ width: window.innerWidth, height: window.innerHeight })

export const useWindowSize = () => {
  const [size, setSize] = useState<{
    width: number
    height: number
  }>(getSize)

  useEffect(() => {
    const handleResize = () => setSize(getSize())
    window.addEventListener('resize', handleResize)
    return () => window.removeEventListener('resize', handleResize)
  }, [])

  return size
}
```

to solve both

app/ui/lib/DatePicker.tsx
```diff
@@ -55,7 +55,7 @@ export function DatePicker(props: DatePickerProps) {
           type="button"
           className={cn(
             state.isOpen && 'z-10 ring-2',
-            'text-sans-md border-default hover:border-raise bg-default relative flex h-11 items-center rounded-l-md rounded-r-md border focus-within:ring-2 focus:z-10',
+            'text-sans-md border-default hover:border-raise bg-default relative flex h-10 items-center rounded-l-md rounded-r-md border focus-within:ring-2 focus:z-10',
```
fakemonster, line 58: i'm starting to suspect there's a common visual component between ListBox and these pickers that could perhaps even do something a little more graceful with rounding. will think about this for a followup

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+  <span className="inline-block" style={{ width: `${depth * 2}ch` }} />
+)
+
+const greenText = 'text-(--color-green-1000) light:text-(--color-green-600)'
+const yellowText = 'text-(--color-yellow-1000) light:text-(--color-yellow-600)'
+
+const Primitive = ({ value }: { value: JsonValue | Date }) => {
+  if (value === null) return <span className={yellowText}>null</span>
+  if (typeof value === 'string') return <span className={greenText}>{`"${value}"`}</span>
+  if (value instanceof Date) return <span className={greenText}>{value.toISOString()}</span>
```
fakemonster: i don't know if we're actually trying to render json, but that's my impression, so let's wrap this in quotes (although i want to come back to this in the meta)

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+      {/* Generate skeleton rows */}
+      <div className="w-full">
+        {[...Array(50)].map((_, i) => (
+          <div
+            key={i}
+            className={cn(
+              'grid h-9 w-full animate-pulse items-center gap-8 px-[var(--content-gutter)] border-secondary',
+              i !== 0 && 'border-t'
+            )}
+            style={{ ...colWidths, animationDelay: `${i * 0.1}s` }}
```
fakemonster: i kind of like it this way, and it might be intentional, but just in case it's not: it's a common move to use animation delay thinking of it as an offset, but it really is just a delay. row 49's first-ever animation is going to be after 4.9 seconds. you can see this pretty well if you just always show the LoadingState and then zoom way out; the whole table is solid while a single "fade band" passes through, then another, and eventually you've got ~3 at a time. if you _do_ want the whole table animating from the jump, delays behave like an offset if they're negative: `i * 0.1 - 5` or `(i - arr.length) * 0.1`

app/util/math.ts
```diff
@@ -112,3 +112,10 @@ export function diskSizeNearest10(imageSizeGiB: number) {
   const nearest10 = Math.ceil(imageSizeGiB / 10) * 10
   return Math.min(nearest10, MAX_DISK_SIZE_GiB)
 }
+
+export function deterRandom(i: number, target: number, range: number) {
```
fakemonster: another slightly odd one. if you want _really_ random, this isn't doing it, if you like the style though (i do), i'd call it `noiseAround` or something similar, as you can see from the output each index is somewhat related to its neighbor

mock-api/audit-log.ts
```diff
@@ -0,0 +1,212 @@
...
+function generateAuditLogEntry(index: number): Json<AuditLogEntry> {
+  const operation = mockOperations[index % mockOperations.length]
+  const statusCode = mockHttpStatusCodes[index % mockHttpStatusCodes.length]
+  const isError = statusCode >= 400
+  const baseTime = new Date()
+  baseTime.setSeconds(baseTime.getSeconds() - index * 5 * 1) // Spread entries over time
+
+  const completedTime = new Date(baseTime)
+  completedTime.setMilliseconds(
+    Math.abs(Math.sin(index)) * 300 + completedTime.getMilliseconds()
```
fakemonster: maybe take Rando instead of whipping up a new generator?

app/util/links.ts
```diff
@@ -40,6 +40,10 @@ export const docLinks = {
     href: 'https://docs.oxide.computer/guides/deploying-workloads#_affinity_and_anti_affinity',
     linkText: 'Anti-Affinity Groups',
   },
+  auditLog: {
+    href: 'https://docs.oxide.computer/guides/audit-logs',
```
fakemonster: i believe you want https://docs.oxide.computer/guides/operator/audit-log

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+
+function StatusCodeCell({ code }: { code: number }) {
+  const color = code >= 200 && code < 500 ? 'default' : 'destructive'
+  return <Badge color={color}>{code}</Badge>
+}
+
+const COLUMNS = [
+  { key: 'timeCompleted', title: 'Time Completed', width: '7.75rem', hideBelow: 0 },
+  { key: 'status', title: 'Status', width: '3rem', hideBelow: 0 },
+  { key: 'operation', title: 'Operation', width: '160px', hideBelow: 0 },
```
fakemonster: operation names can get way longer (try out `anti_affinity_group_member_instance_add`). if you're trying to avoid letting it be dynamic (it could be, we just need some subgridding for alignment!), then this should have some kind of truncation/overflow control. right now it just waltzes into the next column

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+
+  return (
+    <div
+      className="absolute top-0 right-0 left-0 w-full focus-within:z-10"
+      style={{
+        height: `${size}px`,
+        transform: `translateY(${start - scrollMargin}px)`,
+      }}
+    >
+      <div
```
fakemonster, line 383: this really should be a button. i'd expect to hit space and not have my page scroll

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+    parentRef.current
+      ?.querySelector<HTMLElement>(`[data-row-index="${index}"]`)
+      ?.focus({ preventScroll: true })
+  }, [])
+
+  // arrow keys move selection (and focus) between rows; escape closes the
+  // modal. adjacent rows are within the virtualizer's overscan, so they're
+  // already in the DOM when we look them up.
+  useEffect(() => {
+    if (expandedItem === null) return
```
fakemonster: actually, i was kind of expecting arrow keys to work when there was no expanded item as well!

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+              <StatusCodeCell code={result.httpStatusCode} />
+            ))
+            .with({ kind: 'unknown' }, () => <EmptyCell />)
+            .exhaustive()}
+        </div>
+        <div>
+          <Badge color="neutral">{log.operationId.split('_').join(' ')}</Badge>
+        </div>
+        <div className={cn('text-secondary', hideActorId && 'hidden')}>
+          {userId ? (
```
fakemonster, line 420: user and silo ids feel like kind of a dead end. is there anything i can actually do with them?

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,818 @@
...
+      className={cn(
+        'absolute right-0 top-0 flex w-120 flex-col pt-px gap-6 overflow-y-auto z-10 border-l bg-default border-secondary',
+        hasError
+          ? 'mt-10 h-[calc(100dvh-var(--top-bar-height)-40px)]'
+          : 'h-[calc(100dvh-var(--top-bar-height))]'
+      )}
+    >
+      <div className="bg-raise border-secondary flex h-10 items-center justify-between border-b px-2">
+        <div className="flex items-center">
+          <button
```
fakemonster, line 776: it's a bit of a small nicety, but i like putting an `autoFocus` on the close button when there is one. webaim suggests this as a requirement for a11y, though that can get a little iffy when you're somewhere between a dialog and an "expanded view". my rule of thumb has generally been "if there's an X, just focus it". would be great to do the inverse too (refocus the source row on close)

fakemonster (commented):
couple visual issues that don't necessarily fall on a single line:
1. the non-datepicker popovers (docs info, refresh interval picker, range picker) fall behind the header because it's sticky. you can add a higher z-index to them and get away with it but it's pretty screwy. better to lower the sticky header to z-10, and then add an `isolate` class to the entire table to put it in a different stacking context (so it doesn't cover the header, which i assume is the original purpose of the z-20)
2. this isn't exactly using a side modal, but z-index-wise, it needs some sort of... sidemodality. right now tooltips render behind the detail view

app/pages/system/AuditLog.tsx
benjaminleonard: I'm lobbying to have these resolved in the API as silo and user names. Or an object that contains both.

app/hooks/use-window-size.ts
benjaminleonard: Pushing a version without this hook 👍

app/ui/lib/DatePicker.tsx
benjaminleonard: That'd be great, we're also repeating some of this in a few places that's more brittle than a shared class or component.

app/pages/system/AuditLog.tsx
benjaminleonard: TIL!

app/util/links.ts
fakemonster: closer, but you still need the `operator` segment

app/pages/system/AuditLog.tsx
david-crespo: Both the user and the silo in question could already be deleted at the time the log is being viewed, so resolving them to display name/silo name won't always work. The main thing someone would do with these IDs is use them to correlate actions done by the same actor. The thing that's tough here is that to me it would be pretty unusual for anyone to actually try to use this UI to do the kind of forensic work you actually want an audit log for. Anyone who relies on this should be dumping it to an external system regularly anyway, and they would likely be doing the forensics with scripts and queries running against that external system. In order for that not to be true, we would probably need more robust filtering -- right now it's just time range.

So the real use case guiding this page is intrinsically a little muddled. I feel like it's primarily useful as a sort of demo and reinforcement of the idea that the audit log is working the way you expect it to, logging the kinds of things you expect it to log.

fakemonster: when we say "deleted", how hard is that delete? silo aside, for the purposes of auditing I'd expect an auditor to expect to be able to resolve an actor id to _something,_ no matter what's happened to the actor itself. maybe a text filter on actor in the future is nice (I'm sort of imagining a user saying "I can't log in" and an operator going to see whether they ever successfully logged in)

david-crespo: I believe they are soft deletes, but there's no way to recover them other than querying the database directly, which operators can't do. It is a real issue for the usability of the audit log. See https://github.com/oxidecomputer/omicron/issues/8814 and https://github.com/oxidecomputer/omicron/issues/8810.

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,764 @@
...
+          </div>
+        ) : (
+          <Button
+            variant="ghost"
+            onClick={() => fetchNextPage()}
+            disabled={isFetchingNextPage}
+            type="button"
+            loading={isFetchingNextPage}
+          >
+            Load More
```
fakemonster: maybe fine, and i hate to throw the error state around even more, but if a query fails, this'll still say "load more" rather than something like "retry" which would be more accurate

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,764 @@
...
+  const snakeJson = useMemo(() => snakeify(item) as JsonValue, [item])
+  const json = useMemo(() => JSON.stringify(snakeJson, null, 2), [snakeJson])
+
+  return (
+    <div
+      className={cn(
+        'absolute right-0 top-0 flex w-120 flex-col pt-px gap-6 overflow-y-auto z-10 border-l bg-default border-secondary',
+        hasError
+          ? 'mt-10 h-[calc(100dvh-var(--top-bar-height)-40px)]'
+          : 'h-[calc(100dvh-var(--top-bar-height))]'
```
fakemonster: i don't see the merit of lowering this down past the column titles only when there's an error. it seems fine either consistently over the titles or consistently below it. especially given my upcoming error opinion!

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,764 @@
...
+                    currentIndex={currentIndex}
+                    totalCount={allItems.length}
+                    onNavigate={navigateToIndex}
+                    onClose={() => handleToggle(null)}
+                  />
+                )
+              })()}
+          </div>
+          {showError && (
+            <ErrorState error={errorMessage} onDismiss={() => setDismissedError(true)} />
```
fakemonster: if a "load more" click leads to an error, this'll still be up top (and definitely missed). can we show it in the position of the page that should've loaded? i can also take a crack at this if you like

app/pages/system/AuditLog.tsx
```diff
@@ -0,0 +1,764 @@
...
+  const [dismissedError, setDismissedError] = useState(false)
+
+  // pass refetch interval to this to keep the date up to date
+  const { preset, startTime, endTime, dateTimeRangePicker, onRangeChange } =
+    useDateTimeRangePicker({
+      initialPreset: 'lastHour',
+      maxValue: now(getLocalTimeZone()),
+    })
+
+  const { intervalPicker } = useIntervalPicker({
```
fakemonster: having returned to this view... do we actually want an interval picker? right now if i load more pages and then the interval fires, pages 2 and on are just dropped, which seems all too surprising. we could fix that in one of two ways:

1. when a refresh fires, only prepend those new incoming logs to the existing set. this would involve searching for the intersection with the existing list, dropping the redundant ones, and correcting the selected index if there is one. kind of fragile. plus, if the new page _doesn't_ intersect with the logs we already have, we'll need to automatically make further requests to fill the gap
2. let the refresh wipe state, but be "however many pages you had rendered before". that's a little odd cause some logs are still potentially disappearing. and since we're giving an infinite query here, we're effectively saying "go ahead and load a few dozen pages, and we'll refresh _all_ of them on an interval". certainly there's a limit somewhere?

if we had only only explicit reload, it would make dropping pages 2+ much more expected, imo. or we could just let good old full-page refreshing/time period changes be the refresh pattern here

fakemonster (commented):
a few more observations. just like in social media, infinite scroll leads to pain!

app/pages/system/AuditLog.tsx
fakemonster: went ahead and moved the error message below the list (since if the first call fails, that's effectively the same placement as today) in 8eef97e

app/pages/system/AuditLog.tsx
fakemonster: moving the error message to the bottom in [8eef97e](https://github.com/oxidecomputer/console/pull/2860/commits/8eef97e6ed058e659d1c73be690606ef6578ca80) obviates this

fakemonster:
i think these are the issues we would want to create for potential followup:

1. [classier control buttons](https://github.com/oxidecomputer/console/pull/2860#discussion_r3509122941)
2. get rid of the handrolled json viewer (we need one for the alerts page too, and the oxql stuff seems to be tipping towards adding shiki for oxql, so why not handle json that way)
3. show silo name when applicable? i know we've discussed actor name as well, but it may be inappropriate here anyways
4. "timestamp hover" (this is just in the checklist above, idk what it is and we could probably strike it)

david-crespo:
There's a statefulness implied by the listbox that I don't love. There's no difference in practice between having `Last 3 hours` and `Custom` selected if the same time range is in the actual picker — there's no sliding window since we killed polling. So I wonder if we should make the thing more menu-like instead of making it a listbox. This is similar to the query examples in the OxQL explorer, which we turned into buttons: https://github.com/oxidecomputer/console/pull/3312#discussion_r3815582249.

image

david-crespo:
Low priority because it only happens on the preview, but the sticky header goes under the top bar when I scroll.

https://github.com/user-attachments/assets/2254ffde-ce76-4d76-aca2-59ec735438f5

PR #2861, opened 2025-07-24
Combobox optimizations

charliepark:
A little more cleanup to go, but this speeds up Comboboxes on Chrome significantly.


app/components/form/fields/ComboboxField.tsx
```diff
@@ -69,25 +59,62 @@ export function ComboboxField<
     control,
     rules: { required, validate },
   })
-  const [selectedItemLabel, setSelectedItemLabel] = useState(
-    getSelectedLabelFromValue(items, field.value || '')
+
+  // Memoize description computation to avoid recalculating on every render
+  const memoizedDescription = useMemo(() => {
```
benjaminleonard, line 64: I'd be surprised if this memo has a meaningful effect on performance

app/forms/disk-create.tsx
```diff
@@ -85,32 +85,66 @@ export function CreateDiskSideModalForm({
...
+      },
+      blockSize: 512,
+      timeCreated: new Date(),
+      timeModified: new Date(),
+      os: 'linux',
+    }))
+  }, [])
+
+  // Memoize combined images array
+  const images = useMemo(() => [...realImages, ...mockImages], [realImages, mockImages])
```
benjaminleonard, line 118: Again suspect an improvement is negligible here, suspect spreading two arrays is relatively inexpensive

benjaminleonard:
I think we might get better results by migrating to something `react-aria`. Here is their listbox / combobox / virtualizer with 10,000 mock images. We'd just need to make sure that the bug we ran into [previously](https://github.com/oxidecomputer/console/pull/1563) is not still present.

CleanShot 2025-07-24 at 11 00 16

app/components/form/fields/ComboboxField.tsx
david-crespo: It could even be _more_ computation than not memoizing — the `useMemo` call and its dep checks are work too. The point of this would be to prevent the value changing to avoid triggering renders on `<Combobox>` below, but this is just a string and it wouldn't be changing its value between renders, so it is almost certainly triggering re-renders.

benjaminleonard: Why would this trigger a re-render – since `description`, `placeholder` and `allowArbitraryValues` are likely the same values across the lifetime of this component. So the prop on `<Combobox>` wouldn't change.

david-crespo:
Yeah, having to manually memoize so much in the call sites suggest to me something bigger is wrong. I agree with Ben that it's worth considering changing the library out altogether if we're having so much trouble with Headless. At the very least I'd like to have a clearer sense of which optimizations are responsible for the majority of the improvement here. I would guess is a pretty small subset of them.

Scroll and filter performance is definitely better, though the debounce feels kind of bad because it means filtering still feels a little slow. Maybe a shorter window on it would be better.

Saw some fun bugs in Firefox. The double scrollbar is obvious. A much weirder one is that at 10s when it jumps around and starts scrolling up, I am not doing the scrolling myself. All I did was filter for 3, scroll down a while, then hit backspace to clear the `3` from the input. At that point it scrolls to the top gradually all by itself.

https://github.com/user-attachments/assets/a8ae7ce8-994a-4a70-8464-7813ad010271

app/components/form/fields/ComboboxField.tsx
david-crespo: Yeah, that's what I meant about it being a string. If it was constructing an object or array with a spread like the one in disk create, then it would be a new one every time (even if the underlying values are the same) and then it would trigger renders when passed down the chain. So the other ones are more plausible, though in those cases it's still worth determining whether they have much effect. Triggering renders is not always a problem.

charliepark: Yeah, the memoization explosion was from before I added in the virtualization; was going to try backing them out to see whether they were having a noticeable effect, but will read the other comments on this first

charliepark:
Going to approach this with a few more targeted branches.

PR #2868, opened 2025-07-30
Add snat to external ips

### LINKED ISSUE #2866: Display SNAT IPs on instance external IPs list
See https://github.com/oxidecomputer/omicron/pull/8685

charliepark:
This adds SNAT IPs to the list of an instance's external IPs. This is the console counterpart to https://github.com/oxidecomputer/omicron/pull/8685. We're still talking through a few design decisions around ports, but I wanted to get a PR going for it.

Screenshot 2025-07-30 at 3 00 46 PM

Closes #2866


david-crespo:
@bnaecker suggests we should filter out the SNAT IP here. I think that's a good idea.

image

https://github.com/oxidecomputer/console/blob/ef73c18e89a7b1f2bbe488a60fdc30b387e917fd/app/components/ExternalIps.tsx#L25-L29

mock-api/msw/handlers.ts
```diff
@@ -885,11 +889,12 @@ export const handlers = makeHandlers({
       (r) => parseIp(r.first).type === 'v4'
     )
 
-    // in the real backend there are also SNAT IPs, but we don't currently
-    // represent those because they are not exposed through the API (except
-    // through the counts)
+    // Include SNAT IPs in IP utilization calculation, deduplicating by IP address
+    // since multiple instances can share the same SNAT IP with different port ranges
+    const uniqueSnatIps = [...new Set(db.snatIps.map((sip) => sip.external_ip.ip))]
```
charliepark, line 877: Just wanted to call out that this is where we handle the SNAT IP address de-duplication (de-quadification?). Essentially, only pull unique IPs, and use that for the utilization count.

app/pages/project/instances/NetworkingTab.tsx
```diff
@@ -363,6 +385,11 @@ export default function NetworkingTab() {
         },
       }
 
+      if (externalIp.kind === 'snat') {
+        // SNAT IPs can't be detached
+        return [copyAction]
+      }
+
```
david-crespo, line 400: This is good, but it might be slightly nicer to leave the button in and disable it with "SNAT IPs can't be detached" as a hover message. Taking the opportunity to make clear this is on purpose.

david-crespo: Did this in 9da73602532fa3f45a528abc8c8c3d785d3386b7

app/pages/project/instances/NetworkingTab.tsx
```diff
@@ -364,18 +386,38 @@ export default function NetworkingTab() {
...
+      const [label, doDetach] = match(externalIp)
+        .with(
+          { kind: 'ephemeral' },
+          () =>
+            [
+              // oxlint-disable-next-line jsx-key
+              <>this ephemeral IP</>,
+              () =>
+                ephemeralIpDetach({ path: { instance: instanceName }, query: { project } }),
+            ] as const
```
david-crespo: oxlint had to ruin my beautiful tuple match thing with a false positive for the key rule because it's in an array. I should probably report this as an oxlint bug.

david-crespo: I took it out, it was too extra

david-crespo:
I think this is basically ready to go. My only issue is that in the SNAT IP-only case (which may be rare), it's not explained in the UI why there is nothing in the external IPs at the top, but there is this SNAT IP in the table lower down. The tooltip on the kind column _could_ explain this but it's already getting too wordy. I'm tempted to put a tooltip next to every badge that explains only that kind, but it would look silly. Another option would be to change "external IPs" in the properties table to "inbound IPs" or something — not sure whether that is clever or terrible. Yet another option would be to stick an info tooltip next to the empty cell thing to explain, but _only_ in the case where there is a SNAT IP and no others.

image

image

david-crespo:
`TipIcon` in properties table, making the empty cell itself the tooltip (not pictured, but I tried it), and changing "external IPs" to "inbound IPs" all feel bad to me.

image


image

PR #2870, opened 2025-07-31
Add IP pool name to Attach Floating IP dropdown

### LINKED ISSUE #2869: Add IP pool info to dropdown when selecting a floating IP to attach to an instance
Right now, the floating IP selector omits any info about the IP Pool the available floating IPs are coming from. It could be useful to have that info available while the user is selecting one to attach.

Currently:
Image

charliepark:
This adds context to the selector dropdown users interact with when attaching a floating IP to an instance. Specifically, it adds the floating IP's associated IP pool to the information available in the dropdown.

Screenshot 2025-07-31 at 11 24 42 AM

Closes #2869

### DIFF
--- a/app/api/client.ts
+++ b/app/api/client.ts
@@ -16,10 +16,10 @@ import { type ApiError } from './errors'
 import {
   ensurePrefetched,
   getApiQueryOptions,
+  getApiQueryOptionsErrorsAllowed,
   getListQueryOptionsFn,
   getUseApiMutation,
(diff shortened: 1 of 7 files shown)


david-crespo:
I think this is handy. My only concern is that it may not be clear that it's a pool name when the name doesn't have the word "pool" in it. But I'm not sure how much we can do about that without bloating the line with "pool:" or something. @benjaminleonard?

image

image

Issue #2857, opened 2025-07-15
Combobox rework


benjaminleonard:
Currently a combobox and listbox both look are behave mostly the same but the combobox lets you add arbitrary values. But this treatment makes it unclear whether or not the input does accept arbitrary values and we run into issues by using the text input as the method by which you filter the list since you'd expect whatever is in the input to be the value accepted by the form.

I therefore think we can distinguish the two by having:

1. Combobox that behaves basically as it does not but that commits the value on blur (rather than requiring click or enter to assign the custom value).
2. A select that has a separate input to filter the list _within the dropdown_ instead. This way it's clear that it will not be saved when the dropdown is closed, that its distinct from the actual input and its more intuitive which accepts custom input.

Ariakit has [a good example](https://ariakit.org/examples/select-combobox) of the dropdown input for a new filterable listbox.

<img width="272" height="397" alt="Image" src="https://github.com/user-attachments/assets/6f11651d-adcd-4df0-9791-74869ecdefab" />

david-crespo:
Performance with a large number of items is also not very good. With 1000 items it's usable in Firefox but barely usable in Chrome on an M4 Pro. With CPU throttling in Chrome to simulate a slower machine it's totally unusable. Oddly, a recent HeadlessUI update meant to improve performance made it worse:

### Headless 2.2.0

https://github.com/user-attachments/assets/d717ecd8-9ab3-42be-90e6-e811c1322e64

### Headless 2.2.4

https://github.com/user-attachments/assets/2c35da47-53ac-47e0-94cc-df8ea6e8434a

PR #2877, opened 2025-08-11
Update arbitrary combobox value automatically without requiring confirmation

benjaminleonard:
Currently you are required to hit enter or click on the custom option to apply the arbitrary value with a combobox. This is confusing and the better behaviour is just to assume the user wants the custom value regardless.

Just a patch for now, the bigger rework as per #2857 will help make this even clearer. Only affects firewall rule forms for now, since that's the only place that allows arbitrary combobox values.


benjaminleonard:
Oh you're right. I was sure I was seeing the values being cleared on the arbitrary combobox, but I must have been mistaken or run into a bug.

PR #2878, opened 2025-08-11
Input mini table

### LINKED ISSUE #2627: Make it harder to forget to submit targets/filters subform

benjaminleonard:
Still very rough, don't look too closely, lest your eyes be burned.

CleanShot 2025-08-08 at 16 54 09

Closes oxidecomputer/console#2627


benjaminleonard:
image

From Vercel. "Add another" might help here? The implication being the existing one is already added. But that doesn't support the empty state. At which point we could move the CTA inside the empty state when it is empty, or conditionally change the copy. Both options are potentially helping clarify one area, whilst complicating elsewhere.

benjaminleonard:
Also, the empty state to single row transition I struggle with a little too. Mainly because of the height change.

The pattern I've seen more of is just ignoring empty rows on submission. But I did run into some implementation challenges with that and `react-hook-form`

david-crespo:
Looks like we just never nailed down the interaction here.

PR #2882, opened 2025-08-15
Zod v4 again

david-crespo:
Doing it properly this time.

https://github.com/oxidecomputer/oxide.ts/pull/300

### DIFF
--- a/mock-api/msw/handlers.ts
+++ b/mock-api/msw/handlers.ts
@@ -808,10 +808,12 @@ export const handlers = makeHandlers({
     }
     updateDesc(nic, body)
 
-    if (typeof body.primary === 'boolean' && body.primary !== nic.primary) {
-      if (nic.primary) {
-        throw 'Cannot remove the primary interface'
-      }
+    // We used to error here if body.primary was false and nic.primary was true
+    // on the grounds that you can't unset the primary interface. But this turns
+    // out not to match the real API, which ignores primary: false.
+    // https://github.com/oxidecomputer/omicron/blob/61ad056c/nexus/db-queries/src/db/datastore/network_interface.rs?plain=1#L804-L808
+
+    if (typeof body.primary === 'boolean' && body.primary && !nic.primary) {
       db.networkInterfaces
         .filter((n) => n.instance_id === nic.instance_id)
         .forEach((n) => {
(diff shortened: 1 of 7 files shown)


mock-api/msw/handlers.ts
```diff
@@ -808,10 +808,12 @@ export const handlers = makeHandlers({
...
-    if (typeof body.primary === 'boolean' && body.primary !== nic.primary) {
-      if (nic.primary) {
-        throw 'Cannot remove the primary interface'
-      }
+    // We used to error here if body.primary was false and nic.primary was true
+    // on the grounds that you can't unset the primary interface. But this turns
+    // out not to match the real API, which ignores primary: false.
+    // https://github.com/oxidecomputer/omicron/blob/61ad056c/nexus/db-queries/src/db/datastore/network_interface.rs?plain=1#L804-L808
+
+    if (typeof body.primary === 'boolean' && body.primary && !nic.primary) {
```
david-crespo, line 816: These tests were failing:

```sh
e2ec instance-networking -g 'NIC table' && e2ec instance-networking -g 'Transit IPs'
```

https://zod.dev/v4/changelog?id=defaults-applied-within-optional-fields

The above change means that `primary: SafeBoolean.default(false)` is putting `primary: false` on the `body`, where in v3 it was just left out. That was tripping the "I refuse to unset the primary" logic below. The default comes from [here](https://github.com/oxidecomputer/omicron/blob/61ad056cee3d8cfa509799ce2d4f898bcafb54e4/nexus/types/src/external_api/params.rs?plain=1#L950) in omicron and has been there for three years.

mock-api/snapshot.ts
```diff
@@ -77,7 +77,7 @@ export const snapshots: Json<Snapshot>[] = [
     state: 'ready',
   },
   {
-    id: '7fc6ca11-452e-d3e4-9e1c-752ff615abea',
+    id: '75832119-7d9b-4e91-a7c6-76b996b69aaa',
```
david-crespo, line 80: Turns out this was an invalid UUID, correctly rejected by Zod because of the d in d3e4, which is supposed to be a 4 for UUID v4. Solved this on the last attempt, see https://github.com/oxidecomputer/console/pull/2828#discussion_r2096366571.

PR #2893, opened 2025-08-25
Prevent event bubbling on copy-to-clipboard clicks

charliepark:
On #2860, I noticed that clicking the copy buttons was opening the JSON viewer for each row, because the CopyToClipboard component isn't capturing the event bubbling. This fixes that.


david-crespo (approved):
Looks good. Might not need the `e.preventDefault()`: since it's not a submit button, I don't think there is a default behavior.

PR #2896, opened 2025-09-04
metrics: handle varying ncpu within a query.

jmcarp:
The metrics dashboards assume that queries return a consistent number of series
over time. For example, if a query for cpu utilization gets four timeseries at
one timepoint, it shouldhave four timeseries at all timepoints. However, this
assumption isn't met when the cardinality of the given resource changes over
time, such as when the user scales the number of vcpus for an instance. This
assumption leads to two bugs:

* If the number of vcpus changes within a query, we normalize by the largest
  observed vcpu count. In other words, increasing the number vcpus makes
  utilization appear lower at timepoints prior to the scaling event.
* If a user scales up the number of vcpus such that the 0th vcpu timeseries has
  fewer timepoints than subsequent timeseries, we truncate series after the
  0th.

This patch attempts to fix both bugs. First, we count the number of non-null
values per timeseries to normalize cpu use. Second, we consider all timestamps
from all returned series so that we don't accidentally truncate the results.

### DIFF
--- a/app/components/oxql-metrics/util.ts
+++ b/app/components/oxql-metrics/util.ts
@@ -122,21 +122,40 @@ export const sumValues = (timeseries: Timeseries[], arrLen: number): (number | n
     )
   )
 
+export const countValues = (timeseries: Timeseries[], arrLen: number): number[] =>
+  Array.from({ length: arrLen }).map((_, i) =>
+    R.pipe(
+      timeseries,
+      // get point at that index for each timeseries
+      R.map((ts) => ts.points.values.at(0)?.values.values?.[i]),
+      // filter out nulls (undefined shouldn't happen)
+      R.filter((p) => typeof p === 'number'),
+      // count non-null elements per series
+      (points) => points.length
+    )
+  )
+
 // Take the OxQL Query Result and return the data in a format that the chart can use
 // We'll do this by creating two arrays: one for the timestamps and one for the values
 // We'll then combine these into an array of objects, each with a timestamp and a value
 // Note that this data will need to be processed further, based on the kind of chart we're creating
 export const composeOxqlData = (data: OxqlQueryResult | undefined) => {
   let timeseriesCount = 0
-  if (!data) return { chartData: [], timeseriesCount }
+  if (!data) return { chartData: [], valueCounts: [] }
   const timeseriesData = data.tables[0].timeseries
   timeseriesCount = timeseriesData.length
-  if (!timeseriesCount) return { chartData: [], timeseriesCount }
-  // Extract timestamps (all series should have the same timestamps)
-  const timestamps =
-    timeseriesData[0]?.points.timestamps.map((t) => new Date(t).getTime()) || []
-  // Sum up the values across all time series
+  if (!timeseriesCount) return { chartData: [], valueCounts: [] }
+  // Extract timestamps. Series may have different timestamps, e.g. when new
+  // resources are created within the query interval
+  const timestamps = R.pipe(
+    timeseriesData,
+    R.map((series) => series.points.timestamps.map((t) => new Date(t).getTime())),
+    R.flat(),
+    R.unique(),
+    R.sort((a, b) => a - b)
+  )
   const summedValues = sumValues(timeseriesData, timestamps.length)
+  const valueCounts = countValues(timeseriesData, timestamps.length)
   const chartData = timestamps
     .map((timestamp, idx) => ({ timestamp, value: summedValues[idx] }))
     // Drop the first datapoint, which — for delta metric types — is the cumulative sum of all previous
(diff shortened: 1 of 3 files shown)


app/components/oxql-metrics/util.ts
```diff
@@ -122,29 +122,48 @@ export const sumValues = (timeseries: Timeseries[], arrLen: number): (number | n
...
-  const timestamps =
-    timeseriesData[0]?.points.timestamps.map((t) => new Date(t).getTime()) || []
-  // Sum up the values across all time series
+  if (!timeseriesCount) return { chartData: [], valueCounts: [] }
+  // Extract timestamps. Series may have different timestamps, e.g. when new
+  // resources are created within the query interval
+  const timestamps = R.pipe(
+    timeseriesData,
+    R.map((series) => series.points.timestamps.map((t) => new Date(t).getTime())),
+    R.flat(),
```
david-crespo, line 153: Not that perf matters a ton here, but `R.flatMap` "is identical to a [map](https://remedajs.com/docs/#map) followed by a [flat](https://remedajs.com/docs/#flat) of depth 1 (`flat(map(data, ...args))`), but slightly more efficient than calling those two methods separately"

https://remedajs.com/docs/#flatMap

david-crespo:
Thank you for doing this. This is a gnarly bug and the logic is gnarly so I will have to spend a little time looking through it.

app/components/oxql-metrics/util.ts
```diff
@@ -122,29 +122,48 @@ export const sumValues = (timeseries: Timeseries[], arrLen: number): (number | n
...
+  const valueCounts = countValues(timeseriesData, timestamps.length)
   const chartData = timestamps
     .map((timestamp, idx) => ({ timestamp, value: summedValues[idx] }))
     // Drop the first datapoint, which — for delta metric types — is the cumulative sum of all previous
     // datapoints (like CPU utilization). We've accounted for this by adjusting the start time earlier;
     // We could use a more elegant approach to this down the road
     .slice(1)
 
-  return { chartData, timeseriesCount }
+  return { chartData, valueCounts }
```
david-crespo, line 166: Because the elements of `valueCounts` should match up one to one with the elements of `chartData`, it might be clearer to represent that relationship more directly by either making the count a field on the `chartData` element or even pre-normalizing the summed values data here by dividing by the count, so you don't have to return the count at all. I haven't worked through it enough to articulate it properly, just musing.

app/components/oxql-metrics/util.ts
```diff
@@ -122,29 +122,48 @@ export const sumValues = (timeseries: Timeseries[], arrLen: number): (number | n
...
+    R.sort((a, b) => a - b)
+  )
   const summedValues = sumValues(timeseriesData, timestamps.length)
+  const valueCounts = countValues(timeseriesData, timestamps.length)
   const chartData = timestamps
     .map((timestamp, idx) => ({ timestamp, value: summedValues[idx] }))
     // Drop the first datapoint, which — for delta metric types — is the cumulative sum of all previous
     // datapoints (like CPU utilization). We've accounted for this by adjusting the start time earlier;
     // We could use a more elegant approach to this down the road
     .slice(1)
```
david-crespo, line 164: does `valueCounts` also need to get its first value dropped?

app/components/oxql-metrics/util.ts
```diff
@@ -122,29 +122,48 @@ export const sumValues = (timeseries: Timeseries[], arrLen: number): (number | n
...
+    timeseriesData,
+    R.map((series) => series.points.timestamps.map((t) => new Date(t).getTime())),
+    R.flat(),
+    R.unique(),
+    R.sort((a, b) => a - b)
+  )
   const summedValues = sumValues(timeseriesData, timestamps.length)
+  const valueCounts = countValues(timeseriesData, timestamps.length)
   const chartData = timestamps
     .map((timestamp, idx) => ({ timestamp, value: summedValues[idx] }))
```
david-crespo, line 160: if you add `valueCount` here, maybe instead of using `idx`, `summedValues` and `valuesCounts` could be objects indexed by timestamp so we don't have to worry about the arrays lining up index-wise

app/components/oxql-metrics/util.ts
david-crespo: Oh's it's worse than that. This needs to be done per-timeseries prior to aggregation now that we realize they can have different first timestamps.

jmcarp:
Sorry, I forgot about this for a while. I can finish this up sometime, but I remember @JustinAzoff mentioning that a better solution would be to implement a proper `rate()` function in oxql, and then use that here. Do we think we'll have that function for R18, or should I finish this PR as an interim fix?

PR #2900, opened 2025-09-18
Instance CPU types

david-crespo:
See https://github.com/oxidecomputer/omicron/pull/8728. The API change is small and the default value means we don't actually have to change anything in the console yet except avoiding unsetting it in existing instance update calls. When we do add it to the UI, for create it would probably be down at the bottom of the form:

image

We'd also have to display the current value on instance detail and give a way to change it.


app/pages/project/instances/InstancePage.tsx
```diff
@@ -341,7 +341,14 @@ export function ResizeInstanceModal({
...
       query: { project },
-      body: { ncpus, memory: memory * GiB, bootDisk: instance.bootDiskId },
+      body: {
+        ncpus,
+        memory: memory * GiB,
+        // these need to be passed in or they'll be unset
+        bootDisk: instance.bootDiskId,
+        cpuPlatform: instance.cpuPlatform,
+        autoRestartPolicy: instance.autoRestartPolicy,
+      },
```
david-crespo, line 350: Pretty sure the fact that this was missing `autoRestartPolicy` was a bug.

david-crespo:
@iximeow do we want full UI support for the next release or can it wait until v18? Not that it's a big lift, really.

iximeow:
my 2c is that UI support isn't really helpful until someone's got Cosmos to work with, so I wouldn't sweat v17 for this

david-crespo:
I might make a helper with a stricte type than the API that can help us avoid forgetting a key. Another thing we might consider is removing optional from those keys on the API side while keeping them nullable, if that’s possible.

david-crespo:
> I might make a helper with a stricter type than the API that can help us avoid forgetting a key. Another thing we might consider is removing optional from those keys on the API side while keeping them nullable, if that’s possible.

I created this in 60ca75e6827ae810dba1d53f87a70cf248f4f942 and then removed it in 60ca75e6827ae810dba1d53f87a70cf248f4f942 because I did it on the API side.

app/api/__generated__/Api.ts
```diff
@@ -2514,11 +2538,13 @@ export type InstanceUpdate = {
 This policy determines whether the instance should be automatically restarted by the control plane on failure. If this is `null`, any explicitly configured auto-restart policy will be unset, and the control plane will select the default policy when determining whether the instance can be automatically restarted.
 
 Currently, the global default auto-restart policy is "best-effort", so instances with `null` auto-restart policies will be automatically restarted. However, in the future, the default policy may be configurable through other mechanisms, such as on a per-project basis. In that case, any configured default policy will be used if this is `null`. */
-  autoRestartPolicy?: InstanceAutoRestartPolicy | null
+  autoRestartPolicy: InstanceAutoRestartPolicy | null
```
david-crespo, line 2541: no longer optional

PR #2912, opened 2025-10-10
convert silo page tabs to route-based navigation

charliepark:
In preparation for adding the SCIM tab to the silo page, I thought it'd be good to update the silo page's tab routing. 
`/system/silos/maze-war?tab=ip-pools` → `/system/silos/maze-war/ip-pools`
Also adds a redirect to handle stale links.

### DIFF
--- a/app/pages/system/silos/SiloQuotasTab.tsx
+++ b/app/pages/system/silos/SiloQuotasTab.tsx
@@ -117,7 +117,7 @@ function EditQuotasForm({ onDismiss }: { onDismiss: () => void }) {
 
   const updateQuotas = useApiMutation('siloQuotasUpdate', {
     onSuccess() {
-      queryClient.invalidateQueries({ queryKey: ['siloUtilizationView'] })
+      queryClient.invalidateEndpoint('siloUtilizationView')
       addToast({ content: 'Quotas updated' })
       onDismiss()
     },


app/pages/system/silos/SiloPage.tsx
```diff
@@ -5,35 +5,43 @@
...
+export async function clientLoader({ params, request }: LoaderFunctionArgs) {
   const { silo } = getSiloSelector(params)
-  await Promise.all([
-    apiQueryClient.prefetchQuery('siloView', { path: { silo } }),
-    apiQueryClient.prefetchQuery('siloUtilizationView', { path: { silo } }),
-    queryClient.prefetchQuery(siloIdpList(silo).optionsFn()),
-    queryClient.prefetchQuery(siloIpPoolsQuery(silo).optionsFn()),
-  ])
+
+  // Handle old query param-based URLs for backwards compatibility
```
benjaminleonard: Thoughts on sweeping the docs for any stale links and removing this in the pursuit of simplicity?

david-crespo: I had already written 5fbc8fb5affee4abe87586b56edd2a936481dd98 but not pushed it yet

david-crespo: good news:
```
~/oxide/docs
$ rg 'tab='
app/components/Search.tsx
311:            tab={tab}
474:                tab={tab}
```

charliepark: ~~I'm not against it, but wasn't sure if there would be, say, customer in-house docs that would link to the old URLs, or other legacy links. Maybe the odds of that are low enough that we can remove this? But scanning the docs for old links is a good idea regardless.~~ Well, that will teach me to reply before reading the rest of the commits.

PR #2915, opened 2025-10-14
System Update

### LINKED ISSUE #2903: Update: status and repo list page
[Design](https://github.com/oxidecomputer/console/issues/2686)

Built on https://github.com/oxidecomputer/omicron/pull/8897 and https://github.com/oxidecomputer/omicron/pull/9106

Image

### LINKED ISSUE #2905: Update: set target release
[Design](https://github.com/oxidecomputer/console/issues/2686). I don't think it'll be a number picker like in the mockup below, because at any given time there should only be one release you can upgrade to. But basically you'll click a button on a release that lets you set that release as target and there will a confirmation modal.

Image

david-crespo:
Closes #2903 
Closes #2905 

We have read-only stuff and setting target release. I had trouble with upload, so it is in the first commit but it is removed in a subsequent commit.

### Outstanding questions

- [x] How to display "suspended" (Just saying True/False is kinda ugly)
  - Went with Yes/No
- [ ] Should we disable the set target release button in the UI in more conditions?
  - [ ] API disallows set target if an update is in progress (https://github.com/oxidecomputer/omicron/pull/9188) or if new target is an older semver than current target
  - [ ] Should we add a field in the API response telling us whether we can set that release as target right now? This would avoid duplicating API logic in the console and potentially (probably) getting it wrong.

image

### DIFF
--- a/app/pages/system/UpdatePage.tsx
+++ b/app/pages/system/UpdatePage.tsx
@@ -15,6 +15,7 @@ import {
   SoftwareUpdate24Icon,
   Time16Icon,
 } from '@oxide/design-system/icons/react'
+import { Badge } from '@oxide/design-system/ui'
 
 import {
   apiq,
@@ -31,7 +32,6 @@ import { makeCrumb } from '~/hooks/use-crumbs'
 import { confirmAction } from '~/stores/confirm-action'
 import { addToast } from '~/stores/toast'
 import { EmptyCell } from '~/table/cells/EmptyCell'
-import { Badge } from '~/ui/lib/Badge'
 import { CardBlock } from '~/ui/lib/CardBlock'
 import { DateTime } from '~/ui/lib/DateTime'
 import * as DropdownMenu from '~/ui/lib/DropdownMenu'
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -25,7 +25,7 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es2022",
+    "target": "es2023",
```
david-crespo, line 28: Needed to do this in order to get `roundingMode` on the type of the args to `Intl.NumberFormat`. `es2023` is currently the default in TS anyway, so we should be on it anyway.

https://www.typescriptlang.org/tsconfig/#target

app/util/links.ts
```diff
@@ -57,6 +57,7 @@ export const links = {
   systemIpPoolsDocs: 'https://docs.oxide.computer/guides/operator/ip-pool-management',
   systemMetricsDocs: 'https://docs.oxide.computer/guides/operator/system-metrics',
   systemSiloDocs: 'https://docs.oxide.computer/guides/operator/silo-management',
+  systemUpdateDocs: 'https://docs.oxide.computer/guides/operator/system-update',
```
david-crespo, line 60: Doc added in https://github.com/oxidecomputer/docs/pull/658

PR #2922, opened 2025-10-15
Style polish on update release cards

david-crespo:
Definitely an improvement, but kind of weird that properties table keeps shrinking and does `overflow: hidden` instead of scrolling itself or keeping a min width.

### Before

image

### After

https://github.com/user-attachments/assets/c9fa44f0-0f8b-45b1-aa29-3256606c3dbe


david-crespo:
Better. Also made the icons gray and changed the time one to the new outline icon. Ideally the icons would not move down (staying centered) in the card when it wraps, but it doesn't seem worth the effort. 

image

image

image

PR #2923, opened 2025-10-15
Add `AGENTS.md`

david-crespo:
I called it `AGENTS.md` because that's the standard name and it's only Claude Code that's being stubborn about it. `CLAUDE.md` is symlinked to `AGENTS.md`.

It seems like kind of too much to me as-is. Not sure it needs to be this detailed. I was picturing a list of exemplary files of each kind instead, like a list page, a side modal form, a page with tabs, a layout, stuff about routes and tests. On the other hand it's all legit. My main concern is that dropping 4k tokens in right off the bat is a waste.


david-crespo:
Curious how it compares to yours, @benjaminleonard

PR #2926, opened 2025-10-15
Add SCIM UI

### LINKED ISSUE #2885: Want support for SCIM in silo creation
Today, when fleet admin sets up a new silo, the two IDP modes available are JIT-SAML and local user auth. SCIM-SAML will soon be available as another option user can choose.

1. JIT and SCIM modes are mutually exclusive.
2. The silo IDP mode will remain immutable, i.e., a silo can't be changed from JIT to SCIM and vice versa.
3. The SCIM attributes to be provided by fleet admin at setup time are:
  - IDP system (in our initial implementation, only Okta is supported)
  - Attribute mapping (username, maybe email + other attributes - it'll be a short list since we don't use/track most of them)
4. An initial bearer token will be generated during silo setup which fleet admin will specify on the IDP side for authorizing SCIM API requests.

Here is an example of SCIM setup in Duo for some ideas of how the config looks like: https://duo.com/docs/oktasync.

@jmpesp @papertigers - Please update the above as needed.

### LINKED ISSUE #2886: Want SCIM bearer token CRD UI
Silo administrator should be able to view, regenerate, and delete SCIM bearer tokens.

Note: Silo viewers will **not** have access to bearer tokens as they are sensitive information, similar to silo TLS certicates.

charliepark:
This sets up the system management UI to handle SCIM tokens.
Screenshot 2025-10-15 at 7 05 11 AM
Screenshot 2025-10-15 at 7 05 17 AM
Screenshot 2025-10-15 at 7 05 24 AM

A few outstanding things to note:
- "Learn more about SCIM Auth" link is currently not working.
- For consistency, we'll want to make pagination similar to other pseudo-paginated endpoints, like https://github.com/oxidecomputer/omicron/blob/d74f5e3f1ae0a378dcdb9795a0ada2426702b046/nexus/src/external_api/http_entrypoints.rs#L5392
- Name attribute is in designs and might get added to API at some point, but not present yet.

Closes #2885
Closes #2886

### DIFF
--- a/app/pages/system/silos/SiloScimTab.tsx
+++ b/app/pages/system/silos/SiloScimTab.tsx
@@ -25,7 +25,7 @@ import { addToast } from '~/stores/toast'
 import { useColsWithActions, type MenuAction } from '~/table/columns/action-col'
 import { Columns } from '~/table/columns/common'
 import { Table } from '~/table/Table'
-import { CardBlock, LearnMore } from '~/ui/lib/CardBlock'
+import { CardBlock } from '~/ui/lib/CardBlock'
 import { CopyToClipboard } from '~/ui/lib/CopyToClipboard'
 import { CreateButton } from '~/ui/lib/CreateButton'
 import { DateTime } from '~/ui/lib/DateTime'
@@ -34,7 +34,6 @@ import { Message } from '~/ui/lib/Message'
 import { Modal } from '~/ui/lib/Modal'
 import { TableEmptyBox } from '~/ui/lib/Table'
 import { Truncate } from '~/ui/lib/Truncate'
-import { links } from '~/util/links'
 
 const colHelper = createColumnHelper<ScimClientBearerToken>()
 
@@ -151,9 +150,10 @@ export default function SiloScimTab() {
             />
           )}
         </CardBlock.Body>
+        {/* TODO: put this back!
         <CardBlock.Footer>
           <LearnMore href={links.scimDocs} text="SCIM" />
-        </CardBlock.Footer>
+        </CardBlock.Footer> */}
       </CardBlock>
 
       {showCreateModal && (
--- a/app/util/links.ts
+++ b/app/util/links.ts
@@ -130,10 +130,6 @@ export const docLinks = {
     href: links.routersDocs,
     linkText: 'Custom Routers',
   },
-  scim: {
-    href: links.scimDocs,
-    linkText: 'SCIM',
-  },
   sleds: {
     href: links.sledDocs,
     linkText: 'Server Sleds',


app/pages/system/silos/SiloScimTab.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+        <div className="mt-4">
+          <div className="text-sans-semi-md mb-2 font-medium">Bearer Token</div>
+          <div className="text-sans-sm text-raise bg-default border-default flex items-center rounded border font-mono">
+            <div className="flex-1 overflow-hidden p-3 text-ellipsis">
+              {token.bearerToken}
+            </div>
+            <div className="border-default border-l p-3">
+              <CopyToClipboard text={token.bearerToken} />
+            </div>
```
david-crespo, line 251: Think we need design polish from @benjaminleonard. I'd put the yellow bit at the top, for one. Maybe put the expiration at the bottom and display it in one line. Could do without the pseudo form field styling.

Image

david-crespo: heh just kidding. it's a lot closer to the design than I expected. The pseudo-input it too tall though.

image

charliepark:
Most recent change gives a width to the "Expires" column. When it had actual datetimes it looked somewhat normal, but with just the badge it was showing up like this:
Screenshot 2025-10-15 at 8 21 16 PM
So this commit gives the table this look on desktops:
Screenshot 2025-10-15 at 8 25 34 PM

mock-api/msw/handlers.ts
```diff
@@ -1888,6 +1888,49 @@ export const handlers = makeHandlers({
...
+  scimTokenCreate({ query, cookies }) {
+    requireFleetCollab(cookies)
+    const silo = lookup.silo({ silo: query.silo })
+
+    // Generate a 32-character hex string
+    const hexString = uuid().replace(/-/g, '')
+
+    const newToken: Json<Api.ScimClientBearerTokenValue> = {
+      id: uuid(),
+      bearer_token: `scim_${hexString}`,
```
askfongjojo: Shouldn't this be `oxide-scim-`? https://github.com/oxidecomputer/omicron/blob/d743754bb4be24228e9e042ce5262c242d4fd079/nexus/auth/src/authn/external/scim.rs#L22

david-crespo: Yep. I think this was done before that prefix was there.

david-crespo:
Added a docs link to the card and tweaked the subheading:

Screenshot 2025-10-28 at 5 22 02 PM

Updated this warning in consultation with @jmpesp:

Screenshot 2025-10-28 at 5 22 07 PM

Changed the modal title here, otherwise it's the same:

Screenshot 2025-10-28 at 5 22 12 PM

Added SAML + SCIM to the silo create form. Not sure about the spaces around the +, maybe it should be `SAML+SCIM`. I do see that in two comments in omicron but nowhere else.

image

david-crespo (approved):
This is ready to go pending two things:

* Nailing down the docs link (I put `/guides/operator/scim`, but the release notes PR https://github.com/oxidecomputer/docs/pull/658 suggests it will be here https://docs.oxide.computer/guides/operator/silo-management#_configure_identity_provider)
* Merging https://github.com/oxidecomputer/omicron/pull/9301 to make sure the `oxide-scim-` prefixing behavior in the mock API is right

david-crespo:
Took out the docs link, merging. https://github.com/oxidecomputer/console/issues/2953

PR #2951, opened 2025-10-28
Bump Omicron version and add IP Pool type field to IP Pool create form

charliepark:
This PR bumps Omicron to the latest version. To do this, we need to add a `unicast`/`multicast` field to the IP Pools' create form, and also a column to the table showing IP Pools.
Screenshot 2025-10-28 at 3 20 08 PM
Screenshot 2025-10-28 at 2 48 57 PM


david-crespo (approved):
Looks good. Something I am wondering about for the future is whether we should be linking to the docs on more forms. We have it on the list page, but if you're in the form and you want to know what unicast/multicast is, you want the docs, and it would be great to just have a link right there.

Image

PR #2957, opened 2025-10-30
Handle fleet viewers without permission to see SCIM tokens

david-crespo:
Ended up doing quite a bit of shuffling around, but the main thing is switching to `apiqErrorsAllowed` and displaying a different thing when it comes back as an error.

Jane Austen is a fleet viewer but not a fleet or silo admin:

image

### DIFF
--- a/app/pages/system/silos/SiloScimTab.tsx
+++ b/app/pages/system/silos/SiloScimTab.tsx
@@ -17,7 +17,6 @@ import { Badge } from '@oxide/design-system/ui'
 
 import {
   apiqErrorsAllowed,
-  apiQueryClient,
   queryClient,
   useApiMutation,
   usePrefetchedQuery,
@@ -196,7 +195,7 @@ function CreateTokenModal({
 }) {
   const createToken = useApiMutation('scimTokenCreate', {
     onSuccess(token) {
-      apiQueryClient.invalidateQueries('scimTokenList')
+      queryClient.invalidateEndpoint('scimTokenList')
       onSuccess(token)
     },
     onError(err) {


app/pages/system/silos/SiloScimTab.tsx
```diff
@@ -50,151 +71,131 @@ const EmptyState = () => (
...
+            siloSelector={siloSelector}
+            onDismiss={() => setModalState(false)}
+            onSuccess={(token) => setModalState({ kind: 'created', token })}
+          />
+        ))
+        .with({ kind: 'created' }, ({ token }) => (
+          <TokenCreatedModal token={token} onDismiss={() => setModalState(false)} />
+        ))
+        .with(false, () => null)
+        .exhaustive()}
```
david-crespo, line 141: addicted to `ts-pattern`

app/api/hooks.ts
```diff
@@ -245,8 +245,8 @@ export const getApiQueryOptionsErrorsAllowed =
       queryFn: ({ signal }) =>
         api[method](params, { signal })
           .then(handleResult(method))
-          .then((data) => ({ type: 'success' as const, data }))
-          .catch((data) => ({ type: 'error' as const, data })),
+          .then((data: Result<A[M]>) => ({ type: 'success' as const, data }))
+          .catch((data: ApiError) => ({ type: 'error' as const, data })),
```
david-crespo, line 249: This was a bug in `apiqErrorsAllowed` causing the `data` field to have type `any` in both arms. There is a better way to do this without casting but I'm going to do it later.

app/pages/system/silos/SiloScimTab.tsx
```diff
@@ -50,151 +71,131 @@ const EmptyState = () => (
...
-  await apiQueryClient.prefetchQuery('scimTokenList', { query: { silo } })
+  // Use errors-allowed approach so 403s don't throw and break the loader
+  await queryClient.prefetchQuery(apiqErrorsAllowed('scimTokenList', { query: { silo } }))
   return null
 }
 
+type ModalState =
+  | { kind: 'create' }
+  | { kind: 'created'; token: ScimClientBearerTokenValue }
+  | false
```
david-crespo, line 82: this makes it impossible to accidentally have both modals rendered at once

app/pages/system/silos/SiloScimTab.tsx
```diff
@@ -50,151 +71,131 @@ const EmptyState = () => (
...
+          <TokenCreatedModal token={token} onDismiss={() => setModalState(false)} />
+        ))
+        .with(false, () => null)
+        .exhaustive()}
+    </>
+  )
+}
 
+function TokensTable({ tokens }: { tokens: ScimClientBearerToken[] }) {
+  const siloSelector = useSiloSelector()
```
david-crespo, line 147: Moved the table logic to its own component that only renders on success responses so we don't have to pollute with if-else checks.

PR #2960, opened 2025-10-31
Limited collaborator UI

charliepark:
Now that https://github.com/oxidecomputer/omicron/pull/9299/ is in Omicron, we want to use the new permissions in the web console as well. This PR bumps Omicron (will want to bump again before merging this) and also redesigns the access forms to use radio buttons, which allow us to show all options on the screen, with descriptions of what each role enables.

Screenshot 2025-11-03 at 4 34 06 PM

### DIFF
--- a/app/forms/access-util.tsx
+++ b/app/forms/access-util.tsx
@@ -36,19 +36,18 @@ export const defaultValues: AddUserValues = {
 
 // Role descriptions for project-level roles
 const projectRoleDescriptions: Record<RoleKey, string> = {
-  admin: 'Can control all aspects of the project.',
-  collaborator: 'Can manage all resources, including networking.',
-  limited_collaborator: 'Can manage compute resources. Cannot manage networking.',
-  viewer: 'Can read most resources within the project.',
+  admin: 'Control all aspects of the project',
+  collaborator: 'Manage all project resources, including networking',
+  limited_collaborator: 'Manage project resources except networking configuration',
+  viewer: 'View resources within the project',
 }
 
 // Role descriptions for silo-level roles
 const siloRoleDescriptions: Record<RoleKey, string> = {
-  admin: 'Can control all aspects of the silo.',
-  collaborator: 'Can create and own projects. Grants project admin role on all projects.',
-  limited_collaborator:
-    'Can read most resources within the silo. Grants limited collaborator role on all projects.',
-  viewer: 'Can read most resources within the silo. Grants project viewer role.',
+  admin: 'Control all aspects of the silo',
+  collaborator: 'Create and administer projects',
+  limited_collaborator: 'Manage project resources except networking configuration',
+  viewer: 'View resources within the silo',
 }
 
 export const actorToItem = (actor: Actor): ListboxItem => ({


app/forms/access-util.tsx
```diff
@@ -55,3 +76,58 @@ export type EditRoleModalProps = AddRoleModalProps & {
...
 }
+
+type RoleRadioFieldProps = {
+  control: Control<AddUserValues> | Control<{ roleName: RoleKey }>
+  scope: 'Silo' | 'Project'
+}
+
+export function RoleRadioField({ control, scope }: RoleRadioFieldProps) {
+  const roleDescriptions = scope === 'Silo' ? siloRoleDescriptions : projectRoleDescriptions
+  const currentRole = control._formValues.roleName || ''
```
david-crespo: Shouldn't do this, should be something like:

```ts
const currentRole = useWatch({ control, name: 'roleName' })
```

david-crespo: 6e0af828b24528b61121ba7aa181cc0a7ae188c1 fixes this a different way: this only needed to be here because the mechanism inside `RadioFieldDyn` that checks the right radio based on the field value was getting messed up by a `div` being used to put space between the radios. I got rid of the div and fixed the field state thing, but now we need to figure out how to put the spacing back. The `className` prop on `Radio` is unfortunately not it because it doesn't get stuck on the outermost div. We could just add another prop for it, something silly like `containerClass`.

app/forms/access-util.tsx
```diff
@@ -14,20 +17,39 @@ import {
...
+import { HL } from '~/components/HL'
 import { type ListboxItem } from '~/ui/lib/Listbox'
+import { Message } from '~/ui/lib/Message'
+import { Radio } from '~/ui/lib/Radio'
 import { capitalize } from '~/util/str'
 
 type AddUserValues = {
   identityId: string
-  roleName: RoleKey | ''
+  roleName: RoleKey
```
david-crespo, line 29: This `''` value was me a long time ago, but I realized it's silly and we should just make viewer the default. This also required reversing the order so viewer is first, otherwise the default value is at the bottom, which is weird.

app/forms/access-util.tsx
```diff
@@ -14,20 +17,39 @@ import {
...
+  collaborator: 'Can manage all resources, including networking',
+  limited_collaborator: 'Can manage compute resources; can not manage networking',
+  viewer: 'Can read most resources within the project',
 }
 
-export const roleItems = allRoles.map((role) => ({ value: role, label: capitalize(role) }))
+// Role descriptions for silo-level roles
+const siloRoleDescriptions: Record<RoleKey, string> = {
+  admin: 'Can control all aspects of the silo',
+  collaborator: 'Can create and own projects; grants project admin role on all projects',
```
benjaminleonard: Semi-colon in micro copy feels wrong, not sure why!

charliepark:
> This should be a regular modal rather than a side modal, since it's a single input / action.

Okay; just for clarity, does the "Add user" modal remain a side modal, because you have to enter the name and select the role, or is it conceptually a single input and should be a regular modal?

app/forms/access-util.tsx
```diff
@@ -55,3 +77,45 @@ export type EditRoleModalProps = AddRoleModalProps & {
...
+
+export function RoleRadioField({ control, scope }: RoleRadioFieldProps) {
+  const roleDescriptions = scope === 'Silo' ? siloRoleDescriptions : projectRoleDescriptions
+  return (
+    <>
+      <RadioFieldDyn
+        name="roleName"
+        label={`${scope} role`}
+        required
+        control={control as Control<AddUserValues>}
```
david-crespo: Why this doesn't work without the cast: `RadioFieldProps` has this too-clever `parseValue` thing and then `RadioFieldDynProps` omits `parseValue`. I think this may be a TS bug. Here's a repro of the issue stripped down to two lines: https://tsplay.dev/ND13Om

I have two workarounds, one of which is to turn `parseValue` into a regular optional, the other is to inline the`RadioFieldDyn` call in the `RoleRadioField` call sites. I think I prefer the latter. I'll figure it out.

https://github.com/oxidecomputer/console/blob/6764ba573e461f0951324cc8c980e8cefb609f6b/app/components/form/fields/RadioField.tsx#L42-L51

https://github.com/oxidecomputer/console/blob/6764ba573e461f0951324cc8c980e8cefb609f6b/app/components/form/fields/RadioField.tsx#L102-L107

app/forms/instance-create.tsx
```diff
@@ -394,34 +394,19 @@ export default function CreateInstanceForm() {
             </Tabs.Trigger>
           </Tabs.List>
           <Tabs.Content value="general">
-            <RadioFieldDyn
-              name="presetId"
-              label=""
-              control={control}
-              disabled={isSubmitting}
-            >
+            <RadioFieldDyn name="presetId" control={control} disabled={isSubmitting}>
```
david-crespo, line 397: these were the only other `RadioFieldDyn` and they were using `label=""` as a hack around the fact there was a default value if label wasn't passed. That's silly, so I got rid of the default value and stopped passing a label here.

app/components/form/fields/RadioField.tsx
```diff
@@ -126,9 +140,13 @@ export function RadioFieldDyn<
...
   const id = useId()
   const { field } = useController({ name, control })
+  invariant(
+    isOneOf(children, [Radio, RadioCard]),
+    'Children of RadioFieldDyn must be Radio or RadioCard'
+  )
   return (
     <div>
-      <div className="mb-2">
+      <div className="mb-3">
```
david-crespo, line 149: The fact that the only existing callers don't pass a label means that I can change the spacing here without consequence. Looks fine. I considered `mb-4` but it was slightly too much. I think the real issue here is that the label and the values have the same treatment, so there's no sense of hierarchy. A problem for another time.

Image

app/forms/instance-create.tsx
charliepark: nice

david-crespo:
Final copy:

image

david-crespo (approved):
let's party

PR #2971, opened 2025-11-12
refactor: redo API helpers

david-crespo:
These types have kind of been my nemesis since I put them together in #222. Last week I had the revelation that you can use `api.methods.siloView.name` to get the method name string we need for the query key. I had been passing the method name as a string forever in order to have that string on hand. But if we pass the function directly, inferring the request and response types is trivial: they are just the parameters and return type of the function.

From the callsites, the change looks like this.

```diff
- const { data: users } = usePrefetchedQuery(apiq('userList', {}))
+ const { data: users } = usePrefetchedQuery(q(api.userList, {}))
```

No big deal, just a lot of lines.

### DIFF
--- a/app/api/client.ts
+++ b/app/api/client.ts
@@ -31,8 +31,12 @@ const _api = new Api({
   host: process.env.NODE_ENV === 'test' ? 'http://testhost' : '',
 })
 
+// Pull the methods off to make the call sites shorter: `api.siloView`
+// instead of `api.methods.siloView`. We only put this method field there in
+// the generator to make the old way of typing `q` work, so we don't need it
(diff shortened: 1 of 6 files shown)


app/api/hooks.ts
```diff
@@ -85,25 +79,6 @@ type UseQueryOtherOptions<T> = Omit<
   'queryKey' | 'queryFn' | 'initialData'
 >
 
-export const getApiQueryOptions =
-  <A extends ApiClient>(api: A) =>
-  <M extends string & keyof A>(
-    method: M,
-    params: Params<A[M]>,
```
david-crespo, line 92: Good riddance!

PR #2977, opened 2025-11-18
Fetch and prefill VPC list in NIC forms

### LINKED ISSUE #2051: Preselect first VPC in custom NIC form on instance create
There's almost always exactly one VPC. Just select it.

charliepark:
This prefetches the list of VPCs and selects the first one in the "select a VPC" form fields in NIC create form and in the custom NIC subsection of the instance create form.

I accidentally picked up console/2051 instead of omicron/2051 and figured I would just knock it out.

It's worth noting that the NIC form contains both a VPC and a Subnet, which is based on the VPC. At the moment, this implementation only pre-fills the VPC, though we could chain the calls and pre-fill the Subnet as well with just the first one in the list. I think it's better to let the user pick the Subnet, though? Open to thoughts.

Closes https://github.com/oxidecomputer/console/issues/2051


app/forms/network-interface-create.tsx
```diff
@@ -47,12 +53,20 @@ export function CreateNetworkInterfaceForm({
...
 
   const form = useForm({ defaultValues })
 
+  // prefill form with first VPC
+  useEffect(() => {
+    if (vpcs.length > 0 && !form.getValues('vpcName')) {
+      form.setValue('vpcName', vpcs[0].name)
+    }
+  }, [vpcs, form])
+
```
david-crespo, line 89: this should be done through `defaultValues` since it’s prefetched.

app/forms/network-interface-create.tsx
```diff
@@ -47,12 +53,20 @@ export function CreateNetworkInterfaceForm({
   submitError = null,
 }: CreateNetworkInterfaceFormProps) {
   const projectSelector = useProjectSelector()
-
-  const { data: vpcsData } = useQuery(q(api.vpcList, { query: projectSelector }))
-  const vpcs = useMemo(() => vpcsData?.items || [], [vpcsData])
+  const { data: vpcsData } = usePrefetchedQuery(
+    q(api.vpcList, { query: { ...projectSelector, limit: ALL_ISH } })
+  )
+  const vpcs = useMemo(() => vpcsData.items, [vpcsData])
```
david-crespo: useMemo unnecessary

app/forms/network-interface-create.tsx
```diff
@@ -47,12 +43,17 @@ export function CreateNetworkInterfaceForm({
...
 }: CreateNetworkInterfaceFormProps) {
   const projectSelector = useProjectSelector()
-
-  const { data: vpcsData } = useQuery(q(api.vpcList, { query: projectSelector }))
-  const vpcs = useMemo(() => vpcsData?.items || [], [vpcsData])
-
+  const {
+    data: { items: vpcs },
+  } = usePrefetchedQuery(q(api.vpcList, { query: { ...projectSelector, limit: ALL_ISH } }))
+  const defaultValues = {
```
david-crespo: As ugly as it is, it's helpful to get an error on this line if the type changes.

```suggestion
  const defaultValues: SetRequired<SetNonNullable<InstanceNetworkInterfaceCreate>, 'ip'> = {
```

PR #2988, opened 2025-12-14
fix(readme): RFD 223 internal note

aapoalas:
Just a drive-by as I was reading the readme: RFD 223 is no longer an internal document.


david-crespo:
Thanks!

PR #2989, opened 2025-12-16
Show a message that this site requires JavaScript

### LINKED ISSUE #2928: Fallback message for when JavaScript is turned off
You get a blank screen when you have JS off. Now, anyone browsing with JS off by default is probably used to this and knows the deal, but it would be nice to say "This site requires JavaScript." Probably as simple as adding a `<noscript>` tag here:

https://github.com/oxidecomputer/console/blob/690bde05504fc4dc68171cf4bda932bf0aedebbb/index.html#L21-L24

Image

oblakeerickson:
Adds a noscript tag that displays a message that this site requires JavaScript to run.

The easiest way to test this locally is to use Chrome since it can disable JavaScript per tab instead of globally. Open the dev console and click the gear icon and check "Disable "JavaScript" in the Debugger section.

I skipped adding a test since it is a static change and playwright requires js to run.

resolves #2928

### DIFF
--- a/index.html
+++ b/index.html
@@ -23,7 +23,7 @@
     <noscript>
       <div class="min-h-screen grid place-content-center">
         <p class="max-w-lg p-8 text-center text-balance">
-          This site requires JavaScript. If you don't want to enable JavaScript, see the <a href="https://docs.oxide.computer/" target="_blank" rel="noopener noreferrer">Oxide docs</a> to learn about accessing the API through the CLI or SDKs.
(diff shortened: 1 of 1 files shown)


oblakeerickson:
And this is how it looks:

CleanShot 2025-12-16 at 11 31 24@2x

CleanShot 2025-12-16 at 11 36 23@2x

david-crespo:
Thanks for getting this started. I tweaked the colors and layout a bit.

image

david-crespo:
It turns out we have the CSS in prod, just not in local dev, which loads the CSS through JS. You can test this with `npm run preview`.

https://github.com/user-attachments/assets/9455fc49-9258-4c60-93fa-b72f3fda923e

PR #2992, opened 2025-12-17
Disk details side modal

david-crespo:
This change adds disk detail side modals to the disk list page and the instance disks tab. The one on the disk list gets its own route, but since the one on instances is already nested under the storage tab, I decided it's probably not worth giving it a path. I experimented with `?disk=xyz` in the query string, but it feels a bit pointless and complicates the code.

The motivation for this is that while working on local disks in #2984, I talked to @jmpesp and we agreed it is hard to make sense of the error that you can't attach a local disk associated with one sled to an instance with local disks on another sled. To help the user understand and deal with this, we want to put sled ID in the local disk response. Without this detail modal, there would be no place to display that sled ID.

The contents of the modal will have to be updated in #2984. I just wanted to keep the bulk of this change outside of that PR.

While working on this I noticed that the `AGENTS.md` was too long and also out of date in some ways, so I cleaned that up too.

Screenshot 2025-12-17 at 3 49 21 PM

Screenshot 2025-12-17 at 3 49 10 PM

### DIFF
--- a/app/forms/idp/edit.tsx
+++ b/app/forms/idp/edit.tsx
@@ -117,7 +117,7 @@ export default function EditIdpSideModalForm() {
       <TextField
         name="groupAttributeName"
         label="Group attribute name"
-        description="Name of the SAML attribute in the IdP response listing the user's groups"
+        description="Name of the SAML attribute in the IdP response listing the user’s groups"
         required
         control={form.control}
         disabled


app/pages/project/disks/DiskDetailSideModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+  disk,
+  onDismiss,
+  animate = true,
+}: DiskDetailSideModalProps) {
+  return (
+    <SideModal
+      isOpen
+      onDismiss={onDismiss}
+      animate={animate}
+      title="Disk details"
```
david-crespo: I would prefer to call this `Disk`, but it looks stupid. The other modals like this (project image, silo image, identity provider) are all naturally two-word titles. @benjaminleonard @charliepark if you think it looks fine as `Disk`, I'm fine changing it.

Image

test/e2e/snapshots.e2e.ts
```diff
@@ -25,14 +25,26 @@ test('Click through snapshots', async ({ page }) => {
...
 test('Confirm delete snapshot', async ({ page }) => {
   await page.goto('/projects/mock-project/snapshots')
 
   const row = page.getByRole('row', { name: 'disk-1-snapshot-10' })
 
-  // scroll a little so the dropdown menu isn't behind the pagination bar
-  await page.getByRole('table').click() // focus the content pane
-  await page.mouse.wheel(0, 200)
+  // scroll so the dropdown menu isn't behind the pagination bar
+  await row.scrollIntoViewIfNeeded()
```
david-crespo, line 47: Claude figured out this test was breaking with an unwanted nav to disk detail because the click on the table was landing right in the middle of the table, on a disk link cell!

app/pages/project/snapshots/SnapshotsPage.tsx
```diff
@@ -39,11 +40,14 @@ import { docLinks } from '~/util/links'
...
 const DiskNameFromId = ({ value }: { value: string }) => {
+  const { project } = useProjectSelector()
   const { data } = useQuery(qErrorsAllowed(api.diskView, { path: { disk: value } }))
 
   if (!data) return <SkeletonCell />
   if (data.type === 'error') return <Badge color="neutral">Deleted</Badge>
-  return <span className="text-default">{data.data.name}</span>
+  return (
+    <LinkCell to={pb.disk({ project, disk: data.data.name })}>{data.data.name}</LinkCell>
+  )
```
david-crespo: Since the details view is just a side modal, we could theoretically just open it right on the snapshots page and it would work fine. The only problem with that is that it wouldn't be able to have a URL, which is probably also fine. I feel a little weird about there being two places to see disk details. On the other hand the current interaction is kind of jarring. It feels like a long way to go to nav to the disks page with a side modal open.

benjaminleonard:
It's more or less okay, but I want to prod at the design a little. Solo properties table looks funky. Are you in a rush to merge?

app/components/form/SideModalForm.tsx
```diff
@@ -104,7 +104,7 @@ export function SideModalForm<TFieldValues extends FieldValues>({
       <SideModal.Body>
         <form
           id={id}
-          className="ox-form is-side-modal"
+          className="ox-form"
```
david-crespo, line 107: The `is-side-modal` styling mechanism was removed in https://github.com/oxidecomputer/console/pull/2815

PR #2993, opened 2025-12-18
minor: try to fix firewall rule test flake again

david-crespo:


david-crespo:
Three runs, one flake, no failures.

app/forms/firewall-rules-common.tsx
```diff
@@ -494,6 +494,12 @@ const ProtocolFilters = ({ control }: { control: Control<FirewallRuleValues> })
                   placeholder=""
                   validate={icmpCodeValidation}
                   transform={normalizeDashes}
+                  onKeyDown={(e) => {
+                    if (e.key === KEYS.enter) {
+                      e.preventDefault() // prevent full form submission
+                      submitProtocol(e)
+                    }
+                  }}
```
david-crespo, line 502: This wasn't causing the flake because we weren't pressing enter in this field in the test, but it was wrong. You could submit the whole form by pressing enter in this field.

PR #2997, opened 2025-12-23
trivial: move prettier-ignore comments to the line before

david-crespo:
Workaround for https://github.com/oxc-project/oxc/issues/16915 before we switch to oxfmt.

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,9 +21,17 @@
   <body>
     <div id="root"></div>
     <noscript>
-      <div class="min-h-screen grid place-content-center">
+      <div class="grid min-h-screen place-content-center">
         <p class="max-w-lg p-8 text-center text-balance">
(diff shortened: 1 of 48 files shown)


index.html
```diff
@@ -21,9 +21,17 @@
...
-          This site requires JavaScript. If you don't want to enable JavaScript, see the <a href="https://docs.oxide.computer/" class="external-link" target="_blank" rel="noopener noreferrer">Oxide docs</a> to learn about accessing the API through the CLI or SDKs.
+          This site requires JavaScript. If you don't want to enable JavaScript, see the
+          <a
+            href="https://docs.oxide.computer/"
+            class="external-link"
+            target="_blank"
+            rel="noopener noreferrer"
+            >Oxide docs</a
+          >
+          to learn about accessing the API through the CLI or SDKs.
```
david-crespo, line 34: This is due to running the formatter after the change from #2989. It seems our format check, which uses the linter, does not fail if this file isn't formatted. We should really be using the formatter itself to do the format check. Or whatever is most idiomatic in `oxfmt` and `oxlint`. To be continued.

PR #2998, opened 2025-12-23
Fully switch to oxlint

david-crespo:
Oxlint runs in 800ms without caching. Eslint runs in 14 seconds without caching, 600ms with a warm cache and no changes. So Oxlint without caching is about the same as eslint with caching.

```
$ time npm run lint

> oxide@0.0.0 lint
> eslint --ext .js,.ts,.tsx app test mock-api

npm run lint  21.42s user 2.09s system 166% cpu 14.150 total

$ time npm run lint -- --cache

> oxide@0.0.0 lint
> eslint --ext .js,.ts,.tsx app test mock-api --cache

npm run lint -- --cache  0.69s user 0.22s system 83% cpu 1.083 total

$ time npm run lint -- --cache

> oxide@0.0.0 lint
> eslint --ext .js,.ts,.tsx app test mock-api --cache

npm run lint -- --cache  0.67s user 0.17s system 144% cpu 0.579 total
```

We've been running both eslint and oxlint for a while. Now that oxlint supports type-aware rules and JS plugins, I think it's time to drop eslint. The playwright plugin runs via oxlint's experimental [JS plugin support](https://oxc.rs/blog/2025-10-09-oxlint-js-plugins.html). We do lose [`eslint-plugin-react-hook-form`](https://github.com/andykao1213/eslint-plugin-react-hook-form) for now because all 3 rules use `context.getScope()`, which [oxlint doesn't support yet](https://oxc.rs/blog/2025-10-09-oxlint-js-plugins.html). I'll make an issue to add it back when it's supported.

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -13,6 +13,7 @@
     "jsx-a11y",
     "promise",
     // defaults (see https://oxc.rs/docs/guide/usage/linter/plugins.html#supported-plugins)
+    "eslint",
     "unicorn",
     "typescript",
     "oxc"
--- a/app/components/form/fields/RadioField.tsx
+++ b/app/components/form/fields/RadioField.tsx
@@ -80,6 +80,7 @@ export function RadioField<
       </div>
       <RadioGroup
         defaultChecked={field.value}
+        // oxlint-disable-next-line jsx-a11y/aria-proptypes -- https://github.com/oxc-project/oxc/issues/17465
         aria-labelledby={`${id}-label`}
         onChange={(e) =>
           parseValue ? field.onChange(parseValue(e.target.value)) : field.onChange(e)
--- a/app/components/form/fields/TextField.tsx
+++ b/app/components/form/fields/TextField.tsx
@@ -117,6 +117,7 @@ export const TextFieldInner = <
         title={label}
         type={type}
         error={!!error}
+        // oxlint-disable-next-line jsx-a11y/aria-proptypes -- https://github.com/oxc-project/oxc/issues/17465
         aria-labelledby={`${id}-label ${id}-help-text`}
         onChange={(e) => onChange(transform ? transform(e.target.value) : e.target.value)}
         {...fieldRest}
--- a/app/table/cells/BooleanCell.tsx
+++ b/app/table/cells/BooleanCell.tsx
@@ -6,7 +6,6 @@
  * Copyright Oxide Computer Company
  */
 
-/* eslint-disable jsx-a11y/aria-proptypes */
 // there seems to be a bug in the linter. it doesn't want you to use the string
 // "true" because it insists it's a boolean
 import { Disabled12Icon, Success12Icon } from '@oxide/design-system/icons/react'
--- a/package-lock.json
+++ b/package-lock.json
@@ -1793,9 +1793,9 @@
       ]
     },
     "node_modules/@oxlint/darwin-arm64": {
-      "version": "1.35.0",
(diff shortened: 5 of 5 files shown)


.oxlintrc.json
```diff
@@ -1,5 +1,12 @@
 {
   "$schema": "./node_modules/oxlint/configuration_schema.json",
+  "jsPlugins": [
+    {
+      "name": "playwright",
+      "specifier": "eslint-plugin-playwright"
+    }
+    // eslint-plugin-react-hook-form won't work — all its rules use context.getScope()
```
connorshea, line 8: Is this still the case? I believe getScope() may be implemented on the most recent releases of oxlint? https://github.com/oxc-project/oxc/blob/e9d346f1f219025229013da47386d387823c220e/apps/oxlint/src-js/plugins/scope.ts#L359

david-crespo: Looks like it, though only the access control rule errors like this. The other two don't seem to work, though — at least with the destructuring rule, I see things that should fail but aren't getting caught.

```
  × Error running JS plugin.
  │ File path: /Users/david/oxide/console/app/components/form/fields/TlsCertsField.tsx
  │ TypeError: context.getScope is not a function
  │     at check (/Users/david/oxide/console/node_modules/eslint-plugin-react-hook-form/lib/rules/no-access-control.js:43:12)
  │     at file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:20177:17
  │     at walkVariableDeclarator (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:18336:70)
  │     at walkNode (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:17746:4)
  │     at walkNode (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:17467:33)
  │     at walkVariableDeclaration (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:18332:109)
  │     at walkNode (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:17743:4)
  │     at walkNode (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:17467:33)
  │     at walkBlockStatement (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:18108:109)
  │     at walkNode (file:///Users/david/oxide/console/node_modules/oxlint/dist/lint.js:17575:4)
```

connorshea: Warnings/errors aren't integrated into the editor extension yet for JS plugin violations, just in case you're expecting those to show up. 

Otherwise, may be worth opening an issue in the oxc repo about this particular plugin having problems so they can be investigated. 🙇‍♂️

david-crespo: Thanks, putting it on my todo list.

david-crespo: While putting together a repro at https://github.com/david-crespo/oxlint-rhf-plugin-repro, Claude notes:

> Note: Had to use ESLint 8.57.1 because `eslint-plugin-react-hook-form` also uses `context.getScope()` which was removed in ESLint 9 (replaced with `sourceCode.getScope()`). The plugin itself needs updating for ESLint 9 compatibility.

This could explain the above.

connorshea: @david-crespo Looks like that is planned for support, though it won't be until late next month at the absolute earliest: https://github.com/oxc-project/oxc/issues/17130

Reach out if you run into any other problems!

david-crespo: Thank you! Subscribed to that issue.

PR #3005, opened 2026-01-09
Update to new networking API shape, with IPv6

### LINKED ISSUE #2901: V4/V6 radio on IP pool create
Basically redo https://github.com/oxidecomputer/console/pull/2897/commits/b21f0b159b89d55f88be3715b627a2473b1f56a1 when https://github.com/oxidecomputer/omicron/issues/8761 is in. Also make sure adding IPv6 ranges works.

### LINKED ISSUE #3000: IPv6 support
A number of places in the console may need modifications for IPv6 address input/display:

- **IP Pool | Add IP range**
  - remove comment (and any validation) around allowing IPv4 range only
- **Floating IP | Create Floating IP**
  - add IP version radio button to select the appropriate default pool
  - alternatively, user can be prompted to select a specific pool under **Advanced** if the silo has both default v4 and v6 pools
- <s>**VPC subnet | Create Subnet**
  - allow IPv6-only subnet? (currently v4 is mandatory, v6 is optional)</s>
- **Instance | Networking | Network Interfaces (grid)**
  - display also v6 IP and Transit IPs if they are not null 
- **Instance | Networking | Add Network Interface**
  - have two separate IP address input for specific v4 and v6 addresses (currently there is one input field)
- **Instance | Networking | Edit Network Interface**
  - have two separate Transit IP address input fields for v4 and v6 addresses (currently there is one)

Before making child tickets for these items, it'll be good for @bnaecker to confirm what is necessary vs nice-to-have vs N/A.

charliepark:
This PR updates console to work with the new API shape for IP pool selection introduced in https://github.com/oxidecomputer/omicron/pull/9598 and other Omicron networking PRs, like https://github.com/oxidecomputer/omicron/issues/9665. These are summarized in https://github.com/oxidecomputer/console/issues/3000, and I have screenshots of each of the UX changes over in #3017.

Separately, this PR removes the form field in the instance create flow that allows the user to change the hostname, and now uses the instance name (as it did before, as the default).

I'll go over the main changes and how they relate to Console, both for us and for aipr as we go through it. There are still a few decisions to make in terms of forms / UI, but I'll try to note those as well.

Claude's summary of the changes in this PR:

## Key API Shape Changes

1. Pool Selection: pool → poolSelector (Breaking Change)

Before:
pool?: NameOrId | null

After:
poolSelector?: PoolSelector  // discriminated union

type PoolSelector =
  | { type: 'explicit'; pool: NameOrId }
  | { type: 'auto'; ipVersion?: IpVersion | null }  // Required when dual defaults exist

Why: Enables explicit IPv4 vs IPv6 selection when both default pools exist.

---
2. Private IPs: Single Field → Structured IP Stack (Major Breaking Change)

Before:
// Single IP + transit IPs
ip?: string
transitIps?: IpNet[]

After:
// Discriminated union supporting v4-only, v6-only, or dual-stack
ipConfig?: PrivateIpStackCreate

type PrivateIpStackCreate =
  | { type: 'v4'; value: PrivateIpv4StackCreate }
  | { type: 'v6'; value: PrivateIpv6StackCreate }
  | { type: 'dual_stack'; value: { v4: ..., v6: ... } }

Each stack has ip: Ipv4Assignment | Ipv6Assignment (auto or explicit) and version-specific transitIps.

---
3. Instance Network Interface Defaults: Expanded Options

Before:
type: 'default'  // Single default option

After:
type: 'default_ipv4' | 'default_ipv6' | 'default_dual_stack'

Why: Explicit control over default interface IP version(s).

---
4. Dual Defaults: Now Allowed

Before: Max 1 default pool per silo

After: Max 4 defaults per silo:
- 1 IPv4 unicast default
- 1 IPv6 unicast default
- 1 IPv4 multicast default
- 1 IPv6 multicast default

---
5. Pool Types: Unicast vs Multicast

Added poolType: 'unicast' | 'multicast' field:
- Ephemeral/floating IPs → unicast only
- UI now filters pools by type in relevant contexts

---
6. New Core Types

- IpVersion: 'v4' | 'v6' used throughout
- Ipv4Assignment / Ipv6Assignment: { type: 'auto' } | { type: 'explicit'; value: string }
- PrivateIpStack: Discriminated union for runtime IP stack representation
- AddressAllocator: New allocator pattern for floating IPs

---
UI/Form Impact Summary

| Form              | Old Pattern                                 | New Pattern                                               |
|-------------------|---------------------------------------------|-----------------------------------------------------------|
| Instance Create   | externalIps: [{ type: 'ephemeral', pool? }] | externalIps: [{ type: 'ephemeral', poolSelector? }]       |
| Floating IP       | pool?: string                               | poolSelector?: PoolSelector + ipVersion field             |
| Network Interface | Single ip?: string field                    | ipStackType + separate ipv4/ipv6 fields → builds ipConfig |

New Component: IpPoolSelector – radio-button UI for "IPv4 default" / "IPv6 default" / "custom pool" selection with version filtering.

Also, a quick note on selecting Ephemeral IPs for instances:

| Scenario                       | Client Specifies                                   | Behavior                                           |
|--------------------------------|----------------------------------------------------|----------------------------------------------------------------------|
| No unicast defaults            | { type: 'ephemeral' } or omit pool_selector        | FAILS with "No default IP pool configured for this silo"             |
| One unicast default (v4 OR v6) | { type: 'ephemeral' } (defaults to auto)           | SUCCEEDS - Uses the one default pool via Auto { ip_version: None }   |
| Dual defaults (v4 AND v6)      | { type: 'ephemeral' } without ip_version           | FAILS with "Multiple default pools... please specify ip_version"     |
| Dual defaults (v4 AND v6)      | { poolSelector: { type: 'auto', ipVersion: 'v4' }} | SUCCEEDS - Uses specified version's default pool                     |
| Multicast defaults only        | { type: 'ephemeral' }                              | FAILS - Ephemeral IPs require unicast pools (see external_ip.rs:177) |

---
Also …
Closes #2901 
Closes #3000

### DIFF
--- a/app/components/form/fields/IpPoolSelector.tsx
+++ b/app/components/form/fields/IpPoolSelector.tsx
@@ -7,7 +7,7 @@
  */
 import cn from 'classnames'
 import { useMemo } from 'react'
-import type { Control } from 'react-hook-form'
+import type { Control, FieldPath, FieldValues } from 'react-hook-form'
 import * as R from 'remeda'
 
 import {
@@ -42,10 +42,13 @@ function toIpPoolItem(p: SiloIpPool) {
 
 const ALL_IP_VERSIONS: IpVersion[] = ['v4', 'v6']
 
-type IpPoolSelectorProps = {
+type IpPoolSelectorProps<
+  TFieldValues extends FieldValues,
+  TName extends FieldPath<TFieldValues>,
+> = {
   className?: string
-  control: Control<any>
-  poolFieldName: string
+  control: Control<TFieldValues>
+  poolFieldName: TName
   pools: UnicastIpPool[]
   disabled?: boolean
   /** Compatible IP versions based on network interface type */
@@ -53,7 +56,10 @@ type IpPoolSelectorProps = {
   required?: boolean
 }
 
-export function IpPoolSelector({
+export function IpPoolSelector<
+  TFieldValues extends FieldValues,
+  TName extends FieldPath<TFieldValues>,
+>({
   className,
   control,
   poolFieldName,
@@ -61,7 +67,7 @@ export function IpPoolSelector({
   disabled = false,
   compatibleVersions = ALL_IP_VERSIONS,
   required = true,
-}: IpPoolSelectorProps) {
+}: IpPoolSelectorProps<TFieldValues, TName>) {
   // Note: pools are already filtered by poolType before being passed to this component
   const sortedPools = useMemo(() => {
     const compatPools = pools.filter(poolHasIpVersion(compatibleVersions))
--- a/app/forms/instance-create.tsx
+++ b/app/forms/instance-create.tsx
@@ -275,9 +275,17 @@ export default function CreateInstanceForm() {
     q(api.projectIpPoolList, { query: { limit: ALL_ISH } })
   )
 
-  // Only unicast pools can be used for ephemeral IPs
+  // Only unicast pools can be used for ephemeral IPs. Sort once here so
+  // downstream filters (default pool pick, compatible pool list) preserve
+  // the order without needing to re-sort.
   const unicastPools = useMemo(
-    () => (siloPools?.items || []).filter(isUnicastPool),
+    () =>
+      R.sortBy(
+        (siloPools?.items || []).filter(isUnicastPool),
+        (p) => !p.isDefault, // defaults first
+        (p) => p.ipVersion, // v4 first
+        (p) => p.name
+      ),
     [siloPools]
   )
 
@@ -783,35 +791,27 @@ const AdvancedAccordion = ({
     .filter((ip) => !!ip)
 
   // Filter unicast pools by compatible IP versions
-  const compatibleUnicastPools = useMemo(
+  // unicastPools is already sorted (defaults first, v4 first, then by name),
+  // so filtering preserves that order
+  const compatiblePools = useMemo(
     () => unicastPools.filter(poolHasIpVersion(compatibleVersions)),
     [unicastPools, compatibleVersions]
   )
-  const sortedPools = useMemo(
-    () =>
-      R.sortBy(
-        compatibleUnicastPools,
-        (p) => !p.isDefault, // false sorts first, so this defaults first
-        (p) => p.ipVersion, // sort v4 first
-        (p) => p.name
-      ),
-    [compatibleUnicastPools]
-  )
 
   useEffect(() => {
-    if (!assignEphemeralIp || sortedPools.length === 0) return
+    if (!assignEphemeralIp || compatiblePools.length === 0) return
 
     const currentPoolValid =
-      ephemeralIpPool && sortedPools.some((p) => p.name === ephemeralIpPool)
+      ephemeralIpPool && compatiblePools.some((p) => p.name === ephemeralIpPool)
     if (currentPoolValid) return
 
-    const defaultPool = sortedPools.find((p) => p.isDefault)
+    const defaultPool = compatiblePools.find((p) => p.isDefault)
     if (defaultPool) {
       ephemeralIpPoolField.field.onChange(defaultPool.name)
     } else {
       ephemeralIpPoolField.field.onChange('')
     }
-  }, [assignEphemeralIp, ephemeralIpPool, ephemeralIpPoolField, sortedPools])
+  }, [assignEphemeralIp, ephemeralIpPool, ephemeralIpPoolField, compatiblePools])
 
   // Track previous ability to attach ephemeral IP to detect transitions
   const prevCanAttachRef = useRef<boolean | undefined>(undefined)
@@ -819,9 +819,9 @@ const AdvancedAccordion = ({
   // Automatically manage ephemeral IP based on NIC and pool availability
   useEffect(() => {
     const hasCompatibleNics = compatibleVersions.length > 0
-    const hasPools = compatibleUnicastPools.length > 0
+    const hasPools = compatiblePools.length > 0
     const canAttach = hasCompatibleNics && hasPools
-    const hasDefaultPool = compatibleUnicastPools.some((p) => p.isDefault)
+    const hasDefaultPool = compatiblePools.some((p) => p.isDefault)
     const prevCanAttach = prevCanAttachRef.current
 
     if (!canAttach && assignEphemeralIp) {
@@ -839,16 +839,11 @@ const AdvancedAccordion = ({
     }
 
     prevCanAttachRef.current = canAttach
-  }, [
-    assignEphemeralIp,
-    assignEphemeralIpField,
-    compatibleUnicastPools,
-    compatibleVersions,
-  ])
+  }, [assignEphemeralIp, assignEphemeralIpField, compatiblePools, compatibleVersions])
 
   const ephemeralIpCheckboxState = useMemo(() => {
     const hasCompatibleNics = compatibleVersions.length > 0
-    const hasCompatiblePools = compatibleUnicastPools.length > 0
+    const hasCompatiblePools = compatiblePools.length > 0
     const canAttachEphemeralIp = hasCompatibleNics && hasCompatiblePools
 
     let disabledReason: React.ReactNode = undefined
@@ -874,7 +869,7 @@ const AdvancedAccordion = ({
     }
 
     return { canAttachEphemeralIp, disabledReason }
-  }, [compatibleVersions, compatibleUnicastPools])
+  }, [compatibleVersions, compatiblePools])
 
   const closeFloatingIpModal = () => {
     setFloatingIpModalOpen(false)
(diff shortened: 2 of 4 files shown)


app/util/ip-stack.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Extract transit IPs from an instance network interface's IP stack.
+ * For dual-stack, returns the IPv4 transit IPs (since transit IPs are generally version-agnostic in the UI).
+ */
+export function getTransitIpsFromStack(nic: InstanceNetworkInterface): string[] {
+  if (nic.ipStack.type === 'dual_stack') {
+    return nic.ipStack.value.v4.transitIps
+  }
+  return nic.ipStack.value.transitIps
+}
```
david-crespo: Seems like this file is unused.

david-crespo:
You have e2e test failures in IP pools.

playwright.config.ts
```diff
@@ -41,11 +42,21 @@ export default {
     },
     {
       name: 'firefox',
-      use: { ...devices['Desktop Firefox'] },
+      use: {
+        contextOptions: {
+          reducedMotion: 'reduce',
```
david-crespo, line 47: I'm assuming this was related to reducing flake?

charliepark: Yes; the theory from Claude was that with some animation changes in https://github.com/oxidecomputer/console/pull/3007 were causing some flaking in the e2e tests, though all of my local runs were working as expected, so I'm not sure if that was it. It just seemed to be an issue with the CI runs, and they ended up getting resolved by adding a `.blur` step, so it's possible the reducedMotion was moot. But if it speeds things CI runs up a tiny bit, and maybe fixes one or two flakey tests along the way, I'm okay with keeping it.

david-crespo: Not 100% sure, but I think only one or two of the side modals actually changed behavior in #3007 — mostly it was a slightly neater way of doing what we were already doing.

david-crespo: Sounds good to me, though! No need to have that delay in there for the tests.

app/components/form/fields/ip-pool-item.tsx
```diff
@@ -21,6 +22,7 @@ export function toIpPoolItem(p: SiloIpPool) {
             default
           </Badge>
         )}
+        <IpVersionBadge ipVersion={p.ipVersion} className="ml-1.5" />
```
david-crespo: nice, this works well

david-crespo: nice, this works well

AGENTS.md
```diff
@@ -117,3 +117,4 @@
 - Use `validateName` for resource names, `validateDescription` for descriptions, `validateIp`/`validateIpNet` for IPs.
 - Role helpers live in `app/api/roles.ts`.
 - Use ts-pattern exhaustive match when doing conditional logic on union types to make sure all arms are handled
+- Avoid type casts (`as`) where possible; prefer type-safe alternatives like `satisfies`, `.returnType<T>()` for ts-pattern, or `as const`
```
david-crespo, line 120: This wasn't really about anything already here in the PR — it was about what Claude did when I told it to convert something to use `ts-pattern`

charliepark: I think it's great guidance. I've had to correct a few along the way.

app/components/form/fields/IpPoolSelector.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <div className="text-secondary">
+          No IP pools available for this network interface type
+        </div>
+      ) : (
+        <ListboxField
+          name={poolFieldName}
+          items={sortedPools.map(toIpPoolItem)}
+          label={'Pool'}
+          control={control}
+          placeholder={autoSelectDefault ? 'Select a pool' : 'Use default pool'}
```
david-crespo: `Use default pool` is confusing copy here, not really clear what it means. I'm thinking that in both cases `Select a pool` is probably appropriate.

app/components/form/fields/ip-pool-item.tsx
benjaminleonard: Unimportant but I'd suggest the following. It looks identical, but is more durable to things like the `text-sans-md` line height changing and becoming vertically misaligned.

```tsx
<div className="flex flex-col gap-1">
  <div className="flex items-center gap-1.5">
    {p.name}
    {p.isDefault && <Badge color="neutral">default</Badge>}
    <IpVersionBadge ipVersion={p.ipVersion} />
  </div>
  {!!p.description && (
    <div className="text-secondary selected:text-accent-secondary">{p.description}</div>
  )}
</div>
```

app/components/form/fields/IpPoolSelector.tsx
```diff
@@ -0,0 +1,122 @@
...
+  // Note: pools are already filtered by poolType before being passed to this component
+  const sortedPools = useMemo(() => {
+    const compat = getCompatiblePools(pools, compatibleVersions)
+    // sort defaults first, sort v4 first, then name as tiebreaker
+    return R.sortBy(compat, (p) => [!p.isDefault, p.ipVersion, p.name])
+  }, [pools, compatibleVersions])
+
+  const hasNoPools = sortedPools.length === 0
+
+  // Set default pool selection on mount if none selected, or if current pool is no longer valid
```
benjaminleonard: I see this happening on the instance create but not the floating IP form. Is that because there's no `compatibleVersions` passed?

app/components/AttachEphemeralIpModal.tsx
```diff
@@ -39,39 +67,89 @@ export const AttachEphemeralIpModal = ({ onDismiss }: { onDismiss: () => void })
...
+  // Update ipVersion if only one version is compatible
+  useEffect(() => {
+    if (compatibleVersions && compatibleVersions.length === 1) {
+      form.setValue('ipVersion', compatibleVersions[0])
+    }
+  }, [compatibleVersions, form])
   const pool = form.watch('pool')
+  const ipVersion = form.watch('ipVersion')
+
+  const disabledState = useMemo(() => {
```
benjaminleonard: Probably harmless but I wonder if the `useMemo` here is actually helping.

david-crespo:
The main change across that last set of commits is getting rid of `externalIps` as live form state. Previously the form maintained the API-shaped `ExternalIpCreate[]` array and had a hard to follow web of `useEffect`s keeping it in sync with the NIC type, pool selection, and ephemeral IP checkbox.

Now the form tracks the field values only: `assignEphemeralIp` (`boolean`), `ephemeralIpPool` (`string`), and `floatingIps` (`NameOrId[]`). The `externalIps` array for the API is constructed once at submit time from those pieces. This also let me delete the `ephemeralIpVersion` form field.

I also changed `IpPoolSelector` to always mount the `ListboxField` (using `hidden` instead of conditional rendering) so the RHF validation is always registered. Before, if there were no pools, the component didn't render the listbox, which meant the `required` rule wasn't enforced, so you could submit the form with the ephemeral checkbox checked and no pool. I also switched to always sending an explicit pool selector rather than the `auto` variant since we always have a pool.

The rest is supporting test stuff and little details.

app/components/AttachEphemeralIpModal.tsx
david-crespo: Correct, changed in #3032

```ts
  const disabledReason =
    compatibleVersions.length === 0
      ? 'Instance has no network interfaces with compatible IP stacks'
      : compatibleUnicastPools.length === 0
        ? 'No compatible unicast pools available for this instance'
        : !pool
          ? 'Select a pool to continue'
          : undefined
```

app/components/form/fields/IpPoolSelector.tsx
david-crespo: This logic got nuked in #3032, the call sites do it on their own.

https://github.com/oxidecomputer/console/blob/43625b9ab860737fbc1efddf1491b21e72a17c23/app/forms/floating-ip-create.tsx#L48-L53

app/components/form/fields/ip-pool-item.tsx
david-crespo: Done in 89a9199a

app/forms/instance-create.tsx
```diff
@@ -947,16 +947,24 @@ const AdvancedAccordion = ({
...
-            >
-              Allocate and attach an ephemeral IP address
-            </Checkbox>
+            {/* TODO: Wrapping the checkbox in a <span> makes it so the tooltip
+             * shows up when you hover anywhere on the label or checkbox, not
+             * just the checkbox itself. The downside is the placement of the tooltip
+             * is a little weird (I'd like it better if it was anchored to the checkbox),
+             * but I think having it show up on label hover is worth it.
+             */}
+            <span>
```
charliepark, line 951: Clever. Agree; hover on label is much more discoverable.

PR #3007, opened 2026-01-09
Centralize "don't animate" logic for side modals

david-crespo:
Very minor, just had it on my todo list and Opus knocked it out. #2992 added a bunch of `useNavigationType` calls that I realized should be centralized in the side modal component itself


charliepark (approved):
Tried it with `reduce motion` on my Mac and all looks good!

PR #3008, opened 2026-01-13
Test coverage for e2e tests

paudmir:
I tried to clean the mess I have while I was testing different things, so I can't guarantee that this is going to work right away yet it's on the right path 😄 

I did install these packages:
`npm install v8-to-istanbul
npm install babel-plugin-Istanbul
npm install monocart` 
I have a hunch that the monocart one might not be necessary to run what is out there.

With these, one just needs to run the playwright tests as usual (just gotta point the `test` module to the one in the `fixtures.ts` so that the code coverage can run).
e.g. npm run e2ec -- action-menu

This should generate a `.nyc_output` folder with non-empty `coverage.json` file.

To generate the HTML so you can see the report, you'll need to run this command: `npx nyc report --reporter=html`

A report would look like this: 
Screenshot 2026-01-07 141333

PS. THIS IS JUST AN EXAMPLE/guide, so not intended to be deployed @david-crespo

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "update-snapshots": "npm test run -- -u",
     "e2e": "playwright test",
     "e2ec": "playwright test --project=chrome",
-    "e2e-coverage": "rm -rf .nyc_output coverage && COVERAGE=1 npm run e2ec && nyc report --reporter=html --report-dir=coverage && echo 'report at coverage/index.html'",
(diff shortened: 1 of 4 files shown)


david-crespo:
PRETTY COOL

image

david-crespo:
It all seems to work pretty well. How do you think we should consume the coverage reports @paudmir? We can generate the report in CI and make it available on any PR. We could go even farther and have CI fail if a PR causes coverage to drop. A middle ground would be to just track coverage for a while and make it easy to see the effect of a given PR on coverage.

PR #3013, opened 2026-01-21
Fix combobox query clear on page blur

### LINKED ISSUE #3012: Page losing focus discards state in dropdown
I started typing 'hel' in the Boot Disk image dropdown before switching tabs to confirm the image name I wanted. Switching back I expected to find the text I had already entered but instead it all got cleared.

https://github.com/user-attachments/assets/67298711-38d5-4874-9ec0-85655c5e0d04

Beyond just switching tabs this seems to happen if the browser loses focus (e.g., to another window or monitor). It'd be nice if we could preserve the existing input in such cases.

david-crespo:
Closes #3012 

Was curious how Claude would handle this. It cost $1.31. It did not get it until I explained that in the video, we do not see the list options close, we only see the typed-in query disappear. I've confirmed by hand that the fix does appear to work.

Claude's explanation:

> The `onClose` callback in the Combobox component unconditionally cleared the typed query. When the browser tab loses focus, Headless UI fires `onClose` (because the input blurs), which cleared the user's in-progress filter text.
> 
> The fix checks `document.hasFocus()` before clearing. When the document doesn't have focus (user switched tabs/windows), we preserve the query. When the user deliberately closes the dropdown (clicks outside, presses Escape, selects an item), the query is cleared as before.

Just putting this up to look into more later. I want to look into this behavior of Headless, calling `onClose` on blur. I don't understand why we don't see the options list close too. I'm also skeptical that adding 4 seconds worth of tests to the e2e suite is worth it for this one thing — maybe we do need something faster than playwright for component-level tests now that they're so cheap to write. We could have hundreds of them.


test/e2e/combobox-focus.e2e.ts
```diff
@@ -0,0 +1,68 @@
...
+    ) as HTMLInputElement
+    if (input) {
+      input.blur()
+    }
+
+    // Restore after a tick to allow event handlers to fire
+    setTimeout(() => {
+      document.hasFocus = originalHasFocus
+    }, 50)
+  })
```
david-crespo: This is really contrived, but it does fail without the fix in exactly the way you'd expect, and it passes with the fix.

david-crespo:
Confirmed still an issue on main, solution seems fine, just need to resolve the conflicts.

PR #3021, opened 2026-01-30
Read only disks, no size limit on local disks

### LINKED ISSUE #3009: Do not limit local disk to 1023 GiB
The current maximum local disk size validation for a maximum of 1023 GiB doesn't apply to local disks.

The way it is going to work (until we have more capacity/quota management things in place) is checking the zpool available capacity reported by inventory during instance creation. The saga fails when it can't find enough free space for creating the local disk being requested.

david-crespo:
Closes #3009

Threw this together real quick to de-risk and get a sense of the vibes. API built from https://github.com/oxidecomputer/omicron/pull/9731, which isn't merged yet and we're not sure if it will make it in for v18. This is built on top of #3005 because that has earlier API changes in it.

### Fun questions

* I put the read-only badge right in the name cell because it would be excessive to give it its own column, but really it doesn't feel great there either. Maybe if we had a read-only icon that you could hover on to get a label, it would work better in the name cell.
* `Read only: Yes` in the side modal looks silly, also not as grouped together with snapshot ID as I would like.
* Copy for the checkbox label (and possibly help text) — "Make disk read-only" is weird, as Ben points out

image

image

image

### DIFF
--- a/app/components/StateBadge.tsx
+++ b/app/components/StateBadge.tsx
@@ -91,8 +91,11 @@ export const DiskTypeBadge = (props: { diskType: DiskType; className?: string })
   </Badge>
 )
 
+// span is here to prevent it getting underlined in the LinkCell
 export const ReadOnlyBadge = () => (
-  <Badge color="neutral" className="relative">
-    Read only
-  </Badge>
+  <span>
+    <Badge color="neutral" className="relative">
+      Read only
+    </Badge>
+  </span>
 )
--- a/app/components/form/fields/DiskSizeField.tsx
+++ b/app/components/form/fields/DiskSizeField.tsx
@@ -5,11 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
-import type {
-  FieldPathByValue,
-  FieldValues,
-  ValidateResult,
-} from 'react-hook-form'
+import type { FieldPathByValue, FieldValues, ValidateResult } from 'react-hook-form'
 
 import { NumberField } from './NumberField'
 import type { TextFieldProps } from './TextField'
--- a/app/pages/project/disks/DisksPage.tsx
+++ b/app/pages/project/disks/DisksPage.tsx
@@ -29,7 +29,7 @@ import { getProjectSelector, useProjectSelector } from '~/hooks/use-params'
 import { confirmDelete } from '~/stores/confirm-delete'
 import { addToast } from '~/stores/toast'
 import { InstanceLinkCell } from '~/table/cells/InstanceLinkCell'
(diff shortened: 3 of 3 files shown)


app/components/StateBadge.tsx
```diff
@@ -90,3 +90,9 @@ export const DiskTypeBadge = (props: { diskType: DiskType; className?: string })
     {props.diskType}
   </Badge>
 )
+
+export const ReadOnlyBadge = () => (
+  <Badge color="neutral" className="relative">
+    read only
```
benjaminleonard: Such a silly nit, but I feel like even if it's made uppercase its good practice to case them properly

app/forms/disk-create.tsx
```diff
@@ -289,7 +294,16 @@ const DiskSourceField = ({
           />
         )}
 
-        {diskSource.type === 'snapshot' && <SnapshotSelectField control={control} />}
+        {diskSource.type === 'snapshot' && (
+          <>
+            <SnapshotSelectField control={control} />
+            <div className="mt-2">
+              <CheckboxField name="diskBackend.diskSource.readOnly" control={control}>
+                Make disk read-only
```
benjaminleonard, line 376: "Make disk" is strange? Maybe just "Read-only"

david-crespo: I put that in there tentatively because when it was just read-only, it wasn't clear enough that read-only refers to the resulting disk. So I think we need something more than read-only. Unless we just make the label "Read only" and we put help text in there that says "Instances won't be able to write to this disk" or something. I am also slightly torn about the hyphen. Generally I feel like hyphen is right for regular copy, but when it's just "Read-only" by itself it looks bad.

benjaminleonard:
The SSH key issue we can handle separately, will make an issue.

Read only badge in the table seems like the best of the options available, prefer to polluting the table with another column.

Read only: yes is weird. But not sure how to improve that whilst being honest about the API. If underneath we had "Read" and "Read and Write" as an access type that would feel more comfortable. Read only: True?

"Read-only" on its own is confusing when next to the listbox and moving it out of that container would probably require its own title. Again if the API was different we could use radios with the default being "Read and Write".

david-crespo:
`TRUE` and `FALSE` as badges could work?

hawkw:
> `TRUE` and `FALSE` as badges could work?

`RW` and `RO` are very horizontal-space-efficient...you could even be overly cute and do `rwx` and `r-x` like in `ls` output but that's probably a bit weird since we don't have an equivalent to the "execute" bit...

PR #3023, opened 2026-02-02
npm audit fix

david-crespo:
Should close all the dependabot PRs. react-router to 7.13.


app/util/path-builder.spec.ts
```diff
@@ -145,6 +145,7 @@ const getMatches = (pathname: string) =>
         // not used
         id: '',
         data: undefined,
+        loaderData: undefined,
```
david-crespo, line 148: RR added this field on matches, which is what we're following here.

PR #3024, opened 2026-02-03
Put routes table on router page in card block

### LINKED ISSUE #3022: Remove "routes" table title
Remove "routes" table title, we don't do this elsewhere.

Image

In a case where there isn't a tab bar to divide them, we can use a regular line divider as below.

Image

charliepark:
https://console-git-removetabletitle-oxidecomputer.vercel.app/projects/mock-project/vpcs/mock-vpc/routers/mock-system-router
https://console-git-removetabletitle-oxidecomputer.vercel.app/projects/mock-project/affinity/romulus-remus

This removes an irregular title from the page, and adds a divider line to separate the properties table at the top from the routes table below. It was the last use of the `TableControls` component, so I removed that as well. `TableTitle` still has one remaining use, on the Settings page to list `Groups`.

Screenshot 2026-02-02 at 4 48 51 PM

Ben, would love your eyes on that `mt-6` and whether it's the right gap. On pages with tabs we have a `10` level of margin, but that was looking awfully big on the left side of the screen.

Closes #3022

### DIFF
--- a/app/pages/project/affinity/AntiAffinityGroupPage.tsx
+++ b/app/pages/project/affinity/AntiAffinityGroupPage.tsx
@@ -225,7 +225,7 @@ export default function AntiAffinityPage() {
         <PropertiesTable.Row label="Members">{membersCount}</PropertiesTable.Row>
         <PropertiesTable.IdRow id={id} />
       </PropertiesTable>
-      <Divider className="my-10" />
+      <Divider className="my-8" />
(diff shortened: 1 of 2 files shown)


david-crespo:
Was surprised to find this page is basically unique in having a properties table but no tabs.

charliepark:
Screenshot 2026-02-03 at 9 35 38 AM

PR #3027, opened 2026-02-03
Update design system

benjaminleonard:
Proactive update to https://github.com/oxidecomputer/design-system/pull/156

Not a huge change but type more generally is tighter, and radii are smaller.

CleanShot 2026-02-03 at 11 56 39

Will wait until it hits main before merging.


benjaminleonard:
I didn't spot any issues, and visual regression changes are as expected. A low-risk upgrade, I think we can merge into v18.

PR #3032, opened 2026-02-04
Simplify `IpPoolSelector` and call sites

david-crespo:
`IpPoolSelector` is used on instance create, floating IP create, and ephemeral IP attach, but most of the logic inside only applied to instance create. Here I:

* Move logic that syncs the selected pool with the selected IP stack on the network interface of out `IpPoolSelector` and into instance create (the only call site that cares)
  * The IpPoolSelector props are simpler — removed `ipVersionFieldName`, `currentPool`, and `setValue`
* Move default pool logic out of `useEffect`s and into `useForm` `defaultValues`
  * In order to do this, floating IP create now prefetches pools via loader instead of fetching lazily with `useQuery`
* Floating IP create no longer hides the pool selector behind the "Advanced" accordion — it's now always visible
* Derive `ipVersion` at submit time instead of keeping track of it in the form state

I don't think any actual behavior should change here.

### DIFF
--- a/test/e2e/floating-ip-create.e2e.ts
+++ b/test/e2e/floating-ip-create.e2e.ts
@@ -18,7 +18,6 @@ test('can create a floating IP', async ({ page }) => {
     'role=heading[name*="Create floating IP"]',
     'role=textbox[name="Name"]',
     'role=textbox[name="Description"]',
-    'role=button[name="Advanced"]',
     'role=button[name="Create floating IP"]',
   ])
 
@@ -28,19 +27,8 @@ test('can create a floating IP', async ({ page }) => {
     .getByRole('textbox', { name: 'Description' })
     .fill('A description for this Floating IP')
 
-  const advancedAccordion = page.getByRole('button', { name: 'Advanced' })
-  const poolDropdown = page.getByLabel('Pool')
-
-  // accordion content should be hidden
-  await expect(poolDropdown).toBeHidden()
-
-  // open accordion
-  await advancedAccordion.click()
-
-  // pool dropdown should now be visible
-  await expect(poolDropdown).toBeVisible()
-
   // Default pool should be selected (ip-pool-1 is the v4 default)
+  const poolDropdown = page.getByLabel('Pool')
   await expect(poolDropdown).toContainText('ip-pool-1')
 
   // choose pool and submit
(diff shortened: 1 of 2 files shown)


app/components/form/fields/IpPoolSelector.tsx
```diff
@@ -59,44 +51,6 @@ export function IpPoolSelector({
...
-
-  // Update IP version when pool changes
-  useEffect(() => {
-    if (currentPool) {
-      const selectedPool = sortedPools.find((p) => p.name === currentPool)
-      if (selectedPool) {
-        setValue(ipVersionFieldName, selectedPool.ipVersion)
-      }
-    }
-  }, [currentPool, sortedPools, ipVersionFieldName, setValue])
```
david-crespo, line 98: this is not necessary at all if you derive it at submit time

app/components/form/fields/IpPoolSelector.tsx
```diff
@@ -59,44 +51,6 @@ export function IpPoolSelector({
...
-      }
-    }
-  }, [
-    currentPool,
-    sortedPools,
-    poolFieldName,
-    ipVersionFieldName,
-    setValue,
-    autoSelectDefault,
-  ])
```
david-crespo, line 88: I think this is only really necessary for instance create because that's the only one where the stack can change. We calculate it for ephemeral IP modal but we can do it once up front because it's in the context of an instance where the stack is fixed.

app/components/AttachEphemeralIpModal.tsx
```diff
@@ -136,17 +100,23 @@ export const AttachEphemeralIpModal = ({ onDismiss }: { onDismiss: () => void })
...
-              ? { poolSelector: { type: 'explicit', pool } }
-              : { poolSelector: { type: 'auto', ipVersion: effectiveIpVersion } },
+            body: {
+              poolSelector: pool
+                ? { type: 'explicit', pool }
+                : {
+                    type: 'auto',
+                    // v4 fallback here should maybe be an error instead because
+                    // it probably won't work on the API side
+                    ipVersion: hasV4Default ? 'v4' : hasV6Default ? 'v6' : 'v4',
```
david-crespo, line 117: Note `v4` fallback where when there are no defaults doesn't really make sense because auto will fail in that case. What we should do is make the user pick. The simplest thing would be to always make the user pick and just not use the auto side here.

charliepark: I believe this is now how you have it, where the form is disabled with the `No default compatible pool available; select a pool to continue` message, so I don't know that anyone is going to end up hitting this.

charliepark (approved):
So clean

app/components/AttachEphemeralIpModal.tsx
david-crespo: Ah, that's right. Insight applied here for more cleanup https://github.com/oxidecomputer/console/pull/3005/changes/72b671ab63909824eba9ddac47e9944ae2dd1cda

Issue #2932, opened 2025-10-22
Add default VPCs, subnets, etc. to mock db to reflect saga defaults


charliepark:
When creating a new project, nexus has some default resources that it creates for the user. Because these don't happen in the app layer, the mock service worker skips over them, so the new project experience in the local version of console doesn't reflect the real experience. It'd be nice if it were closer to the real thing, possibly just with our adding default values in the mock db.

More details at https://github.com/oxidecomputer/omicron/blob/main/nexus/defaults/src/lib.rs, or in the `_create` files in https://github.com/oxidecomputer/omicron/blob/main/nexus/src/app/sagas, but here are some screenshots:
<img width="722" height="226" alt="Image" src="https://github.com/user-attachments/assets/37cd170c-d789-4201-b1a1-84642b822284" />
<img width="934" height="305" alt="Image" src="https://github.com/user-attachments/assets/59d84c40-6f21-44b9-8cf0-a0baee785636" />
<img width="864" height="225" alt="Image" src="https://github.com/user-attachments/assets/65707e16-840d-4fa3-b493-455a7b26fffa" />
<img width="579" height="236" alt="Image" src="https://github.com/user-attachments/assets/2f2cb82c-fe98-41f2-8ad8-546f057dbbb1" />
<img width="1131" height="226" alt="Image" src="https://github.com/user-attachments/assets/7c962327-0e45-4023-9e07-71371c0ef54b" />
<img width="951" height="175" alt="Image" src="https://github.com/user-attachments/assets/b04f9511-70e5-432b-8d5c-745aa68b1144" />

PR #3033, opened 2026-02-05
copy Omicron's sort order for mock API items

### LINKED ISSUE #2776: Update the mock API to alphabetize paginated list queries
The API will automatically sort items by name ascending if no `sort_by` query param is included to override that. We should update the `paginated` helper in the mock API to match this behavior. [Investigate this to verify, but it's likely correct.]

- [default name or ID sort mode is name ascending](https://github.com/oxidecomputer/omicron/blob/1a1f2c6865b7a31ef223e0bd3520ca628416dd35/common/src/api/external/http_pagination.rs?plain=1#L371-L373)
- [instance list paginates by name or id](https://github.com/oxidecomputer/omicron/blob/1a1f2c6865b7a31ef223e0bd3520ca628416dd35/nexus/src/external_api/http_entrypoints.rs?plain=1#L2033-L2036)

(From [a PR](https://github.com/oxidecomputer/console/pull/2775/files#r2026177829))

charliepark:
This PR copies the Omicron API's sorting in the mock API, so Console results more accurately reflect what a user would see on an actual rack.
**Before:**
<img width="1271" height="359" alt="Screenshot 2026-02-18 at 3 06 45 PM" src="https://github.com/user-attachments/assets/a16d8638-613f-4607-965d-8bb93b121009" />

**After:**
<img width="1267" height="367" alt="Screenshot 2026-02-18 at 3 05 57 PM" src="https://github.com/user-attachments/assets/3edfd1c7-9fe6-42e7-a795-33be31b8f2bd" />


<img width="536" height="402" alt="Screenshot 2026-02-18 at 3 08 08 PM" src="https://github.com/user-attachments/assets/f3335f08-4ca5-4575-9c75-792e763a551b" />
<img width="755" height="333" alt="Screenshot 2026-02-18 at 3 08 45 PM" src="https://github.com/user-attachments/assets/2d5e11c7-6768-41b5-ad9d-b9a9df9b694e" />

This should not affect tables / lists with custom sort orders:
<img width="1260" height="407" alt="Screenshot 2026-02-18 at 3 09 01 PM" src="https://github.com/user-attachments/assets/0b1e5319-d65a-44c0-848c-47810891b0df" />

Closes #2776

---

Changes Made                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                            
1. Updated mock-api/msw/util.ts                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                          
- Added a sortItems() function that implements sorting for all sort modes:                                                                                                                                                                                                                                                                
  - name_ascending / name_descending - sorts alphabetically by name                                                                                                                                                                                                                                                                       
  - id_ascending - sorts lexicographically by ID (UUID)                                                                                                                                                                                                                                                                                   
  - time_and_id_ascending / time_and_id_descending - sorts by timestamp, then by ID
- Updated the paginated() function to:
  - Accept a sortBy parameter in the query options
  - Apply default sorting matching Omicron's behavior:
    - name_ascending for items with a name field (most common)
    - id_ascending for items without a name field
  - Sort items before paginating

2. Updated mock-api/msw/util.spec.ts

- Fixed the pagination test to expect lexicographically sorted results, matching how UUIDs are actually compared in Rust

Key Findings from Omicron Analysis

Based on the Omicron source code (https://github.com/uuid-rs/uuid):
- UUIDs in Rust use lexicographic byte-by-byte comparison because the Ord trait is derived for Uuid(Bytes), which wraps a [u8; 16] array
- Arrays in Rust compare element-by-element from left to right
- This means my implementation using localeCompare() in JavaScript correctly matches Omicron's UUID sorting behavior

The mock API now properly sorts results by default, matching Omicron's pagination behavior where:
- Most endpoints default to name_ascending (alphabetical by name)
- Endpoints without names default to id_ascending (lexicographic by UUID)
- Time-based sorting is only used when explicitly requested

Sources:
- https://docs.rs/uuid/latest/uuid/struct.Uuid.html
- https://github.com/uuid-rs/uuid
- https://github.com/oxidecomputer/omicron/blob/cf38148/common/src/api/external/http_pagination.rs


mock-api/msw/util.ts
```diff
@@ -39,42 +39,199 @@ import { getMockOxqlInstanceData } from '../oxql-metrics'
...
+          'time_created' in b
+            ? new Date(b.time_created as string | Date).valueOf()
+            : -Infinity
+        const aTime = Number.isFinite(aRaw) ? aRaw : -Infinity
+        const bTime = Number.isFinite(bRaw) ? bRaw : -Infinity
+        const timeCompare = bTime - aTime
+        return timeCompare !== 0 ? timeCompare : b.id < a.id ? -1 : b.id > a.id ? 1 : 0
+      })
+  }
+}
```
david-crespo, line 121: this would be about 1/5 as long with `R.sortBy`

app/forms/disk-create.tsx
```diff
@@ -417,7 +420,9 @@ const DiskNameFromId = ({ disk }: { disk: string }) => {
 
 const SnapshotSelectField = ({ control }: { control: Control<DiskCreateForm> }) => {
   const { project } = useProjectSelector()
-  const snapshotsQuery = useQuery(q(api.snapshotList, { query: { project } }))
+  const snapshotsQuery = useQuery(
+    q(api.snapshotList, { query: { project, limit: ALL_ISH } })
```
charliepark, line 424: Our snapshots dropdown was getting cut off; we should think about a more robust fix in a future PR for very long snapshot lists

david-crespo:
I see the `resolvePoolSelector` fix and disk create page size, but what does the sorting buy us in terms of ensuring the console’s correctness as a client? It would make more sense to me if we let you change the sort order on a table — then we’d need to implement it on the API side at least enough to exercise the client.

charliepark:
Sorry; wrote and never hit the button to submit this.

My intent with this is less that we're ensuring correctness in the client, but more that it's a representation of the saga defaults that Omicron sets up. This is tied, in part, to https://github.com/oxidecomputer/console/issues/2932, which would actually add the "default" data to the mock DB, but _this_ PR's intent is to render the mock DB's data in the console it in a similar order to how the API would return it.

david-crespo:
Fidelity to the real API logic is not necessarily a goal in itself. For me, it’s only a goal insofar as it helps us test (manually or automatically) the correctness of the client logic. That’s how we keep the mock server’s complexity manageable — we could go arbitrarily detailed in our imitation of Nexus, but we only do the minimum necessary to exercise the console. Lack of infidelity is more important to me than positive fidelity, and my worry is that this sorting logic as written is a kind of false fidelity — it's not actually very accurate to the API but it gives the impression that it should be. I might rather do nothing and be clear we're doing nothing (the status quo) than give that impression.

For example: in Nexus, the page token encodes not only the identifier for the last item in the previous page but also the sort mode and the limit. The idea is that after the first page, you don't need to pass `sort_by` or `limit` anymore — the page token is sufficient. This PR adds sort modes but does not go all the way toward encoding the relevant stuff in the token. The current composite token complicates things a lot by making the structure of the token conditional on the sort mode, but because it doesn't include the sort mode, it doesn't really add anything over using the ID as the page token. 

Going back to using the ID as the token takes a lot of the complexity away, without losing much — see #3094. With that simplification in place it will be easier to ask whether this logic makes it easier to ensure the console is correct.

charliepark:
I realized a simpler fix with the mock DB, should we even want to go through with the "adding the default resources" (which I do still think could have merit) and should we decide that we want things to be alphabetized, would be to just alphabetize the items in the JS arrays, rather than trying to parse/order the lists. Regardless, there's enough unsettled on this that I'm going to close this PR.

PR #3043, opened 2026-02-06
fix bug where disk limit warning wasn't clearing

### LINKED ISSUE #3038: Local disk: "Cannot attach more than 12 disks to instance" error message persists
After detaching all my disks, when I try to attach a new one I still get the error message saying I cannot attach more than 12. It does let me carry on with the attachment, but looks like that error message persists.

Steps to reproduce:

- Stop an instance
- Try to attach disks until you surpass the limit (12, including boot). 
-- At this point, you will get an error message saying "Cannot attach more than 12 disks to instance" and you should not be able to attach any more.
- Detach all disks until the list is empty.
- Try to attach one disk

Expected: There should not be an error message at this point, since we already removed all disks.

https://drive.google.com/file/d/1fbChtryPBzgsI0pWEzezXrzREJNTeQhL/view?usp=drive_link

charliepark:
This resets the attachDisk form when closing the modal, to prevent the "Cannot attach more than 12 disks to instance" warning from blocking action (especially when one of the disks has been detached), as @paudmir found in #3038.

Closes #3038

### DIFF
--- a/app/pages/project/instances/StorageTab.tsx
+++ b/app/pages/project/instances/StorageTab.tsx
@@ -413,6 +413,7 @@ export default function StorageTab() {
         <AttachDiskModalForm
           onDismiss={() => {
             setShowDiskAttach(false)
+            // clear API errors on the mutation
             attachDisk.reset()
           }}
           onSubmit={({ name }) => {


app/pages/project/instances/StorageTab.tsx
```diff
@@ -411,7 +411,10 @@ export default function StorageTab() {
       )}
       {showDiskAttach && (
         <AttachDiskModalForm
-          onDismiss={() => setShowDiskAttach(false)}
+          onDismiss={() => {
+            setShowDiskAttach(false)
+            attachDisk.reset()
+          }}
```
david-crespo, line 418: I'd comment `// clear API errors on the mutation so they don't persist on next modal open`

david-crespo (approved):
Not sure it's worth an entire 17 second e2e test just for this, but it's fine for now. It would be a lot faster if we were willing to set up the data behind the scenes dynamically, but it would be kind of risky. On the other hand it would mean our mock data didn't all have to be statically defined in the mock files. For this test we could just add 12 disks the instance in the database. Feels kind of risky.

charliepark:
I hear you on the test. I do think a more comprehensive suite of mock data in a `tests` project could be good, so we can have tests with static mock values and they're separate from our normal projects/instances/etc.

Am okay dropping this test, as well.

Issue #3042, opened 2026-02-06
Dual stack ephemeral IP attach on instance networking tab


david-crespo:
I have a WIP for this.

PR #3044, opened 2026-02-06
Disable ephemeral IP attach button instead of hiding

david-crespo:
The logic here is still not taking into account the possibility of dual-stack ephemeral IPs — that's #3042 — but this is a basic bit of setup for that work that it's easier to review on its own.

image

### DIFF
--- a/app/pages/project/instances/NetworkingTab.tsx
+++ b/app/pages/project/instances/NetworkingTab.tsx
@@ -517,10 +517,12 @@ export default function NetworkingTab() {
     getCoreRowModel: getCoreRowModel(),
   })
 
-  // If there's already an ephemeral IP, or if there are no network interfaces,
-  // they shouldn't be able to attach an ephemeral IP
-  const enableEphemeralAttachButton =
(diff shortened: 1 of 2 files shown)


app/pages/project/instances/NetworkingTab.tsx
```diff
@@ -517,10 +517,12 @@ export default function NetworkingTab() {
...
-  // If there's already an ephemeral IP, or if there are no network interfaces,
-  // they shouldn't be able to attach an ephemeral IP
-  const enableEphemeralAttachButton =
-    eips.items.filter((ip) => ip.kind === 'ephemeral').length === 0 && nics.length > 0
+  const ephemeralDisabledReason =
+    nics.length === 0
+      ? 'Instance has no network interfaces'
+      : eips.items.some((ip) => ip.kind === 'ephemeral')
+        ? 'Instance already has an ephemeral IP'
+        : null
```
david-crespo, line 525: We could get more sophisticated here (disabling if there are no pools, for example) but this logic is going to change anyway.

PR #3049, opened 2026-02-06
Vertically centre radio and checkbox with label

### LINKED ISSUE #3035: Checkbox label alignment slightly off
I thought it might be the `items-top` on the span containing the label and checkbox, which doesn't exist (should be `items-start`) but that doesn't fix it. Turns out the text is just slightly taller than the checkbox. A pixel of space on top of the checkbox would probably fix it but I didn't want to get into it right now.

Image

Image

benjaminleonard:
Fixes #3035

### DIFF
--- a/app/forms/image-upload.tsx
+++ b/app/forms/image-upload.tsx
@@ -107,7 +107,7 @@ function Step({ children, state, label, className }: StepProps) {
   return (
     // data-status used only for e2e testing
     <div
-      className={cn('upload-step items-top flex gap-2 px-4 py-3', className)}
+      className={cn('upload-step flex gap-2 px-4 py-3', className)}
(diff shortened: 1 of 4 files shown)


app/forms/image-upload.tsx
```diff
@@ -107,7 +107,7 @@ function Step({ children, state, label, className }: StepProps) {
   return (
     // data-status used only for e2e testing
     <div
-      className={cn('upload-step items-top flex gap-2 px-4 py-3', className)}
+      className={cn('upload-step flex gap-2 px-4 py-3', className)}
```
david-crespo, line 110: This and the side modal `items-top` weren't doing anything because `items-top` is not a thing. In both cases, I confirmed that changing it to `items-start` would have no effect and deleted the class instead.

PR #3055, opened 2026-02-06
Disk forms help text

david-crespo:
Keeping it simple and letting the docs links do most of the work.

image

image

### DIFF
--- a/app/forms/snapshot-create.tsx
+++ b/app/forms/snapshot-create.tsx
@@ -84,6 +84,7 @@ export default function SnapshotCreate() {
       <ComboboxField
         label="Disk"
         name="disk"
+        description="Only disks that can be snapshotted are listed"
         placeholder="Select a disk"
         items={diskItemsForCombobox}
         required


david-crespo:
This PR only does 1 and 5 here (table generated by LLMs as part of a report on places we could stand to clarify local and read-only disks — take priorities with a grain of salt, but they're not bad). I'm inclined not to do any other ones for now. Not the tooltips in the tables, especially not on the badges. I feel like it would be very busy if there were tooltip icons everywhere or every time you moved your mouse over a badge you got a tooltip. I think the form is really the place where it matters because that's where you're making a decision.

| Priority | Location | What to do |
|----------|----------|------------|
| **High** | Disk create: type radio (1a) | Descriptions on radio options |
| **High** | Disk create: read-only checkbox (1d) | "Permanent, cannot be reversed" |
| **High** | Instance create: read-only checkbox (2a) | Same as 1d |
| **High** | Attach disk modal (5) | Show type/read-only in combobox items |
| **High** | Snapshot create (6) | Explain why local disks are absent |
| **Medium** | Disk create: local selected, no source (1b) | Note explaining blank-only |
| **Medium** | Disks page DocsPopover (3a) | Mention disk types in summary |
| **Medium** | ReadOnlyBadge tooltip (7b) | Tooltip explaining permanence |
| **Medium** | DiskTypeBadge tooltip (7a) | Brief type explanation |
| **Low** | Disk detail rows (4a, 4b) | Covered by badge tooltips |
| **Low** | Local disk size note (1c) | Explain sled-capacity limit |
| — | Boot disk always distributed (2b) | No change needed |
| — | Snapshot disabled reason | Already clear |
| — | API utils (canSnapshot) | Already correct |

PR #3057, opened 2026-02-07
Update instance-create to handle both v4 and v6 ephemeral IP options

### LINKED ISSUE #3041: Instance create: allow creating ephemeral IPs of both versions
Only if there are pools of both versions available, I think.

charliepark:
Instances can now have both IPv4- and IPv6-backed ephemeral IPs. This updates the instance create flow to account for that.

It checks / unchecks the ephemeral IP box for the IP version(s) specified in the Network Interfaces section of the form.

**Default NICs**
Screenshot 2026-02-06 at 4 47 32 PM
Screenshot 2026-02-06 at 4 47 49 PM
Screenshot 2026-02-06 at 4 47 56 PM

**Custom NICs**
Screenshot 2026-02-06 at 4 48 16 PM
Screenshot 2026-02-06 at 4 48 34 PM
Screenshot 2026-02-06 at 4 48 53 PM

**No NIC**
Screenshot 2026-02-06 at 4 49 01 PM

Closes #3041

### DIFF
--- a/test/e2e/instance-create.e2e.ts
+++ b/test/e2e/instance-create.e2e.ts
@@ -893,6 +893,20 @@ test('create instance with custom IPv4-only NIC constrains ephemeral IP to IPv4'
   // Open dropdown to check available options - only IPv4 pools should appear
   await v4PoolDropdown.click()
   await expect(page.getByRole('option', { name: 'ip-pool-1' })).toBeVisible()
+  // Close dropdown to avoid obscuring subsequent interactions
+  await page.keyboard.press('Escape')
+
+  // Create the instance
+  await page.getByRole('button', { name: 'Create instance' }).click()
+  await expect(page).toHaveURL(`/projects/mock-project/instances/${instanceName}/storage`)
+
+  // Verify exactly one ephemeral IP row exists, and it is IPv4
+  await page.getByRole('tab', { name: 'Networking' }).click()
+  const externalIpsTable = page.getByRole('table', { name: /external ips/i })
+  const ephemeralRows = externalIpsTable.getByRole('row').filter({ hasText: 'ephemeral' })
+  await expect(ephemeralRows).toHaveCount(1)
+  await expect(externalIpsTable.getByText('v4')).toBeVisible()
+  await expect(externalIpsTable.getByText('v6')).toBeHidden()
 })
 
 test('create instance with custom IPv6-only NIC constrains ephemeral IP to IPv6', async ({
@@ -955,6 +969,20 @@ test('create instance with custom IPv6-only NIC constrains ephemeral IP to IPv6'
   // Open dropdown to check available options - only IPv6 pools should appear
   await v6PoolDropdown.click()
   await expect(page.getByRole('option', { name: 'ip-pool-2' })).toBeVisible()
+  // Close dropdown to avoid obscuring subsequent interactions
+  await page.keyboard.press('Escape')
+
+  // Create the instance
+  await page.getByRole('button', { name: 'Create instance' }).click()
+  await expect(page).toHaveURL(`/projects/mock-project/instances/${instanceName}/storage`)
+
+  // Verify exactly one ephemeral IP row exists, and it is IPv6
+  await page.getByRole('tab', { name: 'Networking' }).click()
+  const externalIpsTable = page.getByRole('table', { name: /external ips/i })
+  const ephemeralRows = externalIpsTable.getByRole('row').filter({ hasText: 'ephemeral' })
+  await expect(ephemeralRows).toHaveCount(1)
+  await expect(externalIpsTable.getByText('v4')).toBeHidden()
+  await expect(externalIpsTable.getByText('v6')).toBeVisible()
 })
 
 test('create instance with custom dual-stack NIC allows both IPv4 and IPv6 ephemeral IPs', async ({
--- a/test/e2e/ip-pool-silo-config.e2e.ts
+++ b/test/e2e/ip-pool-silo-config.e2e.ts
@@ -50,6 +50,9 @@ test.describe('IP pool configuration: myriad silo (v4-only default)', () => {
     // v6 checkbox should be visible but not checked (no v6 default in myriad silo)
     await expect(v6Checkbox).toBeVisible()
     await expect(v6Checkbox).not.toBeChecked()
+    await expect(v6Checkbox).toBeDisabled()
+    await v6Checkbox.hover()
+    await expect(page.getByText('No IPv6 pools available')).toBeVisible()
 
     // IPv4 pool dropdown should be visible with default pool preselected
     const v4PoolDropdown = page.getByLabel('IPv4 pool')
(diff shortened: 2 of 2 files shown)


app/forms/instance-create.tsx
```diff
@@ -841,11 +841,9 @@ const NetworkingSection = ({
...
   }, [attachableFloatingIps, attachedFloatingIps, compatibleVersions])
 
-  const attachedFloatingIpsData = R.compact(
-    attachedFloatingIps.map((floatingIp) =>
-      attachableFloatingIps.find((fip) => fip.name === floatingIp)
-    )
-  )
+  const attachedFloatingIpsData = attachedFloatingIps
+    .map((floatingIp) => attachableFloatingIps.find((fip) => fip.name === floatingIp))
+    .filter((ip): ip is FloatingIp => !!ip)
```
david-crespo: seems like compact is a lodash thing. the docs say to do exactly what you did instead

https://remedajs.com/migrate/lodash/#compact

app/forms/instance-create.tsx
```diff
@@ -438,22 +443,38 @@ export default function CreateInstanceForm() {
...
+          // API requirements (dual-stack requests MUST specify ipVersion on each)
+          const externalIps: ExternalIpCreate[] = [
+            // v4 ephemeral if enabled (order: v4 before v6)
+            ...(values.ephemeralIpv4
+              ? [
+                  {
+                    type: 'ephemeral' as const,
+                    poolSelector: values.ephemeralIpv4Pool
+                      ? { type: 'explicit' as const, pool: values.ephemeralIpv4Pool }
+                      : { type: 'auto' as const, ipVersion: 'v4' as const },
```
david-crespo: I don't think we need the auto case in these because we always have a pool — see the comment in the deleted code

app/forms/instance-create.tsx
```diff
@@ -811,88 +846,104 @@ const NetworkingSection = ({
...
-    .filter((ip) => !!ip)
-
-  // Filter unicast pools by compatible IP versions
-  // unicastPools is already sorted (defaults first, v4 first, then by name),
-  // so filtering preserves that order
-  const compatiblePools = useMemo(
-    () => unicastPools.filter(poolHasIpVersion(compatibleVersions)),
-    [unicastPools, compatibleVersions]
-  )
+    .filter((ip): ip is FloatingIp => !!ip)
```
david-crespo: Type guard shouldn't be necessary as far as I know — is it?

app/forms/instance-create.tsx
```diff
@@ -922,6 +973,54 @@ const NetworkingSection = ({
...
+              required={false}
+              hideOptionalTag
+              label={`${displayVersion} pool`}
+            />
+          </div>
+        )}
+      </div>
+    )
+  }
+
```
david-crespo: Don't define components in render. Probably worth putting in the CLAUDE.md. There's an eslint rule for this, [`react/no-unstable-nested-components`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md), but oxlint doesn't have it yet.

app/forms/instance-create.tsx
```diff
@@ -811,88 +846,104 @@ const NetworkingSection = ({
...
+      to attach an ephemeral IP{version} address
+    </>
+  )
+  const noPoolMessage = (version: IpVersion) => (
+    <>
+      No IP{version} pools available
+      <br />
+      for this instance’s network interfaces
+    </>
+  )
```
david-crespo: These and `getDisabledReason` should be defined at top level probably. They're pure functions.

app/forms/instance-create.tsx
```diff
@@ -922,6 +973,54 @@ const NetworkingSection = ({
...
+          </span>
+        </Wrap>
+        {checked && (
+          <div className="ml-6">
+            <IpPoolSelector
+              control={control}
+              poolFieldName={poolFieldName}
+              pools={pools}
+              disabled={isSubmitting}
+              required={false}
```
david-crespo: In the previous version this wasn't conditionally rendered — I used a class to hide it while keeping it mounted so that I could have the `required` prop change its value on the form, enforcing required when the box was checked but not when it was unchecked. Hard coding `required: false` avoids that problem because the requirement never changes, but it only works if we can guarantee that there is always a pool selected when we expect there to be.

```ts
const defaultV4Pool = compatibleDefaultPools.find((p) => p.ipVersion === 'v4')
const defaultV6Pool = compatibleDefaultPools.find((p) => p.ipVersion === 'v6')
...

    ephemeralIpv4: !!defaultV4Pool && defaultCompatibleVersions.includes('v4'),
    ephemeralIpv4Pool: defaultV4Pool?.name || '',
    ephemeralIpv6: !!defaultV6Pool && defaultCompatibleVersions.includes('v6'),
    ephemeralIpv6Pool: defaultV6Pool?.name || '',

```

This logic means that if there is a v6 pool but there is no default v6 pool, then it's possible to check the IPv6 ephemeral IP box but not select a pool. I was able to confirm this by hand with the `pelerines` silo, which has a v6 pool but no default. It lets me check the box and not pick a pool, and then I get an error from the API.

But if instead you always render `IpPoolSelector`, hide it when the checkbox is unchecked, and have `required={checked}`, you solve this problem with form validation — it's a form error to not have a pool picked. In hindsight I should have left a comment on `required={assignEphemeralIp}` and the hidden class thing explaining what it was for.

app/forms/instance-create.tsx
charliepark: The flow to end up in a state where it's actually useful is pretty rare …

  1. User selects "ip-1" from modal → added to attachedFloatingIps (form state)
  2. React Query refetches floatingIpList (on window focus, reconnect, or stale time expiry)
  3. "ip-1" is now attached to another instance or deleted (via some other user on the system)
  4. "ip-1" filtered out of attachableFloatingIps
  5. Form state still has "ip-1", but .find() can't find it

But TS complains* unless it has either a filter (`.filter((ip) => ip !== undefined)` or `.filter((ip) => !!ip)`), or an `!` assertion `(floatingIp) => attachableFloatingIps.find((fip) => fip.name === floatingIp)!`

* The complaint is in the MiniTable, where an `item` might be undefined.

I'm moving to an assertion as that should throw an error if the user ends up in that state, versus having the Floating IP silently disappear, but I'm open to filtering if that feels like a better situation.

app/forms/instance-create.tsx
david-crespo: Fixed in 503fac634660b9c674b263cd88fa0408b6844be5 and added a test

david-crespo:
With the refactors, this came in at barely a net add in app code — all the additions are tests, which is great!

```
$ jj diff -f 'fork_point(trunk() | @)' --stat app
app/components/form/fields/IpPoolSelector.tsx |  11 +-
app/components/form/fields/ListboxField.tsx   |   4 +
app/forms/instance-create.tsx                 | 337 ++++++++++++++--------------
app/ui/lib/Listbox.tsx                        |   8 +-
4 files changed, 194 insertions(+), 166 deletions(-)
```

PR #3059, opened 2026-02-07
Fix firewall rule test flake in Firefox

david-crespo:
View [https://github.com/oxidecomputer/console/pull/3059/changes?w=1](https://github.com/oxidecomputer/console/pull/3059/changes?w=1) with whitespace changes hidden — there's an indentation change that makes a 5 line change look like 150.

Got Claude on this oldy but goody and I think we really got somewhere. I was able to repro the flake pretty consistently with `--repeat-each 30` (it would fail once or twice) and then confirm that that stopped happening with this fix in place. When manually testing an initial version of the fix, I found a key behavior that it is useful to e2e test, so I added that test, and was able to confirm both manually and automatically that the solution here doesn't have this problem. I also improved the comments around the `onEnter` handler modified here to explain how it works overall and why this particular approach is necessary to avoid the flake.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,16 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Edit|Write|NotebookEdit",
+        "hooks": [
+          {
+            "type": "command",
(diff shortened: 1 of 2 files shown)


david-crespo:
BAHAHAHAHA no of course I didn't run the tests locally in Safari.

image

PR #3066, opened 2026-02-10
tools: oxfmt

david-crespo:
`oxfmt` finally [fixed](https://github.com/oxc-project/oxc/pull/19000) the last thing I was waiting for: getting the Tailwind class order the same as the official prettier plugin. So we're ready to go. As you can see, oxfmt takes as long to format the entire repo as prettier takes for a single file.

| | Single file | Whole repo (475 files) |
|---|---|---|
| **oxfmt (npx)** | 287.5ms ± 3.9ms (10 runs) | 446.1ms ± 11.2ms (10 runs) |
| **prettier (npx)** | 405.2ms ± 3.4ms (10 runs) | 3.213s ± 0.037s (10 runs) |
| | **oxfmt 1.4x faster** | **oxfmt 7.2x faster** |
| **oxfmt (direct)** | 157.6ms ± 1.9ms (18 runs) | 332.4ms ± 48.9ms (10 runs) |
| **prettier (direct)** | 283.0ms ± 1.5ms (10 runs) | 3.107s ± 0.061s (10 runs) |
| | **oxfmt 1.8x faster** | **oxfmt 9.4x faster** |

PR #3067, opened 2026-02-10
trivial: oxfmt gen api

david-crespo:
Running the generator with the new formatter produces different results from what I got in #3066 because the generator produces unformatted code, which formats differently than when you run the formatter on already-formatted code. There are some janky looking changes in here, but since it's in the generated code, I don't care as long as it all still works.


david-crespo:
Ok, this is a very good bug. CI is failing because you need to run `npm run fmt` twice to get  a stable result that won't change again. I'm going to put in a second `npm run fmt` in the script for now and report the bug to oxfmt.

PR #3075, opened 2026-02-13
Use tsgo, upgrade oxlint

david-crespo:
https://devblogs.microsoft.com/typescript/typescript-native-port/
https://github.com/microsoft/typescript-go

We still have to keep the regular typescript dep installed because some libraries depend on it, but we can use `tsgo` for all our actual typechecking. I changed the `tsc` script to call `tsgo` since eventually `tsgo` is going to be released as `tsc` (I think).

| Scenario | tsc | tsgo | Speedup |
|---|---|---|---|
| **Cold** (no incremental cache) | 5.46s | 1.54s | **3.5x** |
| **Incremental** (one file touched) | 1.02s | 280ms | **3.7x** |
| **Incremental no-op** (nothing changed) | 995ms | 278ms | **3.6x** |

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -60,6 +60,24 @@
     "radix": "error",
 
     "@typescript-eslint/no-explicit-any": "error",
+    "@typescript-eslint/consistent-type-assertions": "error",
+    "prefer-const": "error",
+    "import/no-nodejs-modules": "error",
+    "react/no-this-in-sfc": "error",
+    "@typescript-eslint/no-invalid-void-type": "error",
+    "unicorn/relative-url-style": "error",
+
+    // suspicious rules (cherry-picked)
+    "no-extra-bind": "error",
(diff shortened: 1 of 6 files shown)


app/forms/firewall-rules-create.tsx
```diff
@@ -81,7 +81,8 @@ export default function CreateFirewallRuleForm() {
 
   const updateRules = useApiMutation(api.vpcFirewallRulesUpdate, {
     onSuccess(updatedRules) {
-      const newRule = updatedRules.rules[updatedRules.rules.length - 1]
+      // We just appended a rule, so the response list is non-empty
+      const newRule = updatedRules.rules.at(-1)!
```
david-crespo, line 85: this changes nothing, it just came up when I did an audit of things that fail when we turn on `noUncheckedIndexAccess` and I changed it to use `at` because it's shorter

Issue #2264, opened 2024-05-30
Light mode


david-crespo:
We’ll have to do it eventually!

Issue #2183, opened 2024-04-23
Tracking issue for removing `style-src: 'unsafe-inline'` from CSP


iliana:
placeholder one sec

PR #3089, opened 2026-02-24
Replace Radix with Base UI

david-crespo:
#2264 introduced Base UI for its dropdown menu stuff for the theme picker. Base UI is almost a drop-in replacement for Radix, made by some of the same people. So converting our existing Radix usage was pretty easy. I also got to delete `AccordionItem` — guess we weren't using it anymore after we got rid of the advanced sections on floating IP create and instance create.

### DIFF
--- a/app/main.tsx
+++ b/app/main.tsx
@@ -5,6 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
+import { CSPProvider } from '@base-ui/react/csp-provider'
 import { QueryClientProvider } from '@tanstack/react-query'
 import { LazyMotion, MotionConfig } from 'motion/react'
 // import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
@@ -46,18 +47,20 @@ function render() {
 
   root.render(
     <StrictMode>
-      <QueryClientProvider client={queryClient}>
-        <LazyMotion strict features={loadFeatures}>
-          <MotionConfig reducedMotion="user">
-            <ErrorBoundary>
-              <ConfirmActionModal />
-              <SkipLink id="skip-nav" />
-              <RouterProvider router={router} />
-            </ErrorBoundary>
-          </MotionConfig>
-        </LazyMotion>
-        {/* <ReactQueryDevtools initialIsOpen={false} /> */}
-      </QueryClientProvider>
+      <CSPProvider disableStyleElements>
+        <QueryClientProvider client={queryClient}>
+          <LazyMotion strict features={loadFeatures}>
+            <MotionConfig reducedMotion="user">
+              <ErrorBoundary>
+                <ConfirmActionModal />
+                <SkipLink id="skip-nav" />
+                <RouterProvider router={router} />
+              </ErrorBoundary>
+            </MotionConfig>
+          </LazyMotion>
+          {/* <ReactQueryDevtools initialIsOpen={false} /> */}
+        </QueryClientProvider>
+      </CSPProvider>
     </StrictMode>
   )
 }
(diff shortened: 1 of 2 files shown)


vite.config.ts
```diff
@@ -128,11 +126,10 @@ export default defineConfig(({ mode }) => ({
...
     tsconfigPaths(),
-    createHtmlPlugin({
-      inject: {
-        tags: process.env.VERCEL ? [previewAnalyticsTag, ...previewMetaTag] : [],
-      },
-    }),
+    {
+      name: 'inject-html-tags',
+      transformIndexHtml: () => (process.env.VERCEL ? previewTags : []),
+    },
```
david-crespo, line 132: Claude claims this worked natively without `createHtmlPlugin` all along, and the only thing the plugin was doing was adding minification because we weren't using any of its actual features. Oh well. Getting rid of minification is fine because it does nothing for the tiny HTML file besides making it hard to read.

patches/react-remove-scroll+2.6.3.patch
```diff
@@ -1,57 +0,0 @@
...
-             return !lastProps.current.allowPinchZoom;
-@@ -139,10 +121,7 @@ export function RemoveScrollSideCar(props) {
-             document.removeEventListener('touchstart', scrollTouchStart, nonPassive);
-         };
-     }, []);
--    var removeScrollBar = props.removeScrollBar, inert = props.inert;
--    return (React.createElement(React.Fragment, null,
--        inert ? React.createElement(Style, { styles: generateStyle(id) }) : null,
--        removeScrollBar ? React.createElement(RemoveScrollBar, { gapMode: props.gapMode }) : null));
-+    return (React.createElement(React.Fragment, null));
```
david-crespo, line 54: The analogous thing in Base UI is `<CSPProvider disableStyleElements>`, which is great.

app/main.tsx
```diff
@@ -46,18 +47,20 @@ function render() {
...
-            <ErrorBoundary>
-              <ConfirmActionModal />
-              <SkipLink id="skip-nav" />
-              <RouterProvider router={router} />
-            </ErrorBoundary>
-          </MotionConfig>
-        </LazyMotion>
-        {/* <ReactQueryDevtools initialIsOpen={false} /> */}
-      </QueryClientProvider>
+      <CSPProvider disableStyleElements>
```
david-crespo, line 50: Base UI's `ScrollArea` and `Select` components inject an inline `<style>` element to hide native scrollbars (the class `base-ui-disable-scrollbar` with `scrollbar-width: none`). This would be a problem if we ever tighten our CSP to remove `'unsafe-inline'` from `style-src` ([#2183](https://github.com/oxidecomputer/console/issues/2183)).

Unlike Radix, which we had to patch (`react-remove-scroll` and `react-use-escape-keydown`), Base UI has a first-class solution: wrapping the app in `<CSPProvider disableStyleElements>` suppresses all inline `<style>` injection. We then provide the equivalent CSS in our own stylesheet (see `index.css`).

We're not using `ScrollArea` or `Select` yet, but adding this now means we won't accidentally introduce a CSP violation when we do.

To verify this works, I temporarily added a `ScrollArea` to the projects page and confirmed:
- With `disableStyleElements`, no `<style>` elements containing `base-ui-disable-scrollbar` appear in the DOM
- The class is still applied to elements and picks up its rules from our stylesheet
- Native scrollbars are hidden as expected (verified by toggling "Show scroll bars: Always" in macOS settings and commenting out the CSS to see the scrollbar reappear)

david-crespo: cc @iliana we're switching to a UI library that doesn't require that hack to disable style tag injection :)

PR #3095, opened 2026-02-26
Add Fleet Access page

### LINKED ISSUE #2916: System-level access and IAM page
`GET/PUT /v1/system/policy` handle fleet roles. It just occurred to me that it's weird that we don't have a page analogous to Silo Access on the System side.

Image

charliepark:
This adds a system-level access page, with a form for setting fleet-level permissions.

Screenshot 2026-02-26 at 4 59 26 AM

One enhancement we might consider: It looks like there are a few booleans — `silo_admin` and `fleet_viewer` — on the CurrentUser object coming from Omicron, but there is not a `fleet_admin` attribute. If we add that in Omicron, we could disable the "Add User or Group" button and other controls on the System Access page for people without a fleet_admin role.

Closes #2916

### DIFF
--- a/app/api/roles.ts
+++ b/app/api/roles.ts
@@ -18,8 +18,8 @@ import type { FleetRole, IdentityType, ProjectRole, SiloRole } from './__generat
 import { api, q, usePrefetchedQuery } from './client'
 
 /**
- * Union of all the specific roles, which are all the same, which makes making
- * our methods generic on the *Role type is pointless (until they stop being the same).
+ * Union of all the specific roles, which used to all be the same until we added
+ * limited collaborator to silo.
  */
 export type RoleKey = FleetRole | SiloRole | ProjectRole
 
@@ -120,6 +120,10 @@ export function useUserRows<Role extends RoleKey = RoleKey>(
     return roleAssignments.map((ra) => ({
       id: ra.identityId,
       identityType: ra.identityType,
+      // A user might not appear here if they are not in the current user's
+      // silo. This could happen in a fleet policy, which might have users from
+      // different silos. Hence the ID fallback. The code that displays this
+      // detects when we've fallen back and includes an explanatory tooltip.
       name: usersDict[ra.identityId]?.displayName || ra.identityId,
       roleName: ra.roleName,
       roleSource,
--- a/app/pages/system/silos/SiloFleetRolesTab.tsx
+++ b/app/pages/system/silos/SiloFleetRolesTab.tsx
@@ -22,6 +22,8 @@ export default function SiloFleetRolesTab() {
   const siloSelector = useSiloSelector()
   const { data: silo } = usePrefetchedQuery(siloView(siloSelector))
 
+  // `mappedFleetRoles` is keyed by silo role, with each value listing the
+  // fleet roles granted to actors who have that silo role.
   const roleMapPairs: [string, string][] = Object.entries(silo.mappedFleetRoles).flatMap(
     ([siloRole, fleetRoles]) => fleetRoles.map((fleetRole) => [siloRole, fleetRole])
   )


david-crespo:
Looks really good. Nice message in the form. Will review properly tomorrow.

mock-api/msw/handlers.ts
```diff
@@ -2319,7 +2319,25 @@ export const handlers = makeHandlers({
   supportBundleUpdate: NotImplemented,
   supportBundleView: NotImplemented,
   switchView: NotImplemented,
-  systemPolicyUpdate: NotImplemented,
+  systemPolicyUpdate({ body, cookies }) {
+    requireFleetAdmin(cookies)
+
+    const newAssignments = body.role_assignments
+      .filter((r) => fleetRoles.some((role) => role === r.role_name))
```
david-crespo, line 2355: `(r) => fleetRoles.includes(r.role_name)` I think?

app/layouts/SystemLayout.tsx
```diff
@@ -101,6 +103,9 @@ export default function SystemLayout() {
           <NavLinkItem to={pb.systemUpdate()}>
             <SoftwareUpdate16Icon /> System Update
           </NavLinkItem>
+          <NavLinkItem to={pb.systemAccess()}>
+            <Access16Icon /> System Access
+          </NavLinkItem>
```
david-crespo, line 108:
```
System Update
System Access
```

doesn't feel right to me, though in both cases the qualifier is useful. On the other hand we are in the system section. Could we just call them Update and Access? @benjaminleonard @askfongjojo curious what you think. 

Image

What about `Fleet Access`. Slightly uglier but more literal is often better. System Update + Fleet Access might feel less repetitive.

Image

Note the other ones are called Silo Access and Project Access because otherwise they would be incomprehensible. So I think we probably do need a qualifier.

Image

Image

app/pages/system/FleetAccessPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+    <>
+      <PageHeader>
+        <PageTitle icon={<Access24Icon />}>System Access</PageTitle>
+        <DocsPopover
+          heading="access"
+          icon={<Access16Icon />}
+          summary="Roles determine who can view, edit, or administer this fleet."
+          links={[docLinks.keyConceptsIam, docLinks.access]}
+        />
+      </PageHeader>
```
david-crespo, line 257: Users can get fleet roles implicitly from a silo role if the fleet role mapping is set. I wonder if it's worth putting in some kind of info message on this page saying that the explicitly assigned roles don't necessarily exist exhaust the set of users who have permissions on the fleet, and you need to look at mapped fleet roles in the silos to be sure.

I guess ideally we'd actually be able to list those mappings on this page, but it's kind of gnarly because I think any silo could contribute? We could do something really cute and let the user pop a modal where we fetch all the silos and their mapping and list the ones that have mappings defined. Kind of elaborate but not really very hard. The only place it would be a problem is the colo rack, which has a million silos.

app/pages/system/SystemAccessPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+import { identityTypeLabel, roleColor } from '~/util/access'
+import { groupBy } from '~/util/array'
+import { docLinks } from '~/util/links'
+
+const EmptyState = ({ onClick }: { onClick: () => void }) => (
+  <TableEmptyBox>
+    <EmptyMessage
+      icon={<Access24Icon />}
+      title="No authorized users"
+      body="Give permission to view, edit, or administer this fleet"
```
david-crespo: Might also be worth mentioning here the fleet role mapping thing — that this list being empty doesn't necessarily mean nobody can see fleet stuff. Necessarily so 😁 because anyone who can see this page has to be a fleet viewer.

app/layouts/SystemLayout.tsx
charliepark: `Fleet Access` works for me. Agree that the stacked `System` felt a bit off.

david-crespo: Yeah, let's go with that for now. It's kind of a problem for us because fleet and system are synonymous, but I'm ok with using the term that we use in the access subsystem.

AGENTS.md
```diff
@@ -24,7 +24,7 @@
 
 - Run local checks before sending PRs: `npm run lint`, `npm run tsc`, `npm test run`, and `npm run e2ec`.
 - You don't usually need to run all the e2e tests, so try to filter by file and tes t name like `npm run e2ec -- instance -g 'boot disk'`. CI will run the full set.
-- Keep Playwright specs focused on user-visible behavior—use accessible locators (`getByRole`, `getByLabel`), the helpers in `test/e2e/utils.ts` (`expectToast`, `expectRowVisible`, `selectOption`, `clickRowAction`), and close toasts so follow-on assertions aren’t blocked.
+- Keep Playwright specs focused on user-visible behavior—use accessible locators (`getByRole`, `getByLabel`), the helpers in `test/e2e/utils.ts` (`expectToast`, `expectRowVisible`, `selectOption`, `clickRowAction`), and close toasts so follow-on assertions aren’t blocked. Avoid Playwright’s legacy string selector syntax like `page.click(‘role=button[name="..."]’)`; prefer `page.getByRole(‘button’, { name: ‘...’ }).click()` and friends.
```
charliepark, line 27: Nice

mock-api/msw/handlers.ts
charliepark: Just recording what we chatted about in Matrix … the `includes` is definitely cleaner, but runs into a TS issue where the `role_name` is a string and so TS doesn't know how to compare it with the items in `fleetRoles` (each one a `FleetRole`) unless we add a cast, as in `(r) => (fleetRoles as string[]).includes(r.role_name)`.

In order to avoid that casting, we'll just stick with the existing `some` approach.

PR #3101, opened 2026-02-28
Dual-stack ephemeral IPs on instance detail

### LINKED ISSUE #3042: Dual stack ephemeral IP attach on instance networking tab

david-crespo:
Closes #3042 

https://github.com/user-attachments/assets/ffb4f4e3-6010-4396-8803-6cd03fe11f13

### DIFF
--- a/app/components/AttachEphemeralIpModal.tsx
+++ b/app/components/AttachEphemeralIpModal.tsx
@@ -29,11 +29,13 @@ import { ALL_ISH } from '~/util/consts'
 
 type AttachEphemeralIpModalProps = {
   availableVersions: IpVersion[]
+  infoMessage: string | null
   onDismiss: () => void
 }
 
 export const AttachEphemeralIpModal = ({
   availableVersions,
+  infoMessage,
   onDismiss,
 }: AttachEphemeralIpModalProps) => {
   const { project, instance } = useInstanceSelector()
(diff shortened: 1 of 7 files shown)


app/components/AttachEphemeralIpModal.tsx
```diff
@@ -73,25 +70,31 @@ export const AttachEphemeralIpModal = ({ onDismiss }: { onDismiss: () => void })
...
+      : !pool
+        ? 'Select a pool to continue'
+        : undefined
+
+  const message =
+    availableVersions.length === 1
+      ? `Only ${availableVersions[0]} pools are shown because this instance already has a ${availableVersions[0] === 'v4' ? 'v6' : 'v4'} ephemeral IP.`
+      : availableVersions.length === 2
+        ? 'Dual-stack network interfaces support one ephemeral IP per version.'
+        : undefined
```
david-crespo: Good catch from Señor gpt-5.3-codex: this message is only correct for dual-stack instances and, more subtly, it assumes there are pools available for both stacks. Need to make it conditional on the stack config and probably the pools.

> The new info banner can state a false reason for pool filtering. In [app/components/AttachEphemeralIpModal.tsx:79](/Users/david/oxide/console/app/components/AttachEphemeralIpModal.tsx:79), the `availableVersions.length === 1` branch always says the instance already has the opposite-version ephemeral IP. 
> 
> That is not always true:
> 
> - single-stack instances with no ephemeral IP also produce `availableVersions.length === 1`
> - dual-stack instances with one open version because of pool availability (not because the other version is attached) can also produce `length === 1`
> 
> This creates incorrect user-facing guidance in the modal.
> 
> Open question: should that message only appear when the instance is dual-stack and exactly one ephemeral IP is already attached? If yes, the condition needs attached-version context rather than only `availableVersions.length`.

PR #3104, opened 2026-03-04
add created_at and updated_at to access page user lists

charliepark:
This builds off of an update to the API in https://github.com/oxidecomputer/omicron/pull/9958, and adds the created and updated timestamps on the various Access pages.

Screenshot 2026-03-04 at 3 29 26 PM

https://github.com/oxidecomputer/console/pull/3095 does not include these columns, but those will be easy to add once https://github.com/oxidecomputer/omicron/pull/9958 has been merged.


david-crespo:
This doesn't make sense to me as the place to display this info. I would expect to see this on the user and group list views (which we don't have yet).

Fundamentally this table is a list of role assignments, while these timestamps are created and update on the user/group. If these had timestamps I would expect them to be when the roles were assigned. We don't have that, and it wouldn't be very useful anyway since you have to PUT the entire set of assignments every time, which means every assignment int he list would always have the same timestamp.

PR #3116, opened 2026-03-07
Light mode polish

benjaminleonard:
WIP


david-crespo:
New less-saturated blue LGTM

PR #3122, opened 2026-03-12
Move theme arrow to left of dropdown

charliepark:
This moves the Theme arrow from the right-hand side of the row to the left-hand side of the row, so that it points to the modal extension menu.
Screenshot 2026-03-12 at 3 53 21 PM

### DIFF
--- a/app/components/TopBar.tsx
+++ b/app/components/TopBar.tsx
@@ -10,7 +10,6 @@ import { Link } from 'react-router'
 
 import { api, navToLogin, useApiMutation } from '@oxide/api'
 import {
-  DirectionLeftIcon,
   Monitor12Icon,
   Moon12Icon,
   Organization16Icon,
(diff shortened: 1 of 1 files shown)


app/components/TopBar.tsx
```diff
@@ -138,7 +138,7 @@ function UserMenu() {
         <div
           className={cn(
             buttonStyle({ size: 'sm', variant: 'ghost' }),
-            'flex items-center gap-1.5 px-2!'
+            'flex items-center gap-1.5 pl-2! pr-[10px]!'
```
charliepark: Moves the padding on the right of the user's name from 8px to 10px so the name and the menu options below ("Settings", "Theme", "Sign out") line up with the left-hand edge of the name's text.

david-crespo (approved):

PR #3129, opened 2026-03-16
Improve "Jump To" (cmd-k) dialog

### LINKED ISSUE #2572: Jump To modal only allows jumping to first 25 projects
In the Oxide Console pressing <kbd>Ctrl</kbd> + <kbd>k</kbd> (Linux/Windows) or <kbd>Cmd</kbd> + <kbd>k</kbd> (macOS) opens the **Jump To** modal which allows operators to quickly perform actions such as jumping to a project or silo.

If an operator has more than 25 projects in their silo then they will only be able to jump to one of the first 25 projects in lexicographical order. For example, if a silo has 50 projects named `foo01` to `foo50` then they can only jump to projects `foo01` to `foo25`.

An operator should be able to jump to any project within their silo via the **Jump To** modal.

### LINKED ISSUE #3128: Jump To (cmd-k) only searches current page & has missing actions
I noticed on the Colo rack that I can't use cmd-k to search for my project `peter` because it's on the second page. Looks like it only allows jumping to projects that are on the current page. This also impacts other pages too (e.g. instances, project images, silo images, etc).

| Current | Expected |
| - | - |
| Image | Image |

I also noticed that some views have show an action in that menu (Projects->New Project, Silo Images->Promote Image, etc) while in other places that action is missing: Projects->Disks->New Disk; Project/Silo Access->Add User or Group; etc).

| Current | Expected |
| - | - |
| Image | Image |

notpeter:
- Closes: https://github.com/oxidecomputer/console/issues/3128
- Closes: https://github.com/oxidecomputer/console/issues/2572

I was initially excited about using `cmd-k` to navigate the UI with the keyboard but immediately ran into some limitations and so took a stab at improving them.

- Fix multi-page search in cmd-k: Projects, Project Instances, Project VPCs, Project IP Pools, Silos
- Add missing search to cmd-k: Project Floating IPs, Silo Images, Project Images
- Add action from button to cmd-k:
  - Project Affinity: "New Group"
  - Project Access: "Add user or group"
  - Project Disks: "New disk"
  - Project Floating IPs: "New Floating IP"
  - Project Images: "Upload Image"
  - Project Snapshots: "New Snapshot"
  - Project VPC: "New VPC"
  - Silo Access: "Add user or Group"
  - Silo Images: Add "Promote image"

This is my first PR to oxidecomputer/console and was Claude assisted.
I have tested locally and verified these changes seem to work as expected. (see images in #3128)
If this is the wrong approach / redundant, feel free to summarily close.

### DIFF
--- a/app/pages/ProjectsPage.tsx
+++ b/app/pages/ProjectsPage.tsx
@@ -96,11 +96,13 @@ export default function ProjectsPage() {
   )
 
   const columns = useColsWithActions(staticCols, makeActions)
-  const { table } = useQueryTable({ query: projectList, columns, emptyState: <EmptyState /> })
+  const { table } = useQueryTable({
+    query: projectList,
+    columns,
+    emptyState: <EmptyState />,
+  })
 
-  const { data: allProjects } = useQuery(
-    q(api.projectList, { query: { limit: ALL_ISH } })
-  )
+  const { data: allProjects } = useQuery(q(api.projectList, { query: { limit: ALL_ISH } }))
   useQuickActions(
     useMemo(
       () => [
--- a/app/pages/ProjectsPage.tsx
+++ b/app/pages/ProjectsPage.tsx
@@ -5,6 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
+import { useQuery } from '@tanstack/react-query'
 import { createColumnHelper } from '@tanstack/react-table'
 import { useCallback, useMemo } from 'react'
 import { Outlet, useNavigate } from 'react-router'
@@ -24,6 +25,7 @@ import { CreateLink } from '~/ui/lib/CreateButton'
 import { EmptyMessage } from '~/ui/lib/EmptyMessage'
 import { PageHeader, PageTitle } from '~/ui/lib/PageHeader'
 import { TableActions } from '~/ui/lib/Table'
+import { ALL_ISH } from '~/util/consts'
 import { docLinks } from '~/util/links'
 import { pb } from '~/util/path-builder'
 
@@ -94,11 +96,11 @@ export default function ProjectsPage() {
   )
 
   const columns = useColsWithActions(staticCols, makeActions)
-  const {
-    table,
-    query: { data: projects },
-  } = useQueryTable({ query: projectList, columns, emptyState: <EmptyState /> })
+  const { table } = useQueryTable({ query: projectList, columns, emptyState: <EmptyState /> })
 
+  const { data: allProjects } = useQuery(
+    q(api.projectList, { query: { limit: ALL_ISH } })
+  )
   useQuickActions(
     useMemo(
       () => [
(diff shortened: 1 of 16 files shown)


app/pages/project/access/ProjectAccessPage.tsx
```diff
@@ -205,6 +206,18 @@ export default function ProjectAccessPage() {
...
+      () => [
+        {
+          value: 'Add user or group',
+          onSelect: () => setAddModalOpen(true),
+        },
+      ],
+      []
+    )
+  )
+
```
david-crespo, line 220: This made me wonder what happens when you use cmd+k when the form is already open. a) this entry still appears (not ideal but tolerable IMO — kind of complicated to avoid), b) visually the interaction with the form is messy (also not ideal but also tolerable I think)

Image

david-crespo: Oh and it's janky if you click the background because that only affects the form, so you close the form but not the ctrl-k popup. Out of scope for this PR, just noting it.

app/pages/system/silos/SilosPage.tsx
```diff
@@ -91,24 +93,24 @@ export default function SilosPage() {
...
         { value: 'New silo', onSelect: () => navigate(pb.silosNew()) },
-        ...(silos?.items || []).map((o) => ({
+        ...(allSilos?.items || []).map((o) => ({
           value: o.name,
           onSelect: () => navigate(pb.silo({ silo: o.name })),
           navGroup: 'Silo detail',
         })),
       ],
-      [navigate, silos]
+      [navigate, allSilos]
```
david-crespo, line 113: We need to make sure this hook behaves correctly when the silos request comes back (the old one would have been prefetched). I'm pretty sure it does — that's the point of this remove callback. On every re-run of the effect, we remove whatever the last one added before adding the new items.

I would leave this one not-prefetched, as you've done. Since quick actions are hidden until the user opens the menu, it's the perfect thing to defer.

https://github.com/oxidecomputer/console/blob/b43b995462cad77544e1092d964e96cb9588a705/app/hooks/use-quick-actions.tsx#L91-L106

app/pages/ProjectsPage.tsx
```diff
@@ -128,7 +132,7 @@ export default function ProjectsPage() {
         />
       </PageHeader>
       <TableActions>
-        <CreateLink to={pb.projectsNew()}>New Project</CreateLink>
+        <CreateLink to={pb.projectsNew()}>New project</CreateLink>
```
david-crespo, line 135: I did these out of perfectionism, but it doesn't matter in practice because the button text get all caps from CSS anyway.

david-crespo (approved):
Great!

app/pages/project/disks/DisksPage.tsx
```diff
@@ -186,6 +189,27 @@ export default function DisksPage() {
...
+        ...(allDisks?.items || []).map((d) => ({
+          value: d.name,
+          onSelect: () => navigate(pb.disk({ project, disk: d.name })),
+          navGroup: 'Go to disk',
+        })),
+      ],
+      [navigate, project, allDisks]
+    )
+  )
+
```
david-crespo, line 212: I tested the menu with 1000 disks at http://localhost:4000/projects/other-project/disks and filtering and scrolling were instantaneous. My computer is fast, but I think the menu is also just fast. It's just a big `ul`.

sudomateo:
As someone with a `matthewsanabria` project that was unable to Ctrl+K to the project before, thank you!

PR #3131, opened 2026-03-17
Quick actions refactor + new features

david-crespo:
#3129 inspired me to add a couple more features to the quick actions menu, which in turn inspired a big refactor to avoid problems with possible duplicate entries. I also added a bunch of e2e tests that we probably should have had before.

### New features 

- `ctrl+n`/`ctrl+p` keyboard navigation (in addition to arrow keys)
- "Go up" actions derived from breadcrumbs, so you can quickly navigate to parent pages
- Links are real links, which means you can middle click them if you insist

https://github.com/user-attachments/assets/d40d4d80-88b4-41e6-a41b-e7274520fa11

### Refactor

The old store held a flat list of `QuickActionItem`. 

https://github.com/oxidecomputer/console/blob/751a80e64b32bacbfc22379e70c24383685cb853/app/ui/lib/ActionMenu.tsx#L22-L28

`value` there is the text label displayed in the UI and (problematically) is also meant to uniquely identify the item. When calling pages (e.g., the project layout and the instance list page) called `useQuickActions` with a list of items, `useQuickActions` called `addActions`, which deduped the list of items based on `value`. The problem is that items with the same `value` in different groups could collide — removing by value would remove items you didn't meant to remove. We had an `invariant` call making sure the values were unique, but this is actually pretty bad in hindsight because it's probably possible for a user to name something to create a duplicate. (In practice we probably avoided this because all user-settable names are lower-case and all our nav items have upper case letters.)

In any case, the new breadcrumbs feature added a new source of items that could potentially be a duplicate with something else in the list. That is what prompted me to get rid of the deduping mechanism altogether. The new store is a `Map<string, QuickActionItem[]>` keyed by an ID for each calling page, so registration is just `map.set(id, items)` and cleanup is `map.delete(id)`. Deduping is unnecessary because each source component owns its own slot. In theory each source is responsible for deduping its items, but the way the item lists are constructed, this is never really a problem. And if a duplicate sneaks in, it doesn't actually matter.

Finally, since I was already messing everything up, I changed `QuickActionItem` to take both link and callback items, like we do for menu items. They're almost all links — only a few modal-opening actions are not.

### DIFF
--- a/app/hooks/use-quick-actions.tsx
+++ b/app/hooks/use-quick-actions.tsx
@@ -5,46 +5,51 @@
  *
  * Copyright Oxide Computer Company
  */
-import { useEffect, useMemo } from 'react'
-import { useLocation, useNavigate } from 'react-router'
+import { useEffect, useId, useMemo } from 'react'
+import { useLocation } from 'react-router'
 import { create } from 'zustand'
 
 import { useCrumbs } from '~/hooks/use-crumbs'
 import { useCurrentUser } from '~/hooks/use-current-user'
 import { ActionMenu, type QuickActionItem } from '~/ui/lib/ActionMenu'
-import { invariant } from '~/util/invariant'
 import { pb } from '~/util/path-builder'
 
 import { useKey } from './use-key'
 
-type Items = QuickActionItem[]
+export type { QuickActionItem }
 
 type StoreState = {
-  items: Items
+  // Multiple useQuickActions() hooks are active simultaneously — e.g., a
+  // layout registers "Create project" while a nested page registers "Create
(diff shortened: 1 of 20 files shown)


charliepark (approved):
Looks good

sudomateo:
As a Ctrl+N/P user this made me really happy. Thank you!

david-crespo:
> As a Ctrl+N/P user this made me really happy. Thank you!

Bad news! https://github.com/oxidecomputer/console/pull/3187

I could put it back for Mac only, though.

sudomateo:
Dang! Hey at least I can view all my projects now.

PR #3139, opened 2026-03-19
perf: reduce e2e test suite time by ~20%

benjaminleonard:
Reduce e2e test suite time by ~20% locally and ~3x on CI through five targeted optimizations:

1. **Shard e2e tests 3-way per browser** — Playwright distributes tests across machines with `--shard N/M`. This reduces wall-time per browser by ~66% since parallelism shifts from cores to jobs.
2. **Add FAST_MOCK env var** — Skip/reduce artificial API delays in mock handlers. Global delay: 50-150ms (was 200-400ms). Disk import/metrics kept higher (1000ms, 400ms) for tests that observe transient UI states.
3. **Reduce closeToast sleep** — 1000ms → 500ms. Saves ~12s across test suite.
4. **Lower expect timeout** — 10s → 7s for faster failure detection.
5. **Reduce scroll-restore sleeps** — 1000ms → 500ms per test.

Local wall-clock: **2:13 → 1:48** (19% faster). All 267 core tests pass; pre-existing flaky tests (pagination, action-menu) under full contention remain unchanged.

---

Running on GitHub to verify a. it passes and b. it has meaningful improvement without flake


david-crespo:
Eh, I don't know why but it doesn't feel necessary. Closing for now and then we can pull it back in if we feel like the playwright jobs are too slow.

PR #3153, opened 2026-03-25
Link to API and CLI docs in forms

david-crespo:
Fun and easy to do, but it needs a better visual treatment. Looked weird with `API` as a badge — it clashed with the external link icon. I'm always tempted to use the path instead of the operation ID, but the operation ID corresponds better to the method name in the SDK, SDK is probably more common than straight curl, and if they want the path they can get it from the docs.

image


david-crespo:
After a few months away, I think it looks fine?

PR #3158, opened 2026-03-28
Add IP pool sidebar

charliepark:
This adds a button to IP Pool Cells that opens up a sidebar with more info on the IP Pool.
Screenshot 2026-03-27 at 4 57 19 PM
Screenshot 2026-03-27 at 4 52 54 PM

I think it's a little strange that we have two different actions that both have the same UI — in one, we take the user to a new page, in the other, we open up this sidebar with more info. But if that's an issue, it's larger than this, as we already do both things in various points in the UI. I think it's just more noticeable here because the two kinds of cells are right next to each other in the Floating IP table.

We had also discussed the possibility of stacking sidebars, where we would make this IP pool cell in the properties table into a link that would then open another sidebar on top of it. We decided that this would be too confusing, though, so did not implement this.

Screenshot 2026-03-27 at 4 36 06 PM

Closes #3143


app/pages/project/floating-ips/FloatingIpsPage.tsx
```diff
@@ -78,10 +78,10 @@ export async function clientLoader({ params }: LoaderFunctionArgs) {
       .fetchQuery(q(api.ipPoolList, { query: { limit: ALL_ISH } }))
       .then((pools) => {
         for (const pool of pools.items) {
-          const { queryKey } = q(api.ipPoolView, {
-            path: { pool: pool.id },
-          })
-          queryClient.setQueryData(queryKey, pool)
```
david-crespo, line 84: The robot noticed this pre-existing bug: the point of this is to pre-cache the data for the IP pool cells, but it doesn't work because the IP pool cells use the errors-allowed query, which has a different query key. I confirmed that fixing this eliminates the loading state from the cells.

PR #3166, opened 2026-04-02
put in missing quick actions on VPC child tabs

david-crespo:
Noticed there were no create/edit quick actions on firewall rules, VPC routers, or VPC subnets.

image


charliepark (approved):
These look good. I added a few more in https://github.com/oxidecomputer/console/pull/3169 that can either be pulled into this, or directly to main once this one is merged.

PR #3178, opened 2026-04-13
Responsive grid layout: fixed TopBar/Sidebar with document scroll

benjaminleonard:
Refactor of https://github.com/oxidecomputer/console/pull/2087

Switch from a CSS grid layout (where TopBar and Sidebar occupied grid cells alongside ContentPane) to fixed-position TopBar and Sidebar with document-level scrolling. This is the foundation for mobile/tablet support and likely the place that has the potential to cause the most issues so I'm separating the two.

  - Viewport meta tag: proper responsive width=device-width instead of fixed width=1050
  - CSS: add --sidebar-width variable, responsive --content-gutter (smaller below 1000px), remove overflow-y-hidden on body, set height: 100% on html/body/#root
  - TopBar: single fixed element (was two Fragment children occupying grid cells)
  - Sidebar: fixed position with overflow scroll
  - ContentPane: margin-left offset for sidebar, document-level scroll via min-h-[calc(100vh-...)]
  - Scroll restoration: window.scrollY instead of container .scrollTop
  - PageSkeleton: matches new fixed layout structure
  - E2e: scroll-restore test updated for document scroll, expectScrollTop/scrollTo utils use window
  
This PR should produce a console that is visually identical to the current one.

---

Broader responsive changes will come in another PR, and are not a blocker for this.

Todo
- [x] Fix table dropdown overlapping header
- [x] Double check modals are not causing layout shift

### DIFF
--- a/app/components/PageSkeleton.tsx
+++ b/app/components/PageSkeleton.tsx
@@ -6,9 +6,15 @@
  * Copyright Oxide Computer Company
  */
 
+import cn from 'classnames'
 import { useLocation } from 'react-router'
 
-import { ContentPane, PageContainer } from '~/layouts/helpers'
+import {
+  ContentPane,
+  PageContainer,
+  sidebarWrapperClass,
+  topBarWrapperClass,
+} from '~/layouts/helpers'
 import { classed } from '~/util/classed'
 
 import { MswBanner } from './MswBanner'
@@ -29,7 +35,7 @@ export function PageSkeleton({ skipPaths }: { skipPaths?: RegExp[] }) {
       {process.env.MSW_BANNER ? <MswBanner disableButton /> : null}
       <PageContainer>
         {/* TopBar */}
-        <div className="bg-default border-secondary fixed inset-x-0 top-0 z-(--z-top-bar) grid h-(--top-bar-height) grid-cols-[var(--sidebar-width)_1fr] border-b">
+        <div className={topBarWrapperClass}>
           <div className="border-secondary flex items-center gap-2 border-r p-3">
             <Block className="h-8 w-8" />
             <Block className="h-4 w-24" />
@@ -43,7 +49,7 @@ export function PageSkeleton({ skipPaths }: { skipPaths?: RegExp[] }) {
           </div>
         </div>
         {/* Sidebar */}
-        <div className="border-secondary fixed top-(--top-bar-height) bottom-0 left-0 w-(--sidebar-width) border-r p-4">
+        <div className={cn(sidebarWrapperClass, 'p-4')}>
           <Block className="mb-10 h-4 w-full" />
           <div className="mb-6 space-y-2">
             <Block className="h-4 w-32" />
--- a/app/components/PageSkeleton.tsx
+++ b/app/components/PageSkeleton.tsx
@@ -28,18 +28,22 @@ export function PageSkeleton({ skipPaths }: { skipPaths?: RegExp[] }) {
     <>
       {process.env.MSW_BANNER ? <MswBanner disableButton /> : null}
       <PageContainer>
-        <div className="border-secondary flex items-center gap-2 border-r border-b p-3">
-          <Block className="h-8 w-8" />
-          <Block className="h-4 w-24" />
-        </div>
-        <div className="border-secondary flex items-center justify-between gap-2 border-b p-3">
-          <Block className="h-4 w-24" />
-          <div className="flex items-center gap-2">
-            <Block className="h-6 w-16" />
-            <Block className="h-6 w-32" />
+        {/* TopBar skeleton */}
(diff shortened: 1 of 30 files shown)


app/components/PageSkeleton.tsx
```diff
@@ -28,18 +28,22 @@ export function PageSkeleton({ skipPaths }: { skipPaths?: RegExp[] }) {
...
-          <Block className="h-8 w-8" />
-          <Block className="h-4 w-24" />
-        </div>
-        <div className="border-secondary flex items-center justify-between gap-2 border-b p-3">
-          <Block className="h-4 w-24" />
-          <div className="flex items-center gap-2">
-            <Block className="h-6 w-16" />
-            <Block className="h-6 w-32" />
+        {/* TopBar */}
+        <div className="bg-default border-secondary fixed inset-x-0 top-0 z-(--z-top-bar) grid h-(--top-bar-height) grid-cols-[var(--sidebar-width)_1fr] border-b">
```
benjaminleonard, line 32: There's an argument to say we should define the shared classes between the skeleton and regular layout to avoid drift but the juice might not be worth the squeeze.

app/hooks/use-scroll-restoration.ts
```diff
@@ -18,20 +18,19 @@ function setScrollPosition(key: string, pos: number) {
...
- * Note that `location.key` is used in the cache key, not `location.pathname`,
- * so the same path navigated to at different points in the history stack will
- * not share the same scroll position.
+ * Keep track of window scroll position before navigation and restore it on
+ * return (e.g., back/forward nav). Note that `location.key` is used in the
+ * cache key, not `location.pathname`, so the same path navigated to at
+ * different points in the history stack will not share the same scroll position.
  */
-export function useScrollRestoration(container: React.RefObject<HTMLElement | null>) {
+export function useScrollRestoration() {
```
benjaminleonard, line 26: Do we even need this hook now? Since we're using a regular window scroll, perhaps we can use the built-in `react-router` one.

app/layouts/helpers.tsx
```diff
@@ -14,18 +13,12 @@ import { useScrollRestoration } from '~/hooks/use-scroll-restoration'
 import { SkipLinkTarget } from '~/ui/lib/SkipLink'
 import { classed } from '~/util/classed'
 
-export const PageContainer = classed.div`grid h-screen grid-cols-[14.25rem_1fr] grid-rows-[var(--top-bar-height)_1fr]`
+export const PageContainer = classed.div`min-h-full pt-(--top-bar-height)`
```
benjaminleonard, line 16: Top bar is fixed, this leaves space for it. Alternatively we use `sticky` for the top bar but the sidebar would still be fixed and this _feels_ marginally cleaner.

benjaminleonard:
PR is visually identical to main with expanded visual regression tests. Haven't figured out an easy way to test it scrolled since the baseline would require a different scroll pattern to this new PR.

I'd suggest we err on merging sooner so it can be tested for longer.

app/hooks/use-scroll-restoration.ts
benjaminleonard: Similar to #2450 if I remove it I was unable to get `react-router` `ScrollRestoration` to work

app/components/PageSkeleton.tsx
david-crespo: could stick em in a string no problem

david-crespo:
No visual changes — love it. Found one issue: double scroll bar on side modal when height is small enough. Probably worth updating the regression test if it didn't catch this.

https://console-git-bl-responsiveness-grid-layout-oxidecomputer.vercel.app/projects/mock-project/disks-new

image

david-crespo:
4d4e03f938c91eec086b09575a4b9922b3a3985d fixes the side modal scroll bar thing by turning on `modal=true` in the base UI dialog we use for both side model and regular middle modal. `modal={false}` a workaround for a Radix issue covered in #1745, where a native `confirm()` dialog stealing and returning focus triggered Radix's focus-outside handler and caused a dismiss loop.

In base-ui, this covers it and the `modal=false` bit is unnecessary. 

https://github.com/oxidecomputer/console/blob/f25398dd9f16472ce6e756132a152cd2d7f5efec/app/ui/lib/Modal.tsx?plain=1#L54-L61

`modal={false}` has downsides — it disables scroll lock, focus trap, backdrop registration, and outside-content inerting — so it's nice to get rid of it.

In order to make this work I also added a scrim for the modal on top of a side modal situation with a special z-index that makes sure it's between the modal and the side modal and blocks interaction with the side modal. Right now on main you can interact with the image upload side modal while the main upload modal is going.

david-crespo:
I have fixes for the e2e failures but I want to merge #3192 into main and main into this first because it affects at least one of them.

PR #3180, opened 2026-04-14
Fix progress bar accent colour

benjaminleonard:
Latest design system update means`bg-accent` is now `bg-accent-inverse`. Confusing change? Yes, but semantically much better now.

image


david-crespo:
image

image

Issue #1920, opened 2024-01-29
Compressed images fail after upload


augustuswm:
As expected if you attempt to upload a compressed raw image, the upload will fail (we do not support it currently). In terms of network requests we see a number of normal `bulk-write` requests until we reach the end of the file at which point we receive a 400 response from nexus. The error looks like:
```json
{
    "request_id": "79a6c964-b41e-4691-933d-39a2b6626755",
    "error_code": "InvalidRequest",
    "message": "Data length not block size multiple"
}
```

![image](https://github.com/oxidecomputer/console/assets/1377730/11b0053d-2880-4484-b7ff-e59c551e48aa)

While we can't completely verify that the submitted file is correct, we can probably add some checks at file selection time to try and detect the error prior to uploading the entire image. Looking at the form data for the compressed vs uncompressed files in this case:

**Compressed**
```json
{
  "name": "Fedora-Cloud-Base-39-1.5.x86_64.raw.xz",
  "size": 446022784,
  "type": "application/x-xz"
}
```

**Uncompressed**
```json
{
  "name": "Fedora-Cloud-Base-39-1.5.x86_64.raw",
  "size": 5368709120,
  "type": "image/x-panasonic-rw"
}
```

We can probably use the type and size fields to make a guess about if the file will work or not. Given that this a rigorous check (the browser thinks the `.raw` file is a photograph) maybe we display some kind of warning if we detect specific formats that we know do not work.

And if my basic understanding of block size is correct we can also check the size reported by the browser against the selected block size. (Though I don't know how accurate these reported values are).

david-crespo:
We added the soft validation in https://github.com/oxidecomputer/console/pull/2217

However, the failure at the end of upload here seems like API bug that shouldn't be related to the image being compressed. Images that don't work should not fail until you actually try to boot them — the upload process is totally agnostic to the bytes being written to the disk.

david-crespo:
Just ran into this while trying to upload a bad image on purpose to create a failed instance. If we don't change the API logic, we could at least check that the content length in bytes is a multiple of block size before we let the user sit there for a while watching the file upload.

PR #3183, opened 2026-04-16
Validate client-side that image size is a multiple of block size

david-crespo:
See https://github.com/oxidecomputer/console/issues/1920#issuecomment-2788280175. Easiest thing in the world to do. Without this, the user can sit there for half an hour uploading a giant image file and only find out on the very last block that it is invalid.

Screenshot 2026-04-16 at 1 28 12 PM

### DIFF
--- a/app/forms/image-upload.tsx
+++ b/app/forms/image-upload.tsx
@@ -592,8 +592,6 @@ export default function ImageCreate() {
           units="Bytes"
           control={form.control}
           parseValue={(val) => parseInt(val, 10) as BlockSize}
-          // re-run imageFile validation when block size changes
-          rules={{ deps: ['imageFile'] }}
           items={[
             { label: '512', value: 512 },
             { label: '2048', value: 2048 },
--- a/app/forms/image-upload.tsx
+++ b/app/forms/image-upload.tsx
@@ -495,8 +495,8 @@ export default function ImageCreate() {
     setAllDone(true)
   }
 
-  // onChange mode so the file-size / block-size cross-validation surfaces
-  // inline as soon as the user picks a file or changes block size
+  // Surface file validation as soon as the user picks a file. Block-size
+  // changes are still validated on submit.
   const form = useForm({ defaultValues, mode: 'onChange' })
   const file = form.watch('imageFile')
   const blockSize = form.watch('blockSize')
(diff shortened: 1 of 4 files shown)


app/forms/image-upload.tsx
```diff
@@ -590,6 +592,8 @@ export default function ImageCreate() {
           units="Bytes"
           control={form.control}
           parseValue={(val) => parseInt(val, 10) as BlockSize}
+          // re-run imageFile validation when block size changes
+          rules={{ deps: ['imageFile'] }}
```
david-crespo: Omg. Claude is constantly teaching me about RHF features. wtf!

charliepark (approved):
This works. There's a small bit of friction in that if you change the block size before you select an image (a reasonable situation to find yourself in, as the block size comes first in the form), it fires off the validator, which then complains that no image has been loaded.
Image

Perhaps this isn't an issue, but I wanted to note it. I tried a few quick attempts to only conditionally make it required, but it was a bit messy and didn't feel like a better path.

david-crespo:
Good catch, to me that is an issue. One simple approach would be to just remove the explicit dep and rely on it being checked on submit. It’s slightly worse than getting that immediate feedback when you change the block size when an image is present, but better than the spurious error.

benjaminleonard:
Do you have confidence that there's no edge cases here? You might consider a soft validation to allow a user to submit even if it's unlikely to work.

david-crespo:
> Do you have confidence that there's no edge cases here? You might consider a soft validation to allow a user to submit even if it's unlikely to work.

I have no such confidence.

david-crespo:
@iliana or @augustuswm — thoughts on whether there are edge cases that mean we should make my `file.size % blockSize === 0` check a soft validation rather than a hard validation, per [Ben's question](https://github.com/oxidecomputer/console/pull/3183#issuecomment-4269762425)?

david-crespo:
I had the robot try to come up with edge cases that undermine the strict logic here. The only case it could come up with where this is technically stricter than we might want is if the last weird remainder block is all zeros. Because we skip sending all-zero blocks client-side anyway, you would be able to upload such an image with no issue — the unaligned last block would never be sent. This seems like a very unlikely scenario.

PR #3186, opened 2026-04-17
Fix dead space in combobox dropdown

### LINKED ISSUE #3185: Dead space around combobox button
Should open the listbox but doesn't.

https://github.com/user-attachments/assets/40a3a7b7-f53e-48e6-87fd-39c589be9237

charliepark:
This patches up the combobox by adding a pseudo element before the button, and then makes the button full-height.
Have checked it across browsers and all seems well.
Screen Recording 2026-04-17 at 7 52 43 PM mov

It was looking a little faint to me, but here's a quick comparison with the current style:

current:
Screenshot 2026-04-17 at 7 55 56 PM

with pseudo element:
Screenshot 2026-04-17 at 7 56 29 PM

Closes #3185

### DIFF
--- a/app/ui/lib/Combobox.tsx
+++ b/app/ui/lib/Combobox.tsx
@@ -252,7 +252,7 @@ export const Combobox = ({
               {items.length > 0 && (
                 <ComboboxButton
                   className={cn(
-                    'border-secondary my-1.5 flex items-center border-l px-3',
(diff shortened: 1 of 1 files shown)


charliepark:
This is with stroke-default, if we feel like this is better:
Screenshot 2026-04-17 at 8 09 44 PM

david-crespo (approved):
Works great, thanks! I like the line better dim. It's true that when you pay attention to it, it looks a little weird, but when you're just glancing at the button casually, it gives the button a nice boundary without calling attention to itself.

PR #3188, opened 2026-04-17
Upgrade react-hook-form and improve subnet pool member and IP range validation

david-crespo:
Resolvers are really for transforming form input into another shape. We were using it for form-level cross-field validation. The new form-level `validate` API added in https://github.com/react-hook-form/react-hook-form/pull/13195 is perfect for what we were doing. It does not override existing validations and it runs whenever validations are supposed to run (I think `resolver` only runs on submit.

### Interesting changes in RHF

I had Claude go through the releases since the old version and summarize the interesting stuff.

**Features worth a look**
- [v7.69.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.69.0) `reset({ keepIsValid: true })` fix — the existing `keepIsValid` option was buggy before. We don't currently use it but `SideModalForm`'s "name already exists" flow uses `setError` in a post-submit effect; switching to `errors` prop with focus ([v7.57.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.57.0) `focus form field for errors supplied by errors prop`) could be cleaner.
- [v7.65.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.65.0) `<Watch />` + [v7.68.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.68.0) `<FormStateSubscribe />` — render-prop components that re-render only when a named field (or formState slice) changes. `instance-create.tsx` and `firewall-rules-common.tsx` have lots of `useWatch` at the top level; some of that churn could be isolated inside a `<Watch>` render-prop. Worth trying only if profiling shows the re-renders hurt.
- [v7.61.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.61.0) `useWatch({ compute })` — subscribe to the whole form but only surface a derived value when a condition matches. Could collapse `useWatch + useMemo` pairs. `instance-create.tsx:451-463` computes `bootDiskSource` from four `useWatch` calls; a single `useWatch({ compute })` would work.
- [v7.63.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.63.0) `getValues(undefined, { dirtyFields: true })` — extract only dirty fields. Useful for PATCH-style edits; our edit forms mostly send the full object so the win is small.
- [v7.55.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.55.0) `createFormControl` / `subscribe` — subscribe to form state outside React (e.g., from a store). Not obviously useful here.
- [v7.56.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.56.0) reactive `mode` / `reValidateMode` — these become reactive to re-renders. The firewall-rules `HACK` comment at `firewall-rules-common.tsx:136-143` is specifically about the validate/reValidate regime swap, but that HACK is about `isSubmitted` state after reset, not about mode being non-reactive, so this doesn't directly help.

**Bug that might have been biting us**

- [v7.72.1](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.72.1) — `setValue` with `shouldDirty` no longer pollutes unrelated dirty fields. We use `setValue` heavily (silo-create, idp-create, instance-create); if any of those use `shouldDirty` we likely had ghost-dirty fields.
  - This sounds really familiar...........
- [v7.66.1](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.66.1) — `deepEqual` uses `Object.is` for NaN. `subnet-pool-member-add` stores `NaN` as the unset state for optional NumberFields — pre-fix, `isDirty` might have flipped on those.
- [v7.69.0](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.69.0) — race between `setError` and `setFocus` resolved. Relevant to our `SideModalForm` effect that calls both.

### DIFF
--- a/app/forms/ip-pool-range-add.tsx
+++ b/app/forms/ip-pool-range-add.tsx
@@ -5,7 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
-import { useForm, type FieldErrors } from 'react-hook-form'
+import { useForm } from 'react-hook-form'
 import { useNavigate } from 'react-router'
 
 import {
@@ -34,57 +34,11 @@ const defaultValues: IpRange = {
   last: '',
 }
 
-// Using a resolver overrides all field-level validation (required, min, max,
-// etc.), so this function must cover everything. Field-level `required` props
-// still affect UI display (hiding the "optional" label) but are inert for
-// validation.
-
-/**
- * Validates IP range addresses against the pool's IP version.
- * Ensures both addresses are valid IPs and match the pool's version.
- */
-function createResolver(poolVersion: IpVersion) {
-  return (values: IpRange) => {
-    const first = parseIp(values.first)
-    const last = parseIp(values.last)
-
-    const errors: FieldErrors<IpRange> = {}
-
-    // Validate first address matches pool version
-    if (first.type === 'error') {
-      errors.first = { type: 'pattern', message: first.message }
-    } else if (first.type === 'v4' && poolVersion === 'v6') {
-      errors.first = {
-        type: 'pattern',
-        message: 'IPv4 address not allowed in IPv6 pool',
-      }
-    } else if (first.type === 'v6' && poolVersion === 'v4') {
-      errors.first = {
-        type: 'pattern',
-        message: 'IPv6 address not allowed in IPv4 pool',
-      }
-    }
-
-    // Validate last address matches pool version
-    if (last.type === 'error') {
-      errors.last = { type: 'pattern', message: last.message }
-    } else if (last.type === 'v4' && poolVersion === 'v6') {
-      errors.last = {
-        type: 'pattern',
-        message: 'IPv4 address not allowed in IPv6 pool',
-      }
-    } else if (last.type === 'v6' && poolVersion === 'v4') {
-      errors.last = {
-        type: 'pattern',
-        message: 'IPv6 address not allowed in IPv4 pool',
-      }
-    }
-
-    // TODO: if we were really cool we could check first <= last but it would add
-    // 6k gzipped to the bundle with ip-num
-
-    // no errors
-    return Object.keys(errors).length > 0 ? { values: {}, errors } : { values, errors: {} }
+const validateAddress = (value: string, poolVersion: IpVersion) => {
+  const parsed = parseIp(value)
+  if (parsed.type === 'error') return parsed.message
+  if (parsed.type !== poolVersion) {
+    return `IP${parsed.type} address not allowed in IP${poolVersion} pool`
   }
 }
 
@@ -107,10 +61,7 @@ export default function IpPoolAddRange() {
     },
   })
 
-  const form = useForm({
-    defaultValues,
-    resolver: createResolver(poolData.ipVersion),
-  })
+  const form = useForm({ defaultValues })
 
   return (
     <SideModalForm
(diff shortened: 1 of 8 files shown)


app/forms/ip-pool-range-add.tsx
```diff
@@ -34,57 +34,11 @@ const defaultValues: IpRange = {
...
-// Using a resolver overrides all field-level validation (required, min, max,
-// etc.), so this function must cover everything. Field-level `required` props
-// still affect UI display (hiding the "optional" label) but are inert for
-// validation.
-
-/**
- * Validates IP range addresses against the pool's IP version.
- * Ensures both addresses are valid IPs and match the pool's version.
- */
-function createResolver(poolVersion: IpVersion) {
```
david-crespo, line 46: I believe this was originally a resolver because pools were not locked to a single IP version, so we had to make sure the start and end were of the same version. Now we can just compare each one individually to the version of the pool. So this can all just be a field-level validation. This change does not rely on form-level validation, this was just discovered while looking for candidates for form-level validation.

app/forms/subnet-pool-member-add.tsx
```diff
@@ -41,65 +41,62 @@ const defaultValues: MemberAddForm = {
...
-      errors.subnet = { type: 'pattern', message: parsed.message }
-    } else if (parsed.type !== poolVersion) {
-      errors.subnet = {
-        type: 'pattern',
-        message: `IP${parsed.type} subnet not allowed in IP${poolVersion} pool`,
-      }
+// Uses form-level validate (RHF ≥7.72.0) so we can look at all three fields
+// together. Unlike `resolver`, this runs alongside field-level validation, so
+// `required` / `min` / `max` on the fields still apply.
+export function validateForm(poolVersion: IpVersion, values: MemberAddForm) {
```
david-crespo, line 47: Look at this with whitespace changes hidden to see what a small change it is. Basically it's just the return type changing.

https://github.com/oxidecomputer/console/pull/3188/changes?w=1

app/forms/subnet-pool-member-add.tsx
```diff
@@ -41,65 +41,62 @@ const defaultValues: MemberAddForm = {
...
+  const maxBound = poolVersion === 'v4' ? 32 : 128
+  const parsed = parseIpNet(values.subnet)
+  const { minPrefixLength: minPL, maxPrefixLength: maxPL } = values
+  const subnetWidth = parsed.type !== 'error' ? parsed.width : undefined
+  const inRange = (v: number) => !Number.isNaN(v) && v >= 0 && v <= maxBound
+
+  const errors: Partial<Record<keyof MemberAddForm, { type: string; message: string }>> = {}
```
david-crespo, line 54: Don't love this, will revisit.

david-crespo:
Good showing for GPT-5.5 in my first test of it: I had it review this PR and it pointed out I had misunderstood what form-level `validate` is capable of. It seems aimed at form-level errors; it cannot actually set field-level errors. The test added in 36b1884f1d88e401306163fea9a590159421d79f fails because the error that end is supposed to be >= start does not show up. The fix in 6e30acb27355d9544a1bf78c82a8a5043ec84c90 is to go back to field-level validation but take advantage of `deps` (which I just learned about last week) to make so that whenever you change the start of the range, we validate the end too, and vice versa.

PR #3192, opened 2026-04-21
Display server errors in modals, don't use toasts

david-crespo:
Found this while working on #3178. Instance resize, attach floating IP, and attach ephemeral IP all had `onError`s that triggered toasts on server error. The first two also displayed the error in the modal. Ephemeral IP did not. This PR gets rid of the toasts and puts an error display into attach ephemeral IP. It doesn't look very good (maybe we should move the error and give it a little box or something), but it's basically the same as it already was, minus the extra toasts.

(Screenshot is from playwright, hence the weird missing submit button — I think it's right in the middle of switching between the button text and the loading spinner.)

Screenshot 2026-04-21 at 5 20 23 PM


benjaminleonard:
Nice, in the side modals we add it at the top in an error message block. Should probably just follow that pattern.

PR #3200, opened 2026-04-23
Loading spinner in attach ephemeral IP modal

### LINKED ISSUE #3197: Considerable delay on feedback for 'Attach Ephemeral IP' button + not following 'Attach Floating IP' pattern
In Dogfood, I tried attaching an ephemeral IP from the fake IPv6 pool, but when I clicked on the 'ATTACH' button, I got no feedback. The button didn't change with the small spinner, and there was no notification of an error happening.

I then went on to try and attach a floating IP and -as you can see from the video- the ATTACH button does have a spinner, when it gets clicked.

Eventually, the 'Failed to fetch' messages from the ephemeral IP pop up on the bottom left.
In contrast, the failures for floating IPs appear in the same section as the dropdown menu and button to attach one.

https://github.com/user-attachments/assets/a16b2d0c-9ae2-4572-adc7-be1f108dc10b

Image

david-crespo:
Closes #3197. Turns out we were using a bespoke modal + form thing instead of our dedicated `ModalForm` component that handles the loading state in the button. Turns out there are a bunch of modal forms that aren't using `ModalForm` like they probably should. I'll fix the other ones in another PR.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,6 +67,7 @@
 - Store API response objects in the mock tables when possible so state persists across calls.
 - Enforce role checks with `requireFleetViewer`/`requireFleetCollab`/`requireFleetAdmin`, and return realistic errors (e.g. downgrade guard in `systemUpdateStatus`).
 - All UUIDs in `mock-api/` must be valid RFC 4122 (a safety test enforces this). Use `uuidgen` to generate them—do not hand-write UUIDs.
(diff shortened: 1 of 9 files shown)


david-crespo:
@charliepark I pulled in your `submitDisabled?: string` from #3201

charliepark (approved):
Looks good, and I like the sentinel approach to testing. Certainly easier than mucking around with blocking resources in developer tools.

david-crespo:
I had to add it to the CLAUDE.md because Claude was happy to write an interceptor in the test that makes the response an error and I didn’t like that. It’s not terrible, but I didn’t like how it has to match by URL path.

PR #3201, opened 2026-04-23
Use the ModalForm component for Ephemeral IPs

### LINKED ISSUE #3197: Considerable delay on feedback for 'Attach Ephemeral IP' button + not following 'Attach Floating IP' pattern

charliepark:
Currently, our Attach Ephemeral IP modal is using a custom modal component, when we can use the ModalForm component to bring in the benefits of a loading spinner and an error message, like we have on the Attach Floating IP modal.
This builds on #3192, which moved the messages from toasts into the modals.

Screenshot 2026-04-23 at 11 25 31 PM

The spinner is a little high on the button, but that's just because it's hard to capture the moment, and the spinner drops in on the button.
Screenshot 2026-04-23 at 11 28 24 PM

Closes #3197

### DIFF
--- a/app/components/form/ModalForm.tsx
+++ b/app/components/form/ModalForm.tsx
@@ -18,8 +18,8 @@ type ModalFormProps<TFieldValues extends FieldValues> = {
   form: UseFormReturn<TFieldValues>
   children: ReactNode
 
-  /** Must be provided with a reason describing why it's disabled */
-  submitDisabled?: boolean
+  /** When set, disables the submit button and shows this string as the disabled reason */
+  submitDisabled?: string
   onSubmit: (values: TFieldValues) => void
   submitLabel: string
 
@@ -35,7 +35,7 @@ export function ModalForm<TFieldValues extends FieldValues>({
   form,
   children,
   onDismiss,
-  submitDisabled = false,
+  submitDisabled,
   submitError,
   title,
   onSubmit,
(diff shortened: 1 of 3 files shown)


charliepark:
Welp. Looks like this is a dupe of #3200

app/components/form/ModalForm.tsx
```diff
@@ -18,8 +18,8 @@ type ModalFormProps<TFieldValues extends FieldValues> = {
   form: UseFormReturn<TFieldValues>
   children: ReactNode
 
-  /** Must be provided with a reason describing why it's disabled */
-  submitDisabled?: boolean
+  /** When set, disables the submit button and shows this string as the disabled reason */
+  submitDisabled?: string
```
david-crespo, line 22: I don't have this change in mine but I think it might be the right idea. Disabling the button without a message feels kinda junky.

PR #3205, opened 2026-04-30
Improve confirm modal copy across the board

david-crespo:
* Require resource kind so there are no weird "Confirm delete" without a resource
* Reword all confirm modals: "Confirm delete snapshot" -> "Delete snapshot", "Confirm unlink pool" -> "Unlink pool", etc.

### DIFF
--- a/app/pages/ProjectsPage.tsx
+++ b/app/pages/ProjectsPage.tsx
@@ -89,6 +89,7 @@ export default function ProjectsPage() {
         onActivate: confirmDelete({
           doDelete: () => deleteProject({ path: { project: project.name } }),
           label: project.name,
+          resourceKind: 'project',
         }),
       },
     ],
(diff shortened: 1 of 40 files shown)


david-crespo:
Lol at the status changes. See https://github.com/kitlangton/ghui/issues/4

charliepark (approved):
I would maybe prefer `Confirm ${resourceKind} deletion` over `Confirm delete ${resourceKind}`, but not a blocker.

david-crespo:
I think you’re right that there’s something grammatically funny about “Confirm delete snapshot” but I don’t like the passive “deletion” either. Maybe “Delete snapshot” or “Delete snapshot?”

charliepark:
Yeah, that's shorter and snappier. I think either with or without the question mark works. Maybe a slight tilt towards `no question mark`, since the body copy in the modal already has the question mark, but I'm good with either.

david-crespo:
I like it! I'm making `resourceKind` required so they all have it.

image

david-crespo:
image

PR #3208, opened 2026-05-02
minor: use pre-stopped mock instance in e2e tests

david-crespo:
A bunch of tests (11 tests) have to wait for an instance to stop in order to do things you need to stop the instance to do. If the instance is already stopped it saves like 5 seconds per test. The savings will probably be diluted by 5x parallelism, so it may only save 10 seconds on the full run, but it's probably still worth it.


david-crespo:
Saves about 20 seconds overall. I’ll take it!

PR #3210, opened 2026-05-03
Don't show fractional bytes in file size display

### LINKED ISSUE #3209: fractional bytes don't make sense
Image

david-crespo:
filesize's `pad: true` option pads results like "5 B" to "5.00 B", but
fractional bytes are nonsensical. Add a `formatBytes` wrapper that
suppresses padding when the unit is bytes, and use it in FileInput and
the image upload progress display.

Closes #3209

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -54,8 +54,16 @@
     "no-restricted-imports": [
       "error",
       {
-        // prevent confusion due to auto-imports and barrel files
-        "paths": ["."],
+        "paths": [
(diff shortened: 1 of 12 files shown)


david-crespo:
Initial Claude solution was sufficient to fix the problem, but caused some fragmentation by introducing a new helper that we only use in some spots. I updated the helper so it could be used everywhere we currently call `filesize` directly.

charliepark (approved):
👍

PR #3211, opened 2026-05-04
Add full User and Groups lists to Access page

### LINKED ISSUE #1146: Group detail page listing users
When you see a group on the IAM page, you should be able to figure out who is in it. The simple version of this is the group links to a group page that gives you a list of users. Maybe we eventually want to be cute and do this in a popover in the policy table instead, but it could be a long list that could be unwieldy in a table. On the other hand you could do client side fuzzy filtering to let the user figure out if a particular user is in the list.

### LINKED ISSUE #1359: v1 Access & IAM - tabs for all, users, groups
See [latest designs for Access & IAM](https://www.figma.com/file/NcUHVumfS9qzinnZhanL4F/Access-%26-IAM?node-id=1%3A2023&t=L2MmosHZX2J3Uqk6-1).

image

### LINKED ISSUE #2887: Want user and group list for silo admins
This is related to SCIM but not really tied to SCIM. Currently, the only way a silo or project admin can see the list of existing users and groups is via the picklists on the **Add User or Group** popup page in **Access Control** (when they select which entities to be granted access to a silo or project).

With SCIM, there will likely be more utility of such UIs since the use case often involves importing users and groups ahead of time and confirming their existence in the rack to debug IDP-side import issues.

charliepark:
Currently, the Silo Access and Project Access page don't show all users, except when you open up the "Add user or group" form and access the dropdown:
Screenshot 2026-06-26 at 5 20 28 PM
Screenshot 2026-06-26 at 5 20 41 PM
Screenshot 2026-06-26 at 5 20 52 PM
Screenshot 2026-06-26 at 5 20 59 PM

In issue #2887, @askfongjojo requested a view where all users and groups would be displayed. This PR does that, adding a tabbed interface to show Groups / Users, and, then, on those pages, shows all groups and users.
Screenshot 2026-06-29 at 3 13 49 PM
Screenshot 2026-06-29 at 3 10 57 PM

From this tabbed view, an admin can make edits to a specific user / group's permissions, as before:
Screenshot 2026-06-29 at 11 48 11 AM

One ongoing situation that I want to mention is that we don't have a way to order / filter the responses from the API, and the default ordering from the API is by User ID. Instead of having pagination with 50 users per page, I have it selecting all-ish users (1,000) and then sorting them by name on the client. This deviates from the vanilla API response, but gives, I think, a more useful listing of users. I'm open to other approaches, if people have suggestions. I checked with Will about whether that 1,000 user cap would be a problem, and he said that in the far future it might be something to address, but that there are no current customers where that limit would be problematic, based on how they configure users/silos.

Closes #2887 
Closes #1146 
Closes #1359


david-crespo:
The thing that I am nervous about here is merging the concept of the list of users with the concept of role access. The nice thing (even though it's also confusing) about the current access pages is that they only show what is in the role policy response for the silo and the project (or the fleet in the case of the fleet\*). In the design in this PR, we have a user list and then we annotate it with role assignments. The problem for me in practice is that you might have a list of 100 users, but let's say zero or one or two of them has a role assigned, and everything else is done through groups. Similar, you might have 20 groups, but only one of them is the one you use for role assignments. In this situation (which I think is close to typical for our customers), the current role policy page that only shows the role assignments tells you what you want to know in a much more digestible way, and if we want to list all users and groups I'd rather it be on separate tabs that are dedicated to that. Now, we might still want to combine the two approaches — keep the policy page, and then add tabs for users and groups, but those tabs _also_ include role assignments in the big list. 

So I guess the short version is: what if we keep something like the current page as the `Policy` or `Role Policy` tab and add Users and Groups alongside it? Another open question for me is whether they should all be on the Access page or we might want to separate them in the sidebar into something like Silo Access and Users & Groups.


\* Ooh, I think we added the fleet access page since this was started. Something to think about. The concept of listing all users doesn't make sense at the fleet level, so there's necessarily a mismatch there. This is made a little easier.

david-crespo:
I like synthesizing the user's effective role from their group memberships, but I worry that it is doing too much API-like logic. Worth considering modifying the user list API to include group memberships for each user. It might not be a big deal.

image

charliepark:
<s>I asked the robot to create a plan for how to proceed if [Omicron#10828](https://github.com/oxidecomputer/omicron/pull/10828) is merged. It is here: https://gist.github.com/charliepark/a747d243f6f71a0d2aa5bf874e56f157</s>
Edit: Per [this comment from Crespo on Omicron](https://github.com/oxidecomputer/omicron/pull/10828#issuecomment-5018200361), we probably won't update the users API, and will continue to do it locally in the client (as in this PR).

david-crespo:
Ooh, very important: if we keep the top-level `/users` and `/groups` routes we will need route handlers for them in the API so that if someone goes to the route directly it doesn't 404.

https://github.com/oxidecomputer/omicron/blob/4b0ce4bfaa9b2e3fc46a75d05c9a59dff338890c/nexus/src/external_api/http_entrypoints.rs#L8765-L8769

https://github.com/oxidecomputer/omicron/blob/4b0ce4bfaa9b2e3fc46a75d05c9a59dff338890c/nexus/external-api/src/lib.rs#L8921-L8928

david-crespo:
This works, but it might be nice to have a more specific icon to indicate that the role comes from a group.

image

We do have a group icon, though it's 16 instead of 12. Kind of interesting.

image

david-crespo:
That remove is disabled makes sense, but it's a little confusing that you can still use the change role button. I understand what's happening — you can assign the user a role directly in addition to their assignment from the group, but "change user role" doesn't get that across.  

image

You can see the behavior here: setting a user role `<=` the group one doesn't have any visible effect, but setting a strong role does because it becomes the new effective role. I'm not sure what to do about this but it is definitely confusing. To some extent it is a problem of the intrinsic complexity of how our things work, but I wonder if there's a way we could convey it better visually, maybe by doing less collapsing down to the effective role.

https://github.com/user-attachments/assets/e87c0eb5-1c72-4aae-bbea-6321fe82deb2

charliepark:
One possibility on the roles is that we use the +N badge and modal again. Downside is that we're leaning on that tool an awful lot in this one UI, but it could give us the option of showing all roles, in descending order of strength, like we do in the user sidebar of the Project Access view.
Screenshot 2026-07-23 at 11 08 43 AM

Some examples:
Screenshot 2026-07-23 at 11 48 16 AM
Screenshot 2026-07-23 at 11 48 40 AM


EDIT: We experimented with this, but it ended up being noisy (having both the `via $group` icon and the `+N` target) or a little unclear (removing the `via $group` icon), so we've backed off this approach.

app/layouts/SiloLayout.tsx
```diff
@@ -45,6 +47,11 @@ export default function SiloLayout() {
     [pathname, me.siloName]
   )
 
+  // Users & Groups spans two sibling routes, so highlight the nav item on both
+  const inUsersGroups = [pb.siloUsers(), pb.siloGroups()].some(
+    (p) => pathname === p || pathname.startsWith(`${p}/`)
+  )
+
```
david-crespo, line 54: Kind of a smell about the IA maybe? But I can't think of a good top level route that we could stick `/users` and `/groups` under.

app/layouts/SiloLayout.tsx
```diff
@@ -63,9 +70,12 @@ export default function SiloLayout() {
           <NavLinkItem to={pb.siloUtilization()}>
             <Metrics16Icon /> Utilization
           </NavLinkItem>
-          <NavLinkItem to={pb.siloAccess()}>
+          <NavLinkItem to={pb.siloAccess()} end>
```
david-crespo: I don't think this `end` is necessary — maybe left over from when they were all on the same page?

Issue #3083, opened 2026-02-23
Consider invalidating all queries on refresh instead of listing relevant ones


david-crespo:
I was thinking about how we could make #3082 harder to mess up, and I realized that it might work better to  have the refresh button invalidate all queries on the page instead of a hard-coded list of relevant ones. This would mean we can't miss anything. And really we could extend this to every mutation we do — right now we manually invalidate the queries that we determine are relevant to the mutation that just happened, and I'm sure we get it wrong in some cases.

<details>
<summary>Diff showing what it would be like on instance detail</summary>

```diff
diff --git a/app/components/RefreshButton.tsx b/app/components/RefreshButton.tsx
index 2308ca7a8d..6ee3d6ed56 100644
--- a/app/components/RefreshButton.tsx
+++ b/app/components/RefreshButton.tsx
@@ -8,17 +8,22 @@
 
 import { useState } from 'react'
 
+import { queryClient } from '@oxide/api'
 import { Refresh16Icon } from '@oxide/design-system/icons/react'
 
 import { Button } from '~/ui/lib/Button'
 import { SpinnerLoader } from '~/ui/lib/Spinner'
 
-export function RefreshButton({ onClick }: { onClick: () => Promise<unknown> }) {
+/**
+ * Invalidates all queries, which causes active (currently rendered) queries to
+ * refetch. Inactive cached queries are marked stale and refetch on next use.
+ */
+export function RefreshButton() {
   const [refreshing, setRefreshing] = useState(false)
 
   async function refresh() {
     setRefreshing(true)
-    await onClick()
+    await queryClient.invalidateQueries()
     setRefreshing(false)
   }
 
diff --git a/app/pages/project/instances/InstancePage.tsx b/app/pages/project/instances/InstancePage.tsx
index e968ee18f0..596f365c54 100644
--- a/app/pages/project/instances/InstancePage.tsx
+++ b/app/pages/project/instances/InstancePage.tsx
@@ -76,22 +76,6 @@
   return nic ? nic.vpcId : undefined
 }
 
-// this is meant to cover everything that we fetch in the page
-async function refreshData() {
-  await Promise.all([
-    queryClient.invalidateEndpoint('instanceView'),
-    queryClient.invalidateEndpoint('instanceExternalIpList'),
-    queryClient.invalidateEndpoint('instanceNetworkInterfaceList'),
-    queryClient.invalidateEndpoint('instanceDiskList'), // storage tab
[...]

PR #3214, opened 2026-05-11
Add source cell for disks, with side modal

### LINKED ISSUE #2057: Disk: Display the image or snapshot name from which it is created

charliepark:
This PR adds a "source" cell to each row in the Disks table, and to each Instance's "Storage" tab.

Screenshot 2026-05-11 at 4 49 59 PM

For each of those locations, it provides a side modal with more info on the source disk / snapshot / image.
Screenshot 2026-05-11 at 4 50 20 PM
Screenshot 2026-05-11 at 4 50 13 PM

In the Disk sidemodal, the source cell is not a link, to prevent stacking of side modals. This follows the pattern on https://github.com/oxidecomputer/console/pull/3158. I also added a badge to the side modal to indicate whether the source is from an image or a snapshot. We can remove that if we feel like it's too noisy. I left it off on the main view, since you can click into it to get more details.
Screenshot 2026-05-13 at 11 30 29 AM

Note that this image detail side modal has a slightly different layout than the existing image detail side modal, and we might want to just use the existing one. I do think the existing one has more of a pseudo-form feel to it, which might be more appropriate in its current location as a drill-down on the images index list, which looks like this:
Screenshot 2026-05-13 at 4 51 49 PM

Closes #2057


david-crespo:
The test failures were the first hour of the month UTC bug fixed in #3274. No need to merge main, re-running should work.

david-crespo:
I think you should consolidate the image side modals on the newer non-form version, like you said. I think we did the old one as a pseudo-form in anticipation of it becoming editable, and that's not really on the way.

david-crespo:
Pushed some small fixes based on a Fable review. The only one that's more than a few lines is d63b3d17e4deadc32aca60e898001e93e32f8e97, which pulls out a shared `DiskNameFromId` and a couple of query objects that are used in a few spots.

I increased the gap here from 1 to 1.5:

image

I also added `Image` and `Snapshot` to the deleted badge. Fable's idea — I don't think I would have thought of it, but I like it.

image

image

test/e2e/disks.e2e.ts
```diff
@@ -27,6 +27,47 @@ test('Disk detail side modal', async ({ page }) => {
   await expect(modal.getByText('2 GiB')).toBeVisible()
   await expect(modal.getByText('2,048 bytes')).toBeVisible() // block size
   await expect(propertiesTableValue(modal, 'Read only')).toHaveText('False')
+
+  // the ID is truncated for display, but the full ID is in the aria-label,
+  // next to a copy button
+  const idCell = propertiesTableValue(modal, 'ID')
+  await expect(idCell.getByLabel('7f2309a5-13e3-47e0-8a4c-2a3b3bc992fd')).toBeVisible()
+  await expect(idCell.getByRole('button', { name: 'Click to copy' })).toBeVisible()
+})
```
david-crespo, line 36: Something like this had been deleted below (I think rightly, because whatever it is no longer has an ID in the same spot). The robot noticed it was the only place in the e2es where we were exercising this aspect of how `IdRow` works, so it wanted to add an assert about it back somewhere else where it would still work.

david-crespo (approved):
Looks great! 

Something for us to mull over: the instance cells here go to the instance page, but the source cells open a side modal. It's not a big deal, but it would be interesting if there was some way to indicate visually what to expect when you click. Maybe a link vs. button distinction, but it's hard to see how to do that here, plus buttons can open pages too. 

Image

app/forms/image-upload.tsx
```diff
@@ -246,6 +246,8 @@ export default function ImageCreate() {
   const finalizeDisk = useApiMutation(api.diskFinalizeImport)
   const createImage = useApiMutation(api.imageCreate)
   const deleteDisk = useApiMutation(api.diskDelete)
+  // no invalidation needed: the deleted snapshot is the transient one created
+  // by this flow, so nothing can be displaying it
   const deleteSnapshot = useApiMutation(api.snapshotDelete)
```
david-crespo, line 251: There were some invalidations added here that were probably harmless but I'd rather avoid them because they give the impression of being necessary. On the other hand I'm always wondering https://github.com/oxidecomputer/console/issues/3083 whether we should be more aggressive about invalidations. But I don't think we tend to have issues.

PR #3215, opened 2026-05-13
Terminal: Disable bold, tweak line height

benjaminleonard:
**Line Height**
In xterm.js, `lineHeight` is a multiplier on the *measured* cell height, not the font size. So a `lineHeight` of 1 is actually closer to what we'd expect here.

**Bold**
There seems to be a bug where setting bold makes the cell width wider. Not sure if this is on `xterm.js` or its the font metrics. Better instead to just disable bold.

Before:
image

After:
image

### DIFF
--- a/app/components/Terminal.tsx
+++ b/app/components/Terminal.tsx
@@ -58,7 +58,9 @@ function getOptions(): ITerminalOptions {
     fontFamily: '"GT America Mono", monospace',
     fontSize: 13,
     lineHeight: 1,
-    fontWeightBold: 'normal',
+    fontWeightBold: 400,
+    drawBoldTextInBrightColors: true,
+    letterSpacing: 0,
     windowOptions: {
       fullscreenWin: true,
       refreshWin: true,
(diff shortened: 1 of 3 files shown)


benjaminleonard:
Related: https://github.com/oxidecomputer/stlouis/issues/973

charliepark (approved):
Sorry to have missed this earlier. Looks good.

app/components/Terminal.tsx
```diff
@@ -57,7 +57,10 @@ function getOptions(): ITerminalOptions {
     screenReaderMode: true,
     fontFamily: '"GT America Mono", monospace',
     fontSize: 13,
-    lineHeight: 1.2,
+    lineHeight: 1,
+    fontWeightBold: 400,
+    drawBoldTextInBrightColors: true,
+    letterSpacing: 0,
```
charliepark, line 63: The robot notes that these two (`drawBoldTextInBrightColors: true` and `letterSpacing: 0`) are xterm v6 defaults, so might not be necessary.

PR #3216, opened 2026-05-14
Direct user to firewall rules tab from instance networking view

### LINKED ISSUE #2064: Instance: Make firewall rules more visible
Users new to us are not necessarily aware of the need to add/modify firewall rules in order to access their applications outside of the rack. The "VPC" name on the instance page (or Create Instance page) also does not suggest anything related to security for that matter. Here are two suggestions for users to have a better chance of getting the firewall rules right:

1. Create Instance: provide a link on/near the VPC name that lands on the Firewall Rules tab selected by the user (perhaps open a new tab so user can stay on Create Instance); this allows user to confirm if the VPC selected is right for them.
2. Instance details: provide a similar link to the VPC Firewall Rules or add another subtab (named more generally as "Security", similar to other public cloud portals) that actually displays all the rules in that VPC.

charliepark:
This PR adds a message box — directing the user to the Firewall Rules tab of the VPC — to the instance's Networking tab.

image

When there are multiple VPCs:
image

I'm very open to alternate designs and copy.

Closes #2064

### DIFF
--- a/app/pages/project/instances/NetworkingTab.tsx
+++ b/app/pages/project/instances/NetworkingTab.tsx
@@ -59,7 +59,6 @@ import { Button } from '~/ui/lib/Button'
 import { CardBlock } from '~/ui/lib/CardBlock'
 import { CopyableIp } from '~/ui/lib/CopyableIp'
 import { EmptyMessage } from '~/ui/lib/EmptyMessage'
-import { Message } from '~/ui/lib/Message'
 import { TableEmptyBox } from '~/ui/lib/Table'
 import { TipIcon } from '~/ui/lib/TipIcon'
 import { Tooltip } from '~/ui/lib/Tooltip'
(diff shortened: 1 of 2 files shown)


benjaminleonard:
This is perhaps a misappropriation of the message box.
If anything it probably needs to sit within its own property box. Perhaps one that refers to the VPC. Though there isn't really any other related VPC info we'd want to plumb through.

Let me noodle on this further.

david-crespo:
Another option, like what Charlie did in https://github.com/oxidecomputer/console/issues/2064#issuecomment-4454548317, but with a title on the card block.

image

benjaminleonard:
Is an instance only affected by firewall rules on one (the primary) VPC?

david-crespo:
I think any of them can affect the instance, but I'm not 100% sure. Most of the time there will only be one anyway, so in theory we could detect that and phrase it accordingly.

david-crespo:
I was wrong above: only the VPC associated with the primary network interface (which is already the one we're displaying at the top of the page) is relevant. So we can word this better:

> Manage firewall rules affecting this instance in VPC [default](example.com)

david-crespo:
@benjaminleonard feels this is the least bad of our bad ideas. I think it is livable.

### Primary network interface exists

Screenshot 2026-06-01 at 5 00 34 PM

### No primary network interface

Screenshot 2026-06-01 at 5 00 53 PM

app/pages/project/instances/NetworkingTab.tsx
```diff
@@ -739,7 +749,7 @@ export default function NetworkingTab() {
       </CardBlock>
 
       <CardBlock>
-        <CardBlock.Header title="External Subnets" titleId="attached-subnets-label">
+        <CardBlock.Header title="External subnets" titleId="attached-subnets-label">
```
david-crespo, line 752: pre-existing CASING CRIME

PR #3217, opened 2026-05-15
Fix editing issue with combobox values

charliepark:
This is a combobox bug that I apparently began tackling a while back, but then a bunch of other optimizations complicated the branch, and then I never wrapped it up. This branch deals with just the issue at hand.

Currently, editing combobox values is pretty broken. Here's what it looks like if I select a value, then try deleting the selection. Note how the value in the box doesn't change. It's only when selecting several characters and then deleting them that we see the options change, but even then the rendered value isn't cooperating:
Screen Recording 2026-05-14 at 3 32 54 PM mov

With this PR, editing combobox values works as expected:
Screen Recording 2026-05-14 at 3 33 53 PM mov


david-crespo:
I had a draft locally poking at this. I see yours clears the underlying selected value as soon as there's any change to the input value. In mine I preserve the selection until you either select a different value or clear the field entirely. Can't decide which is better. The robot tells me mine is more common and gives me the example of the browser URL bar, though I'm not sure the analogy is very good.

[This example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) from the WAI-ARIA docs doesn't change the value of the input text on blur, but this might only be analogous to our arbitrary values allowed combobox, not to the ones that are limited to an item in the list.

https://github.com/user-attachments/assets/427b50cf-3914-4992-ab26-ee027f08c734

charliepark:
(Last commit was before I saw your comment, so it's unrelated cleanup.)

I chatted with the robot a bit as well, and it reasoned that in the "preservation vs. clearing" question, that preservation is more useful in edit/update forms, so a pre-set value doesn't get blown away, and noted that the majority of our combobox callsites are in **creation** (so: not edit/update) forms. The one situation in the codebase it called out as "somewhat risky but with a backstop" was in the router route edit form at http://localhost:4000/projects/mock-project/vpcs/mock-vpc/routers/mock-custom-router/routes/dc2/edit, where the last item has a combobox and hitting delete would clear that field, but that because it's a required field, the user would have to select something there in order to submit the form.

Of the two paths, I'm leaning towards the "clear the value" approach (as in what's in this PR), but I don't have strong feelings on it.

charliepark:
After chatting about it a bit more, have switched it over to a "preservation" approach, and I think it's working well.

david-crespo (approved):
Thank you, works great! I am following up with a test refactor that moves the combobox e2e tests into a single file.

PR #3219, opened 2026-05-18
Tab hover animation

benjaminleonard:
Happy to be told this is gratuitous and we shouldn't do it. I'll defer to the wisdom of the crowd. It's just a fun detail. Anchor is a modern browser thing, but it just isn't shown otherwise as a fallback.

CleanShot 2026-05-18 at 15 49 27


david-crespo:
Can't say I love it. I find it a little distracting. 

Separate from that, it has a funny behavior in Firefox: notice how the tab flashes green when you go from the currently selected tab to the one next to it. 

https://github.com/user-attachments/assets/937d0d40-60e3-464a-a1bf-655ad74c61dc

benjaminleonard:
rip

PR #3221, opened 2026-05-18
Add virtualization via Headless UI's virtual prop

charliepark:
Headless UI added a `virtual` prop to easily enable virtualization, so we're using that here to virtualize our combobox lists. Scrolling in Chrome definitely feels smoother.

For manual testing, the `other-project` project has a lot of existing disks, so going to http://localhost:4000/projects/other-project/instances-new and setting the boot disk to the existing disks tab will get you a dropdown with a lot of disks


david-crespo:
Tested it in Firefox on the snapshot create form by making all 1000 disks in `other-project` snapshottable (see diff below). HUGE difference. With this change, the combobox opens instantly and filters instantly. Without it, the list takes a good second or two to open and just as long to filter when you type. Very nice.

The only issue I ran into is almost definitely fine: when I scroll as fast as I can through the list, it occasionally hits a wall and stops scrolling, presumably because it's hitting the limit of what is currently rendered. But I am scrolling really, really fast to get this.

```diff
diff --git a/app/api/util.ts b/app/api/util.ts
index f3091f865c..e0c437df75 100644
--- a/app/api/util.ts
+++ b/app/api/util.ts
@@ -225,6 +225,7 @@
   | Pick<Disk, 'state' | 'diskType' | 'readOnly'>
   | { state: DiskState; disk_type: DiskType; read_only: boolean }
 const canSnapshot = (d: SnapshotDisk) => {
+  return true
   const diskType = 'diskType' in d ? d.diskType : d.disk_type
   const readOnly = 'readOnly' in d ? d.readOnly : d.read_only
   return (
```

app/ui/lib/Combobox.tsx
```diff
@@ -143,35 +143,52 @@ export const Combobox = ({
...
+    if (allowArbitraryValues) {
+      return {
+        value: selectedItemValue,
+        label: selectedItemValue,
+        selectedLabel: selectedItemValue,
+      }
+    }
+    return null
+  }, [items, selectedItemValue, allowArbitraryValues])
+
```
david-crespo, line 169: Could do without the `useMemo`, though I'm on the fence. In half the call sites, `items` is reconstructed on every render anyway. I tested without it and filter on type was still instantaneously, so I guess I have a slight lean toward removing it.

app/ui/lib/Combobox.tsx
```diff
@@ -143,35 +143,52 @@ export const Combobox = ({
...
-  // of HUI's `open` render prop because our handler runs before HUI's
-  // (useRender merges user props first) and the render prop can be stale.
-  // The ref stays current because onClose sets it synchronously during
-  // HUI's own keydown handler. With the stale render prop, the handler
-  // could see the combobox as closed one keydown too late, firing onEnter
-  // instead of letting HUI select — hard to notice manually but caused
-  // consistent e2e flakes in Firefox.
+  // Lets onKeyDown distinguish Enter-to-select (open) from Enter-to-submit
+  // (closed). HUI's `open` render prop is stale by one keydown in our
+  // handler ordering — caused Firefox e2e flakes.
```
david-crespo: The details here were helpful — I don't think this shortening is an improvement.

app/ui/lib/Combobox.tsx
```diff
@@ -261,40 +269,38 @@ export const Combobox = ({
...
-                  <ComboboxOption
-                    key={item.value}
-                    value={item.value}
-                    className="border-secondary relative border-b last:border-0"
-                  >
-                    {({ focus, selected }) => (
-                      // This *could* be done with data-[focus] and data-[selected] instead, but
-                      // it would be a lot more verbose. those can only be used with TW classes,
-                      // not our .is-selected and .is-highlighted, so we'd have to copy the pieces
-                      // of those rules one by one. Better to rely on the shared classes.
```
david-crespo, line 281: Another comment I would prefer to keep

app/ui/lib/Combobox.tsx
```diff
@@ -201,39 +218,30 @@ export const Combobox = ({
                   : 'bg-default',
                 disabled && hasError && 'border-error-secondary!'
               )}
-              // Putting the inputRef on the div makes it so the div can be focused by RHF when there's an error.
-              // We want to focus on the div (rather than the input) so the combobox doesn't open automatically
-              // and obscure the error message.
+              // ref on the wrapper, not the input, so RHF can focus on
+              // error without opening the dropdown and hiding the message
               ref={inputRef}
-              // tabIndex=-1 is necessary to make the div focusable
```
david-crespo, line 208: don't remove

app/ui/lib/Combobox.tsx
```diff
@@ -201,39 +218,30 @@ export const Combobox = ({
                   : 'bg-default',
                 disabled && hasError && 'border-error-secondary!'
               )}
-              // Putting the inputRef on the div makes it so the div can be focused by RHF when there's an error.
-              // We want to focus on the div (rather than the input) so the combobox doesn't open automatically
-              // and obscure the error message.
+              // ref on the wrapper, not the input, so RHF can focus on
+              // error without opening the dropdown and hiding the message
```
david-crespo: This one is not that bad but the short version is probably too short

app/ui/lib/Combobox.tsx
```diff
@@ -201,39 +218,30 @@ export const Combobox = ({
...
               tabIndex={-1}
             >
               <ComboboxInput
                 id={`${id}-input`}
-                // If an option has been selected, display either the selected item's label or value.
-                // If no option has been selected yet, or the user has started editing the input, display the query.
-                // We are using value here, as opposed to Headless UI's displayValue, so we can normalize
-                // the value entered into the input (via the onChange event).
+                // Bypass HUI's displayValue so `transform` can normalize what
+                // the user types via onChange
```
david-crespo: Another one I would keep as is

app/ui/lib/Combobox.tsx
```diff
@@ -98,36 +111,23 @@ export const Combobox = ({
...
+  // form on subform submit). Only needed for allowArbitraryValues; without
+  // it, parent-driven clears leave the input still showing the old query.
   useEffect(() => {
     if (allowArbitraryValues && !selectedItemValue) {
       setQuery('')
     }
   }, [allowArbitraryValues, selectedItemValue])
 
-  // If the user has typed in a value that isn't in the list,
-  // add it as an option if `allowArbitraryValues` is true
```
david-crespo, line 130: keep it

app/ui/lib/Combobox.tsx
```diff
@@ -98,36 +111,23 @@ export const Combobox = ({
...
-  //   2. do it on all value changes, not just on clear
-  //
-  // Currently, I don't think there are any arbitraryValues=false cases where we
-  // set the value from outside. There is an arbitraryvalues=true case where we
-  // setValue to something other than empty string, but we don't need the
-  // sync because that setValue is done in onInputChange and we already are
-  // doing setQuery in here along with it.
+  // Clear the query when the parent clears the value (e.g. firewall rules
+  // form on subform submit). Only needed for allowArbitraryValues; without
+  // it, parent-driven clears leave the input still showing the old query.
```
david-crespo: The comment was a bit long but I find these things helpful later when I'm fighting this code

app/ui/lib/Combobox.tsx
```diff
@@ -218,22 +241,17 @@ export const Combobox = ({
...
                     : event.target.value
-                  // updates the query state as the user types, in order to filter the list of items
                   setQuery(value)
-                  // if the parent component wants to know about input changes, call the callback
                   onInputChange?.(value)
                 }}
                 onKeyDown={(e) => {
-                  // When the dropdown is open, Enter should select the
-                  // highlighted option (HUI handles this). When closed,
-                  // Enter should submit the subform via onEnter.
```
david-crespo, line 236: Without this comment I think I would wonder what is going on here.

david-crespo (approved):

PR #3226, opened 2026-05-22
Add `contact_support` on update status page

david-crespo:
New boolean `contact_support` field on update status added in https://github.com/oxidecomputer/omicron/pull/10271.

I tried it inside the properties table as `Contact support: Yes` and it felt terrible.

<details>
<summary>
Robot notes on the API logic behind <code>contact_support</code>
</summary>

[omicron#10271](https://github.com/oxidecomputer/omicron/pull/10271) adds a `contact_support: bool` field to the `system/update/status` API. It is the last piece of a minimal system health check tied to update status, intended as a stopgap until the fault management subsystem lands ([RFD 612](https://rfd.shared.oxide.computer/rfd/0612)).

## What it means

When `contact_support` is `true`, Nexus has detected one or more known conditions in the latest inventory collection (plus a few additional checks) that require Oxide support to resolve. The field collapses several sub-checks into a single boolean because none of the individual conditions are actionable by the customer — the only action is to call support. The detailed breakdown is logged server-side and lands in support bundles.

The intended usage maps to two cases:

- **Before an update**: if `contact_support` is true, the customer should not start an update — resolve the issue with support first.
- **After an update**: if `contact_support` is true, something went wrong; the customer should call support immediately.

## Conditions that trigger `contact_support: true`

- **Unhealthy zpools** — any zpool not in `online` state (e.g., degraded).
- **Enabled SMF services not online** — services that should be running but are in `maintenance`, `offline`, or `degraded`.
- **Stuck sagas** — sagas that have been running longer than ~15 minutes. (A sample of 10,000 done sagas on dogfood showed only 3 exceeded 15 minutes from creation to completion.)
- **Stale inventory collection** — no recent inventory collection (~15 min threshold), meaning Nexus has lost visibility into rack state.
- **Stalled update** — an update is supposed to be in progress but the planner hasn't taken a step in ~30 minutes.

The list is explicitly minimal and not exhaustive — `contact_support: false` does not guarantee the system is fully healthy.

## Suppression during an active update

Health checks often fail transiently during an update, so the API suppresses `contact_support: true` while an update is genuinely in progress. The field only surfaces a true value when either (1) there is no update in progress, or (2) an in-progress update has stalled past the threshold (matching the [10–15 minute guidance](https://github.com/oxidecomputer/omicron/blob/main/docs/reconfigurator-ops-guide.adoc#debug-stuck)) in the Reconfigurator Ops Guide for when support considers an update stuck).

In practice this means the field always presents in one of two contexts: the system is idle (pre-update or post-update), or the update has stalled long enough that the result is no longer a transient artifact.

</details>

## Issues to resolve

- Explain the situation without overdoing it
- Tooltip looks terrible in message box, what if we link to docs instead
- Should probably link to a way to actually contact support, probably the support email that goes to Zendesk

image

### DIFF
--- a/mock-api/system-update.ts
+++ b/mock-api/system-update.ts
@@ -36,7 +36,7 @@ export const updateStatus: Json<UpdateStatus> = {
     '17.0.0': 12,
     '16.0.0': 5,
   },
-  contact_support: false,
+  contact_support: true,
   suspended: false,
   target_release: {
     version: '17.0.0',
--- a/app/pages/system/UpdatePage.tsx
+++ b/app/pages/system/UpdatePage.tsx
@@ -203,9 +203,8 @@ export default function UpdatePage() {
           content={
             <>
               The system has detected one or more known conditions that require Oxide
-              support to resolve. Contact support (
-              <Link to="mailto:support@oxide.computer">support@oxide.computer</Link>) before
-              starting any update, and immediately if an update has recently completed.
+              support to resolve. Do not start an update until you contact support at{' '}
+              <Link to="mailto:support@oxide.computer">support@oxide.computer</Link>.
             </>
           }
           cta={{
--- a/app/pages/system/UpdatePage.tsx
+++ b/app/pages/system/UpdatePage.tsx
@@ -38,6 +38,7 @@ import { CardBlock } from '~/ui/lib/CardBlock'
 import { DateTime } from '~/ui/lib/DateTime'
 import { Divider } from '~/ui/lib/Divider'
 import * as DropdownMenu from '~/ui/lib/DropdownMenu'
+import { Message } from '~/ui/lib/Message'
 import { PageHeader, PageTitle } from '~/ui/lib/PageHeader'
 import { PropertiesTable } from '~/ui/lib/PropertiesTable'
 import { TipIcon } from '~/ui/lib/TipIcon'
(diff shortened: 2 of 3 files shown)


karencfv (commented):
Oooohhh that looks so pretty! Not sure if it would be necessary or not, but this boolean is also intended to get people to not perform an update if they are about to when contact_support is true, would it make sense to say something to that effect? Like, "if you're thinking about updating now, just don't"

just-be-dev:
> would it make sense to say something to that effect? Like, "if you're thinking about updating now, just don't"

I think generally you'd be better off making illegal states unrepresentable. E.g. just don't give them the ability to update. If you do something like disabling the update button (I don't know what's actually there, haven't looked at that), just make sure you note _why_ it's disabled (which this does, but you'd still want treatment on the disabled experience). 

Also, hi.

karencfv:
hi @just-be-dev!!!

> I think generally you'd be better off making illegal states unrepresentable. E.g. just don't give them the ability to update. If you do something like disabling the update button (I don't know what's actually there, haven't looked at that), just make sure you note why it's disabled (which this does, but you'd still want treatment on the disabled experience).

lol, we had a huge conversation about this 😄 https://github.com/oxidecomputer/omicron/issues/9419#issuecomment-3969822607 . We don't want to block an update just yet, the purpose of this field is more or less a warning. A simple example would be that there is a bug somewhere where the only way to fix it is to update the system. We don't want to block people from updating in that instance.

david-crespo:
Final-ish version. Talked to support a bit about the wording. On one hand if the most important thing is "don't update before talking to us", we should say that. On the other hand it's hard to say it in a way that a) doesn't sound weird, and b) doesn't feel like it's giving them permission to ignore as long as they're not doing an update. "Contact support immediate, and definitely do not start an update." I think I want to try the minimal version and see how it goes.

image

mock-api/system-update.ts
```diff
@@ -36,7 +36,7 @@ export const updateStatus: Json<UpdateStatus> = {
     '17.0.0': 12,
     '16.0.0': 5,
   },
-  contact_support: false,
+  contact_support: true,
```
benjaminleonard, line 39: Eventually I think we should figure out some way to toggle some of these properties in the dev environment so we're seeing screens like this in different scenarios. Low priority in this instance.

benjaminleonard (approved):

mock-api/system-update.ts
david-crespo: Yeah, this is a rare case of a fleet-level thing that we can't test by having different values in different silos. We could do something with the test user cookie where some users are in different fleets altogether.

karencfv:
Copied from the Matrix chat:

> We don't really want customers to see this as a health status API or become dependant on it (that's what FM will be for!). Some of the checks are only relevant if thinking about doing an update. For example the contact support checks could return true if a MUPdate is happening. Or, if a sled is newly commissioned, contact_support might be true for a short period of time. As this isn't a full fault management implementation, the contact_support boolean could present false positives. It's purpose is meant for Update purposes only, as a spot check after performing an update (will need to give it some 10 minutes after-ish) or to know if the system is in a state where performing an update could be destructive. That said, if support is OK with this and would like to be contacted whenever the boolean is true that's cool too! Just want to explain the intent of the field a little more

PR #3227, opened 2026-05-22
Display API error messages during image upload

### LINKED ISSUE #1859: Update error message when attempting to upload an image

david-crespo:
Closes #1859. Will merge this into main and then merge main into #3222.

image

### DIFF
--- a/mock-api/msw/util.ts
+++ b/mock-api/msw/util.ts
@@ -113,8 +113,16 @@ export const NotImplemented = () => {
 export const invalidRequest = (message: string) =>
   json({ error_code: 'InvalidRequest', message }, { status: 400 })
 
-export const internalError = (message: string) =>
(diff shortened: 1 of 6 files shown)


mock-api/msw/util.ts
```diff
@@ -113,8 +113,16 @@ export const NotImplemented = () => {
 export const invalidRequest = (message: string) =>
   json({ error_code: 'InvalidRequest', message }, { status: 400 })
 
-export const internalError = (message: string) =>
-  json({ error_code: 'InternalError', message }, { status: 500 })
```
david-crespo, line 117: Turns out `InternalError` was wrong, and we need it to be correct because we're adding client logic in this PR that is conditional on `error_code !== 'Internal'`.

PR #3235, opened 2026-06-02
Instance jumbo frames setting

david-crespo:
WIP to get a feel for it. See https://github.com/oxidecomputer/omicron/pull/10472. It's pretty good as-is — the worst part is that we don't currently have a way of fetching the system-level jumbo frames setting, which would let us hide this setting on the instance page if it was off system-wide (probably true for the majority of users). It feels bad to have a setting on the page that will never work, but at least the copy makes it reasonably clear. We could either add a silo-level network settings endpoint that makes this setting available to end users OR cheat by making the system-level endpoint readable by any logged-in user.

image


benjaminleonard:
Rather than making the system endpoint accessible, can we add some sort of environment endpoint that plumbs it through for developers also.

One other example could be the `cpu_platform`. Perhaps we expose if Turin or Milan are available to specify.

david-crespo:
Yeah, that’s interesting. My instinct is to avoid putting together networking and CPU settings in the same endpoint, but it would also be silly to have 3 different endpoints with one setting each.

PR #3238, opened 2026-06-03
System update: warn on set target release when `contact_support` is true

david-crespo:
Straightforward. The interesting bit here is the addition of an `msw-flags` cookie that we can use in the test to flip `contact_support: true` instead of hang it off a particular test user, which I tried and felt wrong. We can also use it for the jumbo frames setting in #3235 until we have an operator UI for flipping the setting (which we should probably have?).

## `contact_support: false`

image

## `contact_support: true`

image


charliepark (approved):
Neat on the flag test. Looks good.

PR #3242, opened 2026-06-05
Add a button to generate an instance.json body the CLI can use

charliepark:
Just an experiment. Inspired by @sion42x's demo today, this generates a JSON body the CLI can use, after the user has gone through the UI to enter in the appropriate info in the instance create form.

I'm not sure that it makes sense to add this to the UI, but it was easy to whip up and I thought it might be worth chatting through.

 
Screenshot 2026-06-05 at 12 53 38 PM
Screenshot 2026-06-05 at 12 53 47 PM


charliepark:
Ah, I thought I was building this off main, but apparently not. The only commit to pay attention to so far is [b66dd70](https://github.com/oxidecomputer/console/pull/3242/commits/b66dd707fb86e922c57deed4841da88b1ccaa709)

charliepark:
Closing this, in favor of the cleaner #3243

PR #3243, opened 2026-06-05
Create a json body the CLI can use

charliepark:
Just an experiment. Inspired by @sion42x's demo today, this generates a JSON body the CLI can use, after the user has gone through the UI to enter in the appropriate info in the instance create form.

I'm not sure that it makes sense to add this to the UI, but it was easy to whip up and I thought it might be worth chatting through.

Screenshot 2026-06-05 at 12 53 38 PM
Screenshot 2026-06-05 at 12 53 47 PM

(This is a replacement for #3242, which I accidentally built off another branch.)

### DIFF
--- a/app/api/util.ts
+++ b/app/api/util.ts
@@ -26,6 +26,8 @@ import type {
   VpcFirewallRuleUpdate,
 } from './__generated__/Api'
 
+export { snakeify } from './__generated__/util'
+
 // API limits encoded in https://github.com/oxidecomputer/omicron/blob/9dd23096de93c7d6d05ea21f6323de4410060652/nexus/src/app/mod.rs#L142
 
 // These are not actually used in app code, just the mock server. In the app we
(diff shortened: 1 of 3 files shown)


sudomateo:
Added myself primarily to give overall feedback rather than code review.

sudomateo:
I really like the idea of this, thanks for spiking on it!

It reminds me of Azure's functionality to [export ARM templates](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/export-template-portal) for resources which is really helpful when you're looking to automate things.

I see this was wired into the instance create form. I think it would be great to consider wiring this into the overflow menu for existing resources. I believe it's a more common use case to copy the definition for an existing resource rather than fill out a form and then copy it. People often create things via the web console and then, once it's working, look to automate it for the future.

I know we have clone functionality for some resources but I often go into the web console and then the API documentation to look up the schema for an existing resource just to create a request body and modify it to my liking. It would be great if I could just "Copy API Definition" for an existing resource and then just modify it.

charliepark:
Current state of things:
Screenshot 2026-09-22 at 10 14 03 AM

charliepark:
If we decide to add this, probably good to wait on the shiki integration, mentioned in one or two other issues / PRs.

PR #3257, opened 2026-06-18
Upgrade tsgo, oxfmt, oxlint

david-crespo:
Some new rules showed up in the recommended settings, so I had to add some exceptions. They are well-explained and well-justified.


app/components/form/ModalForm.tsx
```diff
@@ -39,7 +39,7 @@ export function ModalForm<TFieldValues extends FieldValues>({
   submitError,
   title,
   onSubmit,
-  submitLabel = 'Save',
```
david-crespo, line 42: These fallbacks were removed because the type was not actually optional, so the fallback could never be used.

Issue #3124, opened 2026-03-13
Explain 4007 user on fleet access page


david-crespo:
The [test privileged user](https://github.com/oxidecomputer/omicron/blob/b0d1a3c25dde8ef86469d9f5374b62fdb29c7935/nexus/db-fixed-data/src/silo_user.rs#L12-L25) shows up by default in production systems and is used by support (indirectly — not exactly for authentication or authorization, as far as I know), so it would be inconvenient for it to be deleted. We would like to get rid of it (https://github.com/oxidecomputer/omicron/issues/2305) but in the meantime we can do some harm reduction in the UI to explain what it is. The ID is well-known, so we can just detect it and give it a special name and a tooltip.

<img width="1189" height="239" alt="Image" src="https://github.com/user-attachments/assets/e7b26a4f-26c1-4e0a-bae5-e76ac1ed65e5" />

PR #3259, opened 2026-06-23
Add explanation for built-in user

fakemonster:
This PR tries to enrich the "name" field on users/groups a little bit to distinguish between the two reasons you'd find a nameless user:

1. you're in a different silo from them and thus can't see their name (have I got that right?)
2. It's the built-in test user as described in the issue [this PR attempts to resolve](https://github.com/oxidecomputer/console/issues/3124)

Screenshot_20260623_142612

### DIFF
--- a/app/api/roles.ts
+++ b/app/api/roles.ts
@@ -13,9 +13,11 @@
  */
 import { useMemo } from 'react'
 import * as R from 'remeda'
+import { match } from 'ts-pattern'
 
 import type { FleetRole, IdentityType, ProjectRole, SiloRole } from './__generated__/Api'
 import { api, q, usePrefetchedQuery } from './client'
+import { USER_TEST_PRIVILEGED_ID } from './util'
 
 /**
  * Union of all the specific roles, which used to all be the same until we added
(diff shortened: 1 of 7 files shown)


app/api/roles.ts
```diff
@@ -140,7 +168,7 @@ type SortableUserRow = { identityType: IdentityType; name: string }
 export function byGroupThenName(a: SortableUserRow, b: SortableUserRow) {
   const aGroup = Number(a.identityType === 'silo_group')
   const bGroup = Number(b.identityType === 'silo_group')
-  return bGroup - aGroup || a.name.localeCompare(b.name)
+  return bGroup - aGroup || displayUserName(a.name).localeCompare(displayUserName(b.name))
```
fakemonster, line 171: we could use this opportunity to push uuid & built-in user names to the bottom (or top?) but for now i figured i'd leave it be

app/pages/system/FleetAccessPage.tsx
```diff
@@ -151,20 +153,27 @@ export default function FleetAccessPage() {
...
+                  .with({ kind: 'stranger' }, () => (
+                    <TipIcon>
+                      Can't resolve name because{' '}
+                      {row.identityType === 'silo_user' ? 'user' : 'group'} is not in your
+                      silo
+                    </TipIcon>
+                  ))
+                  .with({ kind: 'built-in' }, () => (
+                    <TipIcon>
+                      This is a system account for bootstrapping before configuring your
```
fakemonster, line 170: my "day 2 at this company" understanding of the test user!

david-crespo:
Good start. We'll need to work on the copy. As of https://github.com/oxidecomputer/omicron/pull/10064, I think there should be no deleterious effect of the user deleting the role assignment on this built-in user, so (assuming we confirm that) we may actually want to encourage them to do that.

fakemonster:
current state is that i'm pretty darn confident you can delete the role assignment, but that a nexus restart is just going to reintroduce it, so the message isn't too useful, unless we change nexus to stop populating it on startup. although, perhaps that's not reasonable to do until built-in silo users are fully unnecessary, since it seems quite possible that even after setting up a silo with an idp, you could remove it and once again need the built-in user?

PR #3260, opened 2026-06-23
Add VPC ID and IPv6 prefix data to VPC page table

charliepark:
I was chatting with @sudomateo about a few improvements to the console, and one request was that we add the VPC ID to the header table, as it isn't currently available from the VPC page (and going back to the index page to grab it from the overflow menu is awkward). Second, as we get more IPv6 networking info into console, adding the IPv6 prefix to the header could be useful, and keeps the number of rows balanced. I am open to a different layout of the rows (are the timestamps useful? could remove?), but I think this works.

Before:
Screenshot 2026-06-23 at 4 00 43 PM
After:
Screenshot 2026-06-23 at 3 30 09 PM

In working on this I noticed that the logic we had in place for rows with a CopyToClipboard button had an early return that meant that the "copy" button would only show up when the truncation was in effect, even though it's a useful affordance even when the truncation isn't in play. So I fixed that so it shows up whenever the callsite includes the `hasCopyButton` prop set.

### DIFF
--- a/app/pages/settings/ProfilePage.tsx
+++ b/app/pages/settings/ProfilePage.tsx
@@ -44,7 +44,7 @@ export default function ProfilePage() {
 
       <PropertiesTable className="-mt-8 mb-8">
         <PropertiesTable.Row label="Display name">{me.displayName}</PropertiesTable.Row>
-        <PropertiesTable.CopyableRow label="User ID" text={me.id} />
+        <PropertiesTable.IdRow label="User ID" id={me.id} />
       </PropertiesTable>
 
       <TableTitle id="groups-label" className="mb-4">
--- a/app/pages/settings/ProfilePage.tsx
+++ b/app/pages/settings/ProfilePage.tsx
@@ -13,6 +13,7 @@ import { Settings24Icon } from '@oxide/design-system/icons/react'
 import { useCurrentUser } from '~/hooks/use-current-user'
 import { getActionsCol } from '~/table/columns/action-col'
 import { Table } from '~/table/Table'
+import { CopyToClipboard } from '~/ui/lib/CopyToClipboard'
 import { PageHeader, PageTitle } from '~/ui/lib/PageHeader'
 import { PropertiesTable } from '~/ui/lib/PropertiesTable'
 import { TableTitle } from '~/ui/lib/Table'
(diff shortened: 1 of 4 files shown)


sudomateo (commented):
Thank you for getting the ID in there! It's looking good to me but I'll defer to Crespo and/or Ben.

app/pages/settings/ProfilePage.tsx
```diff
@@ -45,10 +44,7 @@ export default function ProfilePage() {
 
       <PropertiesTable className="-mt-8 mb-8">
         <PropertiesTable.Row label="Display name">{me.displayName}</PropertiesTable.Row>
-        <PropertiesTable.Row label="User ID">
-          {me.id}
-          <CopyToClipboard className="ml-1" text={me.id} />
-        </PropertiesTable.Row>
+        <PropertiesTable.CopyableRow label="User ID" text={me.id} />
```
charliepark: Technically, this should probably be an IdRow, but it's rendered in a full-width cell, so truncation isn't needed, and we currently don't show the copy-to-clipboard button in IdRows when the cell text isn't truncated. So this wouldn't have a copy button. @fakemonster points out that we might want to talk through the truncation rules and "when do we show the copy button?" logic on the IdRow component, but we can do that outside of this PR.

app/ui/lib/Truncate.tsx
```diff
@@ -26,22 +26,28 @@ export const Truncate = ({
   hasCopyButton,
   tooltipDelay = 300,
 }: TruncateProps) => {
-  if (text.length <= maxLength) {
-    return <div>{text}</div>
-  }
```
charliepark, line 31: Removed this early return as it meant we weren't showing the CopyToClipboard button consistently

fakemonster:
i know i was pushing for changes here, but if we've got other plans for truncation that might obviate whatever we do here, maybe we back out of the last three commits (middle truncate, IdRow changes, character limit change)? since we're mixing js and css truncation, and the js truncation is used to decide to show tooltips, we actually need the forced truncation at 32 chars so that we still show the tooltip when there's css truncation (and then of course, if there's css truncation, we don't want to middle truncate!)

fakemonster (approved):

david-crespo:
Looks good, my only concern was whether the placement of the ID matches other pages.

charliepark:
IDs tend to be in the last row; there are examples on both the left side and the right side. Datetime cells (created / modified) tend to be in the right-hand column. I added notes on those to #3268, in case we want to align them across resource pages.

david-crespo (approved):

PR #3263, opened 2026-06-25
Update copy and logic on project access dropdowns

### LINKED ISSUE #2963: Misleading disabled tooltip on change role button
Just noticed this really bad message on project access when the role comes from the silo. 

https://github.com/oxidecomputer/console/blob/ab76183153337c509bcde7007184cc655c43df71/app/pages/project/access/ProjectAccessPage.tsx#L165-L170

Image

It's wrong in multiple ways:

* The current user may well be a silo admin who does in fact have permission to change that role
* Adding a project role is possible even if the user is only a project admin and not a silo admin
  Image

The reason it's disabled is because conceptually, in order to modify that actor's role on the project, you have to add a new role assignment on the project, not modify the existing one on the silo. We could do a lot better in this UI to make that clear.

charliepark:
This PR makes a few changes to the tooltips, buttons, and logic of the Project Access table and each row's action menu dropdown.

First, it updates the logic around whether the user actually has the permissions to make changes to the (users' / groups') permissions. If the user doesn't have the appropriate permissions, the buttons are disabled. Previously, the logic was "does the user/group being investigated currently have a project-level permission?", which doesn't quite cover the same territory and is less useful. So this PR adds a `canEditRoles` attribute to check against the current user and whether they can make changes here.

Second, this PR changes the verbiage of the dropdown button. For a user/group *with* a current project role, the menu now says "Change project role". For a user/group *without* a project role, it now says "Add project role". Screenshot 2026-06-25 at 3 02 20 PM Screenshot 2026-06-25 at 3 02 05 PM

In a case where a user has a silo role that supersedes a project role (`SILO.ADMIN`, `PROJECT.COLLABORATOR`), the dominant role (`SILO.ADMIN`) will still render on the table, but the presence of the project role will set the copy to say "Change project role": Screenshot 2026-06-25 at 3 24 47 PM

(Note: The "project" in "Change/Add project role" is new, but I think an improvement to reinforce the tier of access the admin is analyzing. We might update "Delete" to say "Delete project role" to help in the same way. I'm open to thoughts on that! Similarly, we could change the dropdowns on the Fleet/Silo Access pages to say "fleet role" / "silo role")

Third, when the role has been inherited from the silo (meaning: no project role), there's no way to delete the role from the current view, so the Delete button is disabled and the tooltip now says "This role is inherited from the silo" Screenshot 2026-06-25 at 3 02 39 PM

Fourth, a subtle change in the `change project role` form … if the user/group *has* a project role, that role's radio button is pre-selected in the form (which is unchanged from the current behavior). When the admin is *adding* a role, meaning there is not currently a project role, no radio buttons are pre-selected. (changing a role) Screenshot 2026-06-25 at 4 01 02 PM (adding a role) Screenshot 2026-06-25 at 4 00 53 PM I'm not sure I love the duplication of "project role" in the header and at the top of the radio buttons.

Closes oxidecomputer/console#2963


david-crespo:
This is a good direction. I just noticed "Delete" is also problematic — you might read it as "delete user" and be afraid to click it. "Delete role assignment" is too long, but it would be cool if we could come up with something short that conveys that. "Delete assignment" is ok, though "assignment" isn't likely the user's name for it. "Unassign" is pretty ugly but might get the job done? We could leave it as "Delete" and explain on hover, but I don't like that either.

benjaminleonard:
- "Remove" I guess a synonym for delete but feels different
- "Remove access"?

charliepark:
Since "Add project role" and "Change project role" seem good, what about just using "Remove project role"?

charliepark:
Screenshot 2026-07-02 at 2 17 58 PM

david-crespo:
Great!

charliepark:
This had a bunch of overlap with #3211, so I've moved this code over there. That one is getting pretty big, though, so I'll try to wrap it up soon. Won't close this yet, but also will move active development over to #3211.

charliepark:
Going to close this, as #3211 is the more active branch.

PR #3265, opened 2026-06-26
Fix an e2e flake, maybe: use `fillNumberInput`

david-crespo:
Saw a flake in https://github.com/oxidecomputer/console/actions/runs/28251857097/job/83705140277, sicced the robot on it. Unfortunately it was not able to repro the flake locally, but it seems plausible anyway.

### 🤖 Summary

**Flake:** `firewall-rules.e2e.ts › can create firewall rule` (firefox)

**Root cause.** The test typed `5` into the Priority field, but the rule was created with `Priority: "0"` (the default) — failing all 3 retries with:

```
Expected: ...{"Priority": "5", ...}
Received: ...{"Priority": "0", ...}
```

The Priority field is a React Aria NumberInput, which can drop a value set via plain `page.fill()` when the form re-renders before the edit commits. The codebase already documents this exact race and has a helper for it: `fillNumberInput` (test/e2e/utils.ts:68), which clicks, fills, blurs, and polls until the value sticks.

The old call also used the deprecated legacy string-selector syntax (`'role=textbox[name="Priority"]'`) that CLAUDE.md discourages.

**Fix.** Swapped both raw Priority fills in the file to use `fillNumberInput`:
- line 44 (`can create firewall rule`, the one that flaked) — `'5'`
- line 586 (`name conflict error on edit`, same NumberInput risk) — `'37'`

**Verification.** The flake didn't reproduce locally in 15 runs (it needs CI's heavier parallel load), but the error is unambiguous and the fix matches the established pattern. After the change, both affected tests passed 20/20 on firefox; lint clean.

The fix is committed as its own rev (`b8c7d432`) on top of main for you to review/squash.

### DIFF
--- a/test/e2e/firewall-rules.e2e.ts
+++ b/test/e2e/firewall-rules.e2e.ts
@@ -11,6 +11,7 @@ import {
   expect,
   expectRowVisible,
   fillAndSelectComboboxOption,
+  fillNumberInput,
   selectOption,
   test,
   type Locator,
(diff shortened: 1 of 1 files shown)


charliepark (approved):
Seems reasonable. We might also do a `'role=textbox[name="…` → `page.getByRole…` migration (separately) in the rest of the codebase once this is in. Or migrating other fillNumberInputs if we think other tests might be flakey.

PR #3267, opened 2026-06-29
feat: checkbox to make pool default when linking to silo

sudomateo:
Added a checkbox to the IP pool and subnet pool linking modals to make the pool the default pool for the silo.

Closes https://github.com/oxidecomputer/customer-support/issues/413.

Amp-Thread: https://ampcode.com/threads/T-019f1129-f325-73e9-96da-94a8a24fa617

image

image

### DIFF
--- a/app/pages/system/silos/SiloIpPoolsTab.tsx
+++ b/app/pages/system/silos/SiloIpPoolsTab.tsx
@@ -273,28 +273,61 @@ function LinkPoolModal({ onDismiss }: { onDismiss: () => void }) {
   const { silo } = useSiloSelector()
   const { control, handleSubmit } = useForm({ defaultValues })
 
+  function invalidate() {
+    queryClient.invalidateEndpoint('siloIpPoolList')
+    queryClient.invalidateEndpoint('systemIpPoolSiloList')
+  }
+
   const linkPool = useApiMutation(api.systemIpPoolSiloLink, {
-    onSuccess() {
-      queryClient.invalidateEndpoint('siloIpPoolList')
-      queryClient.invalidateEndpoint('systemIpPoolSiloList')
-      onDismiss()
-    },
+    onSuccess: invalidate,
     onError(err) {
       addToast({ title: 'Could not link pool', content: err.message, variant: 'error' })
     },
   })
+  // Promoting to default is a separate request from linking. The API rejects
+  // linking *as default* when the silo already has a default for the pool's
+  // version+type, but the update endpoint demotes the existing default
+  // automatically. So "link as default" = link non-default, then promote.
+  const promotePool = useApiMutation(api.systemIpPoolSiloUpdate, { onSuccess: invalidate })
 
-  function onSubmit({ pool, isDefault }: LinkPoolFormValues) {
-    if (!pool) return // can't happen, silo is required
-    linkPool.mutate({ path: { pool }, body: { silo, isDefault } })
+  async function onSubmit({ pool, isDefault }: LinkPoolFormValues) {
+    if (!pool) return // can't happen, pool is required
+    try {
+      await linkPool.mutateAsync({ path: { pool }, body: { silo, isDefault: false } })
+    } catch {
+      return // onError already toasted; leave the modal open to retry
+    }
+    if (isDefault) {
+      try {
+        await promotePool.mutateAsync({ path: { silo, pool }, body: { isDefault: true } })
+      } catch {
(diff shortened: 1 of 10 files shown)


david-crespo:
Looks good, only thing to change might be having the message on default IP pool reflect the type of the pool, since you can have a default for each IP version. "default IPv4 pool for silo" etc. In theory we could also say "unicast" but since you can have a default each for unicast and multicast too, but it's borderline.

Also the checkbox should be tested in a playwright test.

sudomateo:
> Looks good, only thing to change might be having the message on default IP pool reflect the type of the pool, since you can have a default for each IP version. "default IPv4 pool for silo" etc. In theory we could also say "unicast" but since you can have a default each for unicast and multicast too, but it's borderline.
> 
> Also the checkbox should be tested in a playwright test.

Done in https://github.com/oxidecomputer/console/compare/15d66cca370b6a3b91b30207d011c2121fdb95a2..fbb52ee2d7e41a7e16699b34d1255eb49f035bc1.

app/pages/system/silos/SiloIpPoolsTab.tsx
```diff
@@ -334,6 +340,12 @@ function LinkPoolModal({ onDismiss }: { onDismiss: () => void }) {
               required
               control={control}
             />
+
+            <CheckboxField name="isDefault" control={control}>
+              {selectedPool
+                ? `Make default IP${selectedPool.ipVersion} ${selectedPool.poolType} pool for silo`
+                : 'Make default pool for silo'}
+            </CheckboxField>
```
david-crespo, line 362: Ah, funny. Forgot you don't have a version until you've selected a pool. Oh well!

sudomateo: Yeah I don't love that the text is different until you choose a pool but I think the selected pool text is helpful.

sudomateo:
Pushed a `npm run fmt` fix in https://github.com/oxidecomputer/console/compare/fbb52ee2d7e41a7e16699b34d1255eb49f035bc1..4b2a4839da370f76c3deba9e35942e8fdc3d772c

david-crespo:
I went way overboard but the result is really good. You can see it in the commit list, but here is what happened:

The checkbox was only added to the link IP pool modal on the silo IP pools page, but was missing from the IP pool silos page, so I added it to the latter. Then I noticed the mock server had some out of date logic around what happens when you try to link an IP pool and set it default at the same time. What happens is: if there is already a default, it will error if you try to link and set default at the same time. You are supposed to first link it, and then set it as default. This is to make it a little harder to overwrite the existing default in a single API call. I brought the mock server up to date.

In doing that, I realized the user of the web console would be very annoyed if they checked the box and we errored out, telling them to do something we could have just done for them. There is no reason the web console has to make a single call for every action! By checking the box, the user has indicated that they want to make the thing default, and we (being a Web Application) can actually inform them right there of the consequences of that action. So the result of all that is this. I like it! 

image

When the user checks the box, we link the thing as non-default and then we make a second call to make it the default. The only thing I don't really like about it is that we're doing all the error handling through toasts: if the initial link fails (rare), we keep the modal open and there's a toast saying it failed (this is how it already worked before this change). If the link succeeds and the make default fails, we have already closed the modal but the toast tells them to try again through the row action. It's not that bad. Maybe in an ideal world we'd keep the modal open in both cases and show the errors inline.

https://github.com/user-attachments/assets/2879a6b6-3a15-452b-a98a-8f465d6bec6b

https://github.com/user-attachments/assets/3fa9d6ec-b527-41e0-a06f-ab17e4e36cc8

app/pages/system/useLinkPoolSiloFlow.ts
```diff
@@ -0,0 +1,138 @@
...
+      } catch {
+        addToast({
+          title: promoteErrorTitle,
+          content: 'Use the row menu to make it the default.',
+          variant: 'error',
+        })
+      }
+    }
+    return true
+  }
```
david-crespo, line 79: We don't do very much of this kind of action sequencing in the console but it's really fine, especially since we're relying on the toasts for error handling rather than returning the error state from the hook.

sudomateo:
Muahahah, I see my nerdsnipe worked. Thanks for getting this into a shippable state!

david-crespo:
The test failure is a bug that only happens during the first hour of the month UTC lmao. Pushing a fix as a separate PR.

charliepark (approved):
Looks good! Thank you, @sudomateo (also Crespo!)!

PR #3272, opened 2026-06-30
minor: fix extra space at the bottom of combobox options

david-crespo:
Get rid of that little space at the bottom of the list. The hard-coded item height was slightly off.

image


david-crespo:
Looking into what this means for the taller rows. I think it's fine because `ITEM_HEIGHT` seems to act more like a minimum.

image

PR #3273, opened 2026-07-01
Fix for weird combobox panel flash

david-crespo:
We're running into https://github.com/tailwindlabs/headlessui/issues/3834, which is so pervasive it even afflicts Headless UI's own combobox docs. Basically when the combobox renders initially it doesn't know what width it should, so I think it renders with a tiny width and is very tall. Then it snaps into shape. The videos in that issue match exactly what I see when I open a combobox.

The fix is small and it works, but I'm marking it as a draft because I haven't thought about it very much.


david-crespo:
Weirdly, couldn't repro the issue now. Will have to try harder.

PR #3275, opened 2026-07-01
Tell password managers to ignore first and last and client ID

david-crespo:
This makes no sense!

image


app/forms/idp/create.tsx
```diff
@@ -168,6 +168,7 @@ export default function CreateIdpSideModalForm() {
         label="Technical contact email"
         required
         control={form.control}
+        {...noPasswordManager}
```
david-crespo, line 171: This one is an email, so it could conceivably come from a password manager, but it seems unlikely.

PR #3277, opened 2026-07-02
Refactor value + unit renderings to use unified layout

### LINKED ISSUE #3276: refactor to use existing sizeCellInner function more
As I was running #3214 through a few checks and the robot noticed that several locations in the codebase use variations of the code in `sizeCellInner` to render stylized "size + unit" text on the screen.

```
The pattern formatBytes(x) → {value} <span className="text-tertiary">{unit}</span> is hand-rolled in several places rather than going through sizeCellInner:

- InstanceResourceCell.tsx:22 — {memory.value} <span className="text-tertiary">{memory.unit}</span>. This is byte-identical to sizeCellInner's output and could just be sizeCellInner(value.memory). The one exact duplicate.
- InstancesPage.tsx:117-120 (memory column) — same shape but the unit span adds ml-1 on top of the literal space, so it's a double gap. Near-duplicate, not identical.
- InstancePage.tsx:222-224 (ram) — split into a text-default span plus a text-tertiary ml-1 span with a leading space.
- SledPage.tsx:78-79 (usable physical ram) — text-default pr-0.5 + text-tertiary.
- SiloQuotasTab.tsx:35 — its own Unit = classed.span\ml-1 text-secondary`(note:text-secondary, not text-tertiary`).
```

I haven't verified each of these locations, but it seems like we could clean them up pretty easily.

Also note that "sizeCellInner" might make more sense with a different name (and possibly located elsewhere) once these various places in the UI are updated to use the common component / helper.

charliepark:
When I was working on #3214, I came across a number of occasions where we rendered a value and a unit, like `6 vCPUs` or `128 GiB`. We had a few different ways of rendering those. This PR creates a new component, `<ValueUnit>`, which renders the value, formatted, and the unit, also formatted. It also adds a `<Size>` component, built on top of `<ValueUnit>`.

Currently, the files changed in this branch do NOT step on the toes of https://github.com/oxidecomputer/console/compare/main...standardize-unit-helpers, which is addressing some of the notes in https://gist.github.com/david-crespo/5fe725886475d18afa6cc459daf1acc3. Nevertheless, it might be best for that one to get wrapped up and merged first.

The screenshots aren't actually that exciting, because the goal here isn't really to change anything visually, but just to give context, look at the CPU / Memory columns:

Before:
Screenshot 2026-07-14 at 1 12 15 PM

After:
Screenshot 2026-07-14 at 1 11 58 PM

Closes #3276


app/pages/system/inventory/sled/SledPage.tsx
```diff
@@ -74,10 +71,10 @@ export default function SledPage() {
...
           <span className="text-default">{sled.rackId}</span>
         </PropertiesTable.Row>
-        <PropertiesTable.Row label="usable physical ram">
-          <span className="text-default pr-0.5">{ram.value}</span>
-          <span className="text-tertiary">{ram.unit}</span>
-        </PropertiesTable.Row>
+        <PropertiesTable.SizeRow
+          label="usable physical ram"
+          bytes={sled.usablePhysicalRam}
+        />
```
david-crespo, line 77: Nice. The unit label was too close before.

Image

Image

david-crespo (approved):

Issue #2647, opened 2025-01-09
Instance resize modal submit button doesn't enable until you blur the input


david-crespo:
https://github.com/user-attachments/assets/7bad0f1f-65eb-4bb6-b646-6803fc5d1a9c

benjaminleonard:
I believe this is due to how the `react-aria` number input works. It takes the value in the input and applies it on blur. Though maybe if the submit button is enabled when the input is touched it would still work?

david-crespo:
If it's hard to fix on the input side, I lean toward leaving the button enabled and having it refuse to submit if the values haven't changed. Might as well enable it to begin with if it's going to be enabled on touch.

PR #3282, opened 2026-07-07
Fire off number updates eagerly

### LINKED ISSUE #3249: Unable to resize an image until you click/tab to another field first
When trying to resize a stopped image, I changed the RAM value and tried to click Resize. The option was disabled. You can enable it by clicking on a different field, pressing tab, etc. but it should be able to work without that.

fakemonster:
react-aria waits until a blur event to parse inputs and emit numbers. This is nice except that validation is stuck waiting for a blur event (so, e.g., all your form contents are valid, but the submit button is disabled, because we're waiting for you to blur your input). While it'd be nice for react-aria to support some kind of "eager mode", we can pretty easily write one ourselves, by forcing a `commit` whenever we get suspicious.

Resolves https://github.com/oxidecomputer/console/issues/3249 / https://github.com/oxidecomputer/console/issues/2647


app/ui/lib/NumberInput.tsx
```diff
@@ -31,6 +31,20 @@ export function NumberInput(props: NumberInputProps) {
...
 
+  // react-aria only fires props.onChange on commit (blur / Enter / stepper),
+  // but we want form state to update as soon as it would produce a different
+  // field value. Committing whenever state.inputValue changes to an
+  // unambiguous number lets react-aria keep controlling parsing, clamping
+  // etc., but forces it to be more eager.
+  //
+  // Context: https://github.com/adobe/react-spectrum/issues/7984
+  useEffect(() => {
+    if (String(Number(state.inputValue)) === state.inputValue || state.inputValue === '') {
```
david-crespo: Put this in a function and unit test it.

app/ui/lib/NumberInput.tsx
```diff
@@ -31,6 +31,20 @@ export function NumberInput(props: NumberInputProps) {
   const { groupProps, inputProps, incrementButtonProps, decrementButtonProps } =
     useNumberField(props, state, inputRef)
 
+  // react-aria only fires props.onChange on commit (blur / Enter / stepper),
+  // but we want form state to update as soon as it would produce a different
+  // field value. Committing whenever state.inputValue changes to an
```
benjaminleonard, line 38: Very interesting, I like it

david-crespo:
The browser mode version of the tests is basically the same, I just prefer this over the testing-library stuff. We have very little of the latter left and I'd like to phase it out.

david-crespo (approved):

PR #3284, opened 2026-07-07
Add External IPs column to Instances page

### LINKED ISSUE #2484: Show external IPs on instance table
It would require a request per instance (as currently implemented) but it's very useful info. It would be especially cool to badge the IPs with the pool they came from too: show `eng-vpn` or `public` as appropriate.

Image

charliepark:
This adds a column to the Instances table, showing an external IP for each instance. Any extra external IPs are included in a tooltip. We don't show SNAT IPs, but do list Floating IPs first, then Ephemeral IPs. The sort order is the same as the "External IPs" cell in the header of the Instance Page. We include a CopyToClipboard icon/button in the cell as well.

Closes #2484

Screenshot 2026-07-07 at 3 57 46 PM

### DIFF
--- a/app/components/ExternalIps.tsx
+++ b/app/components/ExternalIps.tsx
@@ -8,20 +8,16 @@
 
 import { useQuery } from '@tanstack/react-query'
 import { Link } from 'react-router'
-import * as R from 'remeda'
 
-import { api, q, type ExternalIp } from '@oxide/api'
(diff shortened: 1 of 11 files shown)


david-crespo:
Moved the state column left after discussion in chat. (It looks a bit better on dogfood where they're all running or stopped.)

image

PR #3289, opened 2026-07-16
prefetch VPCs/Subnets so networking tab loads without skeleton cells

### LINKED ISSUE #1731: Jarring pop-in on instance network interfaces page
On the NIC page, we have to fetch external IPs, VPC, and VPC subnet info _after_ the list of NICs comes back (because we only know what to fetch from that list). In order to avoid blocking the page with serial requests, we fetch that stuff on render and get the dreaded loading spinner and pop-in. The thing is, there can only be up to 6 or 8 NICs (I forget), and there's probably fewer, and most likely they all have the same VPC and subnet. So fetching this stuff in the loader after waiting for the list to come back is probably fine. I would certainly trade a few hundred ms to avoid this travesty.

https://github.com/oxidecomputer/console/blob/6237fcf03690479050944e2f0915252e1a05437e/app/pages/project/instances/instance/tabs/NetworkingTab.tsx#L38-L66

2023-08-10-nic-pop-in

charliepark:
Closes #1731 

Issue 1731 is pretty old, and not 100% an issue any more — some prefetching already happens, and the main benefits of this change are below the fold, so I doubt people would actually see the loading skeleton cell most of the time — but I saw the issue and figured it would be a quick one to knock out and we could decide whether to merge it or not.

Essentially, what this is doing: once the NICs have been loaded, use those values to prefetch the VPC and Subnet views, so their values can be entered into the table as it loads, sans skeleton cell.
* The primary VPC is already available to the table (unrelated to this PR), so the main VPC-related benefit here is in cases when an instance has secondary VPCs.
* The Subnet prefetching is easier to see the benefit of:

**Before:**

https://github.com/user-attachments/assets/8d7254bc-376b-484d-96e2-3dd74608359e

**After:**

https://github.com/user-attachments/assets/743202db-f7e5-4702-85ac-80efc6f0a1b0

But, again, this is generally below the fold of the page, so I think the revulsion Crespo felt with the earlier version of the Networking view is already somewhat softened.

### DIFF
--- a/app/pages/project/instances/NetworkingTab.tsx
+++ b/app/pages/project/instances/NetworkingTab.tsx
@@ -128,12 +128,29 @@ const staticSubnetCols = [
 export async function clientLoader({ params }: LoaderFunctionArgs) {
   const { project, instance } = getInstanceSelector(params)
   await Promise.all([
-    queryClient.fetchQuery(
-      q(api.instanceNetworkInterfaceList, {
-        // we want this to cover all NICs; TODO: determine actual limit?
-        query: { project, instance, limit: ALL_ISH },
-      })
-    ),
+    // Prefetch the by-ID VPC and subnet views the NIC table cells look up, so
+    // VpcNameFromId/SubnetNameFromId hit a warm cache instead of popping in
+    // after the table paints. IDs come from the NIC list, so chain off it; NICs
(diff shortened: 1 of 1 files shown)


david-crespo (approved):
I noticed the VPC cell doesn’t have a loading state in the before video. That’s because we are already prefetching the primary NIC’s VPC for the instance details at the top. The addition in this PR covers other VPCs if there are any.

PR #3290, opened 2026-07-16
Add silo description to combobox item in "link silo" modal

### LINKED ISSUE #2175: Put silo descriptions in link silo modal listbox

charliepark:
Another quick pickup. Adds a line with each silo's description. Closes #2175.

Screenshot 2026-07-16 at 3 26 25 PM


david-crespo:
Looks good. Made me wonder if we should have an abstraction for these descriptions. Found two that are not good with long contents: one broken, one just failing to truncate, allowing a long description to fill up the screen. Taking a quick look at pulling something out that makes these look more like the one you added.

attach-fip-listbox

pool-listbox-long

PR #3294, opened 2026-07-19
`ItemLabel` for combobox and listbox

david-crespo:
Followup to #3290. Initially it didn't seem like it was worth abstracting because the call sites were too different, but then it turned out the variation was actually bad, and making them all the same fixes the issue. Note the first commit just extracts `ItemDescription`, and then the second goes all the way and makes a component for the entire item label.

## Before

Broken wrapping:

attach-fip-listbox

Too long description is allowed to dominate:

pool-listbox-long

## After

Not broken:

attach-fip-listbox-after

Nice truncation:

pool-listbox-after


charliepark (approved):
Looks great!

Issue #3270, opened 2026-06-30
Initial `oxql` metrics explorer


benjaminleonard:
Want to add a system-level page that let's the user explore metrics and construct `oxql` queries.<br><br>Acting at least initially as:

1. A way to learn and experiment with `oxql`
   1. Educational to understand query shapes
   2. Discover available schemas
2. An ad-hoc troubleshooting workflow
   1. Perhaps initially a smaller set of predefined query shapes; e.g. sled CPU, sled networking, instance CPU
   2. Less concerned about pure utility here to begin with, we can always follow up to optimise troubleshooting workflows later

## UX

Probably a text-first or hybrid query surface. Might punt on a bi-directional UX unless there's a simple robust way to do it. Provide lightweight controls that write into the query rather than a fully separate visual builder.

Adding presets for queries and time-ranges.

## Visualisation

Constrained set of chart presentations selected by metric schema/type/unit. Either using the data type to infer the chart type, or just getting Claude to make a map of the schemas → chart type. Maybe we'll need to clamp the range, or reduce the points. Manual refresh rather than polling. We could persist the query state in the URL param or add a share button that generates a link even if we're not adjusting the param of the active query.<br><br>---<br><br>Will's lovely demo is a useful reference here: [https://github.com/oxidecomputer/oxql-vis](<https://github.com/oxidecomputer/oxql-vis>)

PR #3295, opened 2026-07-21
Migrate to uplot

fakemonster:
This PR replaces recharts with uPlot. The core of the decision here is really just "stop using an svg-based chart", but uPlot is a fast option even among canvas-based renderers, and (with a bit of patience) still quite flexible when it comes to styling.

Setting the code itself aside, this should not be that noticeable a change. Charts will render faster, but the smaller datasets we're using right now weren't really a problem for recharts; it would be much more noticeable once we started work on https://github.com/oxidecomputer/console/issues/3270. The three most noticeable changes today:

1. a switch from svg to canvas will, in lower dpr screens (like, 1x), make the linework more stark. I consider this a benefit for those of us with pitiful 1x displays; we get some small slice of the Apple experience.
2. the ticks are picked by uplot now. Now charts with very low numbers (say, a max of 1.5) won't get squished to the bottom, and we'll have generally "nice" numbers (5/15 minutes marks, y-axis multiples of 100, 1000, etc).
3. the top grid line has been removed. Since we give uplot control over ticks now, it seems a bit odd to stick a line there willy-nilly.

As an example of difference 2 in particular, here's the visual comparison test for instance network metrics:

|Recharts|uPlot|
|-|-|
instance-metrics-network-expected|instance-metrics-network-actual


david-crespo:
Those are some good ticks. Does this also fix the missing x-axis ticks when there’s a big gap in the data?

app/ui/styles/index.css
```diff
@@ -29,6 +29,7 @@
 @import '@oxide/design-system/styles/light.css';
 @import '@oxide/design-system/styles/preflight.css' layer(base);
 @import 'simplebar-react/dist/simplebar.min.css' layer(components);
+@import 'uplot/dist/uPlot.min.css' layer(components);
```
benjaminleonard, line 32: This seems fine but worth checking it's not overriding something we don't want it to. Like:

```
.u-legend th {
	font-weight: 600;
}
```

We're not using title or legend here but we might elsewhere.

app/components/TimeSeriesChart.tsx
```diff
@@ -58,51 +32,108 @@ function isSameDay(d1: Date, d2: Date) {
...
+const withAlpha = (color: string, alpha: number) => color.replace(/\)\s*$/, ` / ${alpha})`)
+
+// uPlot draws to a canvas, so it can't consume CSS custom properties directly. We subscribe to the
+// theme instead.
+function getChartTheme(): ChartTheme {
+  const style = getComputedStyle(document.body)
+  const v = (name: string) => style.getPropertyValue(name)
+  return {
+    fontFamily: v('--font-mono'),
+    stroke: v('--content-accent-tertiary'),
```
benjaminleonard: Was already like this, but we should probably switch to the equivalent semantic stroke token `--stroke-accent-secondary`

benjaminleonard:
Currently not working in the vercel preview, maybe related error:

`validate.ts:79  Note that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.`

image

benjaminleonard:
Unfortunately wide gamut / `display-p3` colour does not work. It looks like we can patch it (untested on regular displays) but we might want to try and upstream up a PR to `uPlot` if possible.

```tsx
// uPlot's canvas defaults to sRGB, which clips our oklch theme colors, so charts look washed out.
// uPlot exposes no way to pass context attributes, so we patch the canvas default globally.
function defaultCanvasToDisplayP3() {
  try {
    const attrs = document
      .createElement('canvas')
      .getContext('2d', { colorSpace: 'display-p3' })
      ?.getContextAttributes?.()
    if (attrs?.colorSpace !== 'display-p3') return // browser doesn't support it
  } catch {
    return
  }
  const original = HTMLCanvasElement.prototype.getContext
  HTMLCanvasElement.prototype.getContext = function (
    this: HTMLCanvasElement,
    contextId: string,
    options?: object
  ) {
    return original.call(this, contextId, { ...options, colorSpace: 'display-p3' })
    // cast needed because a plain function can't satisfy getContext's overload set
  } as typeof original
}
defaultCanvasToDisplayP3()
```

david-crespo:
I don't think the CSP warning is related, it's there on main too. I think it's from this, which would only affect the preview because I don't think we use zod in production: https://github.com/colinhacks/zod/issues/4461. Still probably worth fixing.

app/ui/styles/index.css
fakemonster: uplot seems to follow a pretty reliable convention of prefixing its classes with `u-`, so we could forbid usage of `u-*` classnames within our ts. tough to do comprehensively without introducing false positives (like, you should be able to write "u-turn" as non-class text!), but we could handle the obvious cases like `className="p-4 u-legend"` to avoid most surprise collisions

fakemonster:
> Unfortunately wide gamut / `display-p3` colour does not work. It looks like we can patch it (untested on regular displays) but we might want to try and upstream up a PR to `uPlot` if possible.

I'd rather you pushed that change just cause I can't personally verify it, but regardless of uPlot support, this patch would be fine to have. It would become a no-op (irt uPlot) if uPlot _did_ add p3 support. in terms of being safe to have laying around, the only variation you could have here is perhaps electing not to set the p3 option, to respect it if some future caller explicitly chose srgb (uPlot just isn't setting this option, so it'd still have the intended effect here)

app/components/TimeSeriesChart.tsx
fakemonster: handled in [1c8a4bc](https://github.com/oxidecomputer/console/pull/3295/commits/1c8a4bc06f581752553e81e448736daf71ba24f7)

david-crespo:
nice effect on bundle size -- `TimeSeriesChart` is 70% smaller gzipped now.

image

app/components/TimeSeriesChart.spec.tsx
```diff
@@ -70,4 +73,23 @@ describe('safe redrawing', () => {
...
+    dataPropsPassed.length = 0
+    const { rerender } = render(<TimeSeriesChart {...props((v) => `${v}%`)} data={data} />)
+    rerender(<TimeSeriesChart {...props((v) => `${v} pct`)} data={data} />)
+
+    expect(dataPropsPassed.length).toBeGreaterThan(1) // it re-rendered
+    expect(new Set(dataPropsPassed).size).toBe(1) // but every render passed the identical reference
+
+    rerender(<TimeSeriesChart {...props((v) => `${v}%`)} data={[...data]} />)
+    expect(new Set(dataPropsPassed).size).toBe(2) // unless the reference changes
+  })
```
david-crespo, line 94: Interesting kind of test. Basically it guards against removing the `useMemo`? I’m ambivalent but there’s not much downside. To test it in a way the feels less like reimplementation, you’d have to assert about whether some action causes re-renders or not, which I don’t think I’ve seen done. Kind of an interesting idea.

app/ui/styles/index.css
benjaminleonard: Oh in this case, I mean that we don't use that weight in the console. So within a legend it would likely fall back to the janky browser bold that we don't want. Again probably fine in this case but worth keeping an eye on.

benjaminleonard:
Looks great!

app/ui/styles/index.css
fakemonster: ah! i completely misunderstood. yeah it'd definitely jump up to full bold. but we do need some of these styles, so i figure we'll just keep an eye out

app/components/TimeSeriesChart.spec.tsx
fakemonster: i do think in my oncoming work this'll be more important to assert. a deep comparison could be quite a large amount of unnecessary checking in a chart with, say, 40 lines and a few thousand time points

david-crespo (approved):
🚀

PR #3300, opened 2026-07-22
Un-abstract users/groups tabs to silo-only

david-crespo:
Initially I think #3211 had users and groups at both the silo and project level and then pared it back to only have it at the silo level, but kept the abstractions used to make it work in both spots. This undoes the abstraction.

Redone from #3298 because I made that one a stack, and it turns out that prevents you from merging it into the underlying one.

### DIFF
--- a/app/api/roles.ts
+++ b/app/api/roles.ts
@@ -85,11 +85,20 @@ export function updateRole<Role extends RoleKey>(
   return { roleAssignments }
 }
 
-/** Map from identity ID to role name for quick lookup. */
+/**
+ * Map from identity ID to role name for quick lookup. If an actor has multiple
+ * assignments in the same policy (which the API allows) take the strongest one.
+ */
 export function rolesByIdFromPolicy<Role extends RoleKey>(
   policy: Policy<Role>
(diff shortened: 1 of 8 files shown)


app/api/roles.ts
```diff
@@ -85,11 +85,21 @@ export function updateRole<Role extends RoleKey>(
...
 ): Map<string, Role> {
-  return new Map(policy.roleAssignments.map((a) => [a.identityId, a.roleName]))
+  const map = new Map<string, Role>()
+  for (const { identityId, roleName } of policy.roleAssignments) {
+    const existing = map.get(identityId)
+    if (!existing || roleOrder[roleName] < roleOrder[existing]) {
+      map.set(identityId, roleName)
+    }
+  }
+  return map
```
david-crespo, line 101: This is fun, this was actually a bug — it was keeping whatever the last assignment in the policy was instead of the strongest one. ~~Looking into how this is used and whether this is a better way to do it.~~ Yeah I'm pretty sure this is right.

charliepark (approved):
Ah, thanks for catching this!

PR #3305, opened 2026-07-24
Add confirmation step when admin removes their admin role at silo level

### LINKED ISSUE #3303: Add warning when a silo admin is about to revoke their admin permissions
Currently, there's nothing stopping a Silo Admin from editing their own permissions and downgrading their role to something other than `admin`. Because only Silo Admins can modify permissions this means they could end up putting themselves in a state where they're a silo viewer (or whatever) and then can't restore their admin role. Although we aren't going to add a restriction to the API, we can add some friction to the Console UI.

There's a bit of complexity with the fact that someone could be a Silo Admin because of a group assignment, but the simplest bit of friction would just add a confirmation when a Silo Admin is affecting their own account and are changing the role to something other than Admin.

charliepark:
When working on #3211, I noticed that a silo admin can update their own silo admin role to a role with fewer permissions (say, "viewer"), which would act as a one-way door, preventing them from un-do-ing that change. We don't want to fully prevent that from happening, but we can have a warning, just like we do for when they might delete their own role.

Keeping this outside of #3211 just because that one's long enough already, and this is mostly unrelated to that. Probably makes sense to get #3211 merged first, though.

Screenshot 2026-07-24 at 4 33 32 PM

Closes #3303

### DIFF
--- a/app/forms/silo-access.tsx
+++ b/app/forms/silo-access.tsx
@@ -18,6 +18,9 @@ import { Access16Icon } from '@oxide/design-system/icons/react'
 
 import { ListboxField } from '~/components/form/fields/ListboxField'
 import { SideModalForm } from '~/components/form/SideModalForm'
+import { HL } from '~/components/HL'
+import { useCurrentUser } from '~/hooks/use-current-user'
+import { confirmAction } from '~/stores/confirm-action'
 import { SideModalFormDocs } from '~/ui/lib/ModalLinks'
 import { ResourceLabel } from '~/ui/lib/SideModal'
 import { docLinks } from '~/util/links'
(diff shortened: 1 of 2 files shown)


fakemonster (approved):
~more of a "will" than a "may", isn't it?~ oh, i suppose if you're a fleet admin then this really is just a "may"!

charliepark:
> i suppose if you're a fleet admin then this really is just a "may"!

There's also the complicating factor that IF a user is in a group and IF that group has silo admin permissions, then the user would still have admin permissions, but it gets a bit in the weeds to get into that.

On the other hand, we go with the scarier message of "This will remove your own silo access." on the "delete" flow:
Screenshot 2026-07-29 at 3 04 41 PM

fakemonster:
i think either is basically fine, preferring your original though. It's too much to enumerate the different ways you can obtain to this permission

PR #3308, opened 2026-07-28
Add write support for Internet Gateways

benjaminleonard:
Previously read-only. This adds the full write API.

---

1. Create and delete gateways:
    -  "New internet gateway" side modal
    - Delete row action. With cascade by default. I'm thinking a checkbox for that might be better. The API description is incomplete? "Also delete routes targeting this gateway." But the robot says it also soft-deletes IP pools and addresses. Or just the associations?
    
image
image
    
    
2. Full detail page
    - Replaces the read-only side modal
    - Cards for attached IP pools, attached addresses and routes

image


app/pages/project/vpcs/InternetGatewayPage.tsx
```diff
@@ -0,0 +1,396 @@
...
+  })
+
+  const routesData = useMemo(
+    () => (matchingRoutes || []).map(([router, route]) => ({ router, route: route.name })),
+    [matchingRoutes]
+  )
+  const routesTable = useReactTable({
+    columns: useMemo(
+      () => [
+        routeColHelper.accessor('router', {
```
benjaminleonard, line 234: Do we want to flip this and have the route first instead of the router?

benjaminleonard:
More of an API design question. But how helpful is the name here in practice? I do wonder if the name of the attached item is sufficient.

image

benjaminleonard:
image

Added the ability to toggle `cascade` with a radio card in the delete modal.

benjaminleonard:
Delete logic should actually be a little more intuitive. We might want to gate on: https://github.com/oxidecomputer/omicron/issues/10976

ijsnow:
> Delete logic should actually be a little more intuitive. We might want to gate on: [oxidecomputer/omicron#10976](https://github.com/oxidecomputer/omicron/issues/10976)

FYI I opened [a PR](https://github.com/oxidecomputer/omicron/pull/11016) to unblock this

benjaminleonard:
Thanks @ijsnow! I'll defer to the network folks on the Omicron PR.

PR #3311, opened 2026-08-05
List and download Support Bundles

charliepark:
This PR creates a basic view at the System level for Support Bundles viewing and downloading, and creating new support bundles.
Screenshot 2026-09-09 at 7 08 48 PM

Support Bundles are listed newest-first. `ACTIVE` support bundles can be downloaded; download button is disabled for `FAILED` / `COLLECTING` / `DESTROYING` states.

This is a somewhat blunt page at the moment; talking with Sean, there are some enhancements we'll be able to do in the not-distant future, like filtering based on a time window or otherwise reducing the size of the bundle payload being downloaded. I did explore a file viewer partway through the building of this PR, but talking through it with Oliver, he suggested that the scope of files the user would have to manually go through in the client would be difficult, and the UX would be rough.

**NOTE:** Ben is working on a new icon for this.


hawkw:
This looks nice! 

Though one note is that some of the UI looks weird on mobile if the screen is too narrow to contain it:

image

This may be a more widespread problem than just this page, though, given that we don't collapse the sidebar?

charliepark:
Yes, thanks for calling that out! Responsive issues are definitely a larger thing, though I think fault management scenarios highlight the issue where someone might be viewing the client on a non-standard screen / in a state of crisis, so we want to tighten that all up!

app/components/StateBadge.tsx
```diff
@@ -85,6 +86,28 @@ export const SnapshotStateBadge = (props: { state: SnapshotState; className?: st
...
+
+export const SupportBundleStateBadge = (props: {
+  state: SupportBundleState
+  className?: string
+}) => (
+  <Badge
+    color={SUPPORT_BUNDLE_COLORS[props.state]}
+    className={cn(props.className, badgeClasses)}
+  >
+    {(props.state === 'collecting' || props.state === 'destroying') && (
```
fakemonster: really minor given the context, but i always like to use a match for this kind of thing, so the compiler can require our input if we add e.g. a `rebuilding` state in the future

mock-api/msw/handlers.ts
```diff
@@ -2012,6 +2012,107 @@ export const handlers = makeHandlers({
...
+        db.supportBundles = db.supportBundles.filter((b) => b.id !== bundle.id)
+      }, 3000)
+    }
+
+    return 204
+  },
+  // the generated handler type only allows status code returns for binary
+  // endpoints, but the dispatcher passes Response instances through untouched
+  // @ts-expect-error
+  supportBundleDownload({ path, cookies }) {
```
fakemonster: when i attempted a bundle download with the msw, i got thrown off by there being an implementation for bundle download when it isn't actually reachable (since we're navigating to the download link instead of fetching it). maybe somewhere it's worth a comment noting this only does something if you implement downloading as a fetch?

mock-api/msw/handlers.ts
```diff
@@ -2012,6 +2012,107 @@ export const handlers = makeHandlers({
...
+    }, 3000)
+
+    return json(newBundle, { status: 201 })
+  },
+  supportBundleUpdate({ path, body, cookies }) {
+    requireFleetAdmin(cookies)
+    const bundle = lookupById(db.supportBundles, path.bundleId)
+    // https://github.com/oxidecomputer/omicron/blob/99249b4/nexus/db-queries/src/db/datastore/support_bundle.rs#L736-L742
+    // byte length, not string length, to match Nexus
+    if (body.user_comment && new TextEncoder().encode(body.user_comment).length > 4096) {
```
fakemonster: i'm not sure if it's against convention to pull constants from the api utils, but it wouldn't hurt to share this one

app/pages/system/SupportBundlesPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+  </div>
+)
+
+const colHelper = createColumnHelper<SupportBundleInfo>()
+
+const staticColumns = [
+  colHelper.accessor('id', {
+    header: 'ID',
+    cell: (info) => (
+      <Truncate text={info.getValue()} maxLength={14} position="middle" hasCopyButton />
```
fakemonster: feels redundant to be able to copy the ID here and in the quick actions

fakemonster (commented):
code looks good. my only qualms:

1. the icon, as you noted
2. the double copy-id thing. fine to go in of course but has me wondering if we should try to migrate all the tables at once to keep the id-copying method predictable
3. do we want some sort of indicator for pages that rely on the experimental API?

charliepark:
Great point on /experimental. Would feel more comfortable with pulling it to /v1 before rolling this out, though I'm not sure about the other reasons for keeping it in /experimental

app/pages/system/SupportBundlesPage.tsx
benjaminleonard: +1

benjaminleonard:
Not sure we have a pattern for it currently but one of the main parts of this page, downloading the bundle, is nested within the menu. Maybe there's another pattern here.

Do we want to support the `support_bundle_index` endpoint? We could, by clicking on the bundle, open a route that fetches the support bundle index and offers a button to download it.

app/pages/system/SupportBundlesPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+const bundleList = getListQFn(
+  api.supportBundleList,
+  { query: { sortBy: 'time_and_id_descending' } },
+  {
+    refetchInterval: ({ state: { data } }) =>
+      data?.items.some((b) => b.state === 'collecting' || b.state === 'destroying')
+        ? POLL_INTERVAL
+        : false,
+  }
+)
```
david-crespo, line 101: Similar to instance list, nice. Might be good to make the concept of transitional state more explicit and make it a little fancier, like the instance list logic.

app/components/StateBadge.tsx
david-crespo: Same deal, pull out explicit transitional state list to match on here and in the polling logic.

david-crespo:
In a call we discussed adding a download button right in the table since it's the main thing you want to do with these. Need to try it out and see if it feels repetitive. Could be text, could be an icon. Otherwise row actions is fine.

benjaminleonard:
Text would be funky IMO. Icon _might_ work. 

But the approach in my comment above might be a better fix. Clicking the row would open a modal with the index metadata, the download action would live in there.

app/util/support-bundle.ts
```diff
@@ -0,0 +1,80 @@
...
+    queryKey: ['supportBundleIndex', bundleId],
+    queryFn: async ({ signal }) => {
+      const res = await fetch(bundleIndexUrl(bundleId), { signal })
+      if (!res.ok) throw new Error(`Error fetching bundle index (${res.status})`)
+      const text = await res.text()
+      return text.split('\n').filter((line) => line.length > 0)
+    },
+    // bundle contents never change once collection is complete
+    staleTime: Infinity,
+  })
```
david-crespo: It turns out this is a giant request — 4MB gzipped, 56MB unzipped. For just a file count I'm thinking maybe we can skip it? Or add an endpoint if we really care? It doesn't add that much. If we were going to add an actual file index that might be cool though. I think I would remove it for now and then use it for a file index view later. 

Image

Image

mock-api/msw/handlers.ts
```diff
@@ -2043,6 +2050,102 @@ export const handlers = makeHandlers({
...
+      setTimeout(() => {
+        db.supportBundles = db.supportBundles.filter((b) => b.id !== bundle.id)
+      }, 3000)
+    }
+
+    return 204
+  },
+  // the generated handler type only allows status code returns for binary
+  // endpoints, but the dispatcher passes Response instances through untouched
+  // @ts-expect-error
```
david-crespo, line 2135: made an issue to fix this in the generator https://github.com/oxidecomputer/oxide.ts/issues/386

app/pages/project/access/ProjectAccessPage.tsx
```diff
@@ -104,7 +104,7 @@ export default function ProjectAccessPage() {
   const { data: projectPolicy } = usePrefetchedQuery(projectPolicyView(projectSelector))
   const projectRows = useUserRows(projectPolicy.roleAssignments, 'project')
 
-  const rows = useMemo(() => {
+  const rows: UserRow[] = useMemo(() => {
```
david-crespo, line 107: This change was made necessary by a rather obscure TypeScript bug interacting with React Table. When I merged main into this branch, the `satisfies` version started failing typechecking. Claude was able to repro the issue on main by adding a single line that doesn't look like it should do anything. 

https://github.com/microsoft/typescript-go/issues/3973
https://github.com/microsoft/typescript-go/issues/4827

Image

app/util/support-bundle.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+  vi.stubGlobal('fetch', (url: string) => {
+    urls.push(url)
+    return Promise.resolve(new Response(null, { status: 204 }))
+  })
+
+  await api.supportBundleDownload({ path: { bundleId: 'bundle-id' } })
+
+  // 'http://testhost' is the client host under NODE_ENV=test (app/api/client.ts)
+  expect(urls).toEqual(['http://testhost' + bundleDownloadUrl('bundle-id')])
+})
```
david-crespo, line 31: This is kinda neat.

david-crespo (approved):
LET'S JAM

PR #3312, opened 2026-08-05
Oxql page

fakemonster:
Keeping in draft for a moment, because we may not want to merge this as-is, but this is a basic loop of "send a query, show it". There's the open design question of "do we prefer a builder form", which should be easier to discuss live. I'm also more than open to input on where we want to take the mock service worker for this. As-is it works just fine, but the OxQL results space is much wider than what the MSW is configured to deliver. You basically _have_ to point at real data to see the fully variety of visualizations.


david-crespo:
Idea: `query=` param so you can bookmark a query.

app/pages/system/OxqlPage.tsx
```diff
@@ -0,0 +1,594 @@
...
+          <Listbox
+            label="Load an example"
+            placeholder="Select an example query"
+            selected={currentQuery || null}
+            items={exampleItems}
+            onChange={(text) => {
+              setTextareaRowCount(getTextareaHeightForQuery(text))
+              form.setValue('query', text)
+            }}
+          />
```
david-crespo: I'd make this a dropdown menu button instead of a listbox to avoid giving the impression of a state that syncs with the contents of the textarea. It's just a one-time replace on select, so menu buttons are more appropriate. It would also make it visually smaller.

app/util/path-builder.ts
```diff
@@ -115,6 +115,7 @@ export const pb = {
   siloImage: (params: PP.SiloImage) => `${pb.siloImages()}/${params.image}`,
 
   fleetAccess: () => '/system/access',
+  oxql: () => '/system/oxql',
```
david-crespo: might be good to call this `systemOxql` since we'll probably have `siloOxql`

app/pages/system/OxqlPage.tsx
fakemonster: to be fair, my intent is to make the query picker, however it's implemented, not feel like an independent input. so it being in sync (like, if you pick a query, delete a paren, and reintroduce it, it re-highlights the query you match) is intentional. but i do like the low profile of the dropdown here. would you imagine it being stateful, meaning you click "Power shelf fan speeds" and that remains visible as you edit, or would you just keep resetting the dropdown state to "Pick an example" instantly?

david-crespo: I was picturing it like this — no state. To me that fits with the queries being starting points.

image

fakemonster: i can vibe with that. let me toy with it (and placement)

fakemonster: Screenshot_20260819_124241

comme ça? i didn't want to toy with encroaching on the vertical space of the textarea itself, but i think kicking this to the right helps suggest it isn't "part of the form" itself. could make a case for putting it below as well

david-crespo: I think that's fine for now. @benjaminleonard is going to work on it too, so I wouldn't try to get it too perfect. I could see him making some kind of horizontal bar above that has this button in it, etc. Anything is possible.

PR #3316, opened 2026-08-10
Bump API to latest, filter out service IP pools from list fetches

david-crespo:
I confirmed on dogfood that the filtering works, the two service pools get filtered out.


app/pages/system/networking/IpPoolsPage.tsx
```diff
@@ -130,7 +134,7 @@ export default function IpPoolsPage() {
   })
 
   const { data: allPools } = useQuery(
-    q(api.systemIpPoolList, { query: { limit: ALL_ISH } })
+    q(api.systemIpPoolList, { query: { assignment: 'silos', limit: ALL_ISH } })
```
fakemonster, line 137: not relevant here, but noting while we're on the subject that we're querying ip pools twice on this page

fakemonster (approved):

bnaecker (approved):

app/pages/system/networking/IpPoolsPage.tsx
david-crespo: Yep, unfortunately this is how we have to do it to give the quick actions menu the full list while also letting the table fetch its single page. It's especially silly on lists that are expected to be short, like this one, where the two fetches are almost certainly going to give you the same results. There are a few ways we could try to avoid it, but I don't think it's worth it.

PR #3319, opened 2026-08-12
Instance create: disable default networking when no `default` VPC

david-crespo:
The reason this half-worked is that we were disabling `Default` when there were no VPCs at all. Now we specifically check for a VPC called `default`. I also improved the styling on a disabled radio button — before the text was normal color and you couldn't tell it was disabled.

image

---

image

### DIFF
--- a/app/components/form/fields/NetworkInterfaceField.tsx
+++ b/app/components/form/fields/NetworkInterfaceField.tsx
@@ -8,7 +8,7 @@
 import { useState } from 'react'
 import { useController, type Control } from 'react-hook-form'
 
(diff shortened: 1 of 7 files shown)


app/components/form/fields/NetworkInterfaceField.tsx
```diff
@@ -79,15 +87,22 @@ export function NetworkInterfaceField({
         aria-labelledby="network-interface-type-label"
       >
         <div className="space-y-2">
-          <Radio
-            name="networkInterfaceType"
-            value="default"
-            disabled={!hasVpcs || disabled}
```
david-crespo, line 85: This was the wrongest bit.

PR #3320, opened 2026-08-12
Add webhooks to the UI

charliepark:
This PR adds the Alerting sidebar and the two tabs that nest under it, Alerts and Receivers. It also then shows the list of Receivers, all of which are Webhook Receivers for the present and near-term future. The webhook receiver creation form uses a richer input for the alert subscriptions that Ben created in #3326.

There's an Alerts list that @fakemonster is working on separately.

alerting - receivers
alerting - receiver-new-1
alerting - receiver-new-2
alerting - alerts


app/pages/system/alerting/AlertReceiversTab.tsx
```diff
@@ -0,0 +1,159 @@
...
+const colHelper = createColumnHelper<AlertReceiver>()
+
+const staticColumns = [
+  colHelper.accessor('name', {
+    cell: makeLinkCell((receiver) => pb.alertReceiver({ receiver })),
+  }),
+  colHelper.accessor('subscriptions', {
+    header: 'Events',
+    cell: (info) => (
+      <ListPlusCell tooltipTitle="Other events">
```
hawkw:
```suggestion
    header: 'Alerts',
    cell: (info) => (
      <ListPlusCell tooltipTitle="Other alerts">
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+    </CardBlock>
+  )
+}
+
+// Event classes
+
+const subscriptionColHelper = createColumnHelper<{ subscription: string }>()
+const subscriptionCols = [
+  subscriptionColHelper.accessor('subscription', {
+    header: 'Event class',
```
hawkw:
```suggestion
// Alert classes

const subscriptionColHelper = createColumnHelper<{ subscription: string }>()
const subscriptionCols = [
  subscriptionColHelper.accessor('subscription', {
    header: 'Alert class',
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+    [receiver.subscriptions]
+  )
+  const table = useReactTable({ columns, data: rows, getCoreRowModel: getCoreRowModel() })
+
+  return (
+    <CardBlock>
+      <CardBlock.Header
+        title="Event classes"
+        description="The event classes the webhook is subscribed to"
+      >
```
hawkw, line 206:
```suggestion
        title="Alert classes"
        description="The alert classes the webhook receiver is subscribed to"
```

or perhaps:
```suggestion
        title="Alert subscriptions"
        description="The alert classes the webhook receiver is subscribed to"
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+        <Button size="sm" onClick={() => setShowAddModal(true)}>
+          Add event class
```
hawkw: I think I'd prefer to call this a subscription rather than a "class" since it may also be a glob
```suggestion
          Add subscription
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+      </CardBlock.Header>
+      <CardBlock.Body>
+        {rows.length ? (
+          <Table table={table} aria-label="Event classes" />
+        ) : (
+          <TableEmptyBox>
+            <EmptyMessage
+              icon={<Webhooks24Icon />}
+              title="No subscriptions"
+              body="Subscribe to an event class to receive events"
```
hawkw:
```suggestion
          <Table table={table} aria-label="Alert classes" />
        ) : (
          <TableEmptyBox>
            <EmptyMessage
              icon={<Webhooks24Icon />}
              title="No subscriptions"
              body="Subscribe to an alert class to receive alerts"
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+      addToast(<>Subscribed to <HL>{result.subscription}</HL></>)
+      onDismiss()
+    },
+  })
+
+  return (
+    <ModalForm
+      form={form}
+      onDismiss={onDismiss}
+      title="Add event class"
```
hawkw:
```suggestion
      title="Add subscription"
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+            of events, like <InlineCode>hardware.**</InlineCode> or{' '}
+            <InlineCode>**.remove</InlineCode>.
+          </>
+        }
+      />
+      <ComboboxField
+        control={control}
+        name="subscription"
+        label="Subscription"
+        placeholder="Enter event pattern"
```
hawkw:
```suggestion
        placeholder="Enter alert pattern"
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+      }
+      loading={addSubscription.isPending}
+      submitError={addSubscription.error}
+    >
+      <Message
+        variant="info"
+        content={
+          <>
+            Event subscriptions may include simple globs to subscribe to multiple categories
+            of events, like <InlineCode>hardware.**</InlineCode> or{' '}
```
hawkw:
```suggestion
            Alert subscriptions may include simple globs to subscribe to multiple classes
            of alerts, like <InlineCode>hardware.**</InlineCode> or{' '}
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+    columns,
+    data: secrets,
+    getCoreRowModel: getCoreRowModel(),
+  })
+
+  return (
+    <CardBlock>
+      <CardBlock.Header
+        title="Secrets"
+        description="Payloads are signed with every secret. Rotate by adding a new secret before deleting the old one."
```
hawkw, line 366: There's a doc section on this we might want to link to here: https://docs.oxide.computer/guides/alerts/reliable-receivers#_zero_downtime_webhook_secret_rotation

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+  { value: 'delivered', label: 'Delivered' },
+  { value: 'pending', label: 'Pending' },
+  { value: 'failed', label: 'Failed' },
+]
+
+const deliveryColHelper = createColumnHelper<AlertDelivery>()
+const staticDeliveryCols = [
+  // shortId for these two to force truncation
+  deliveryColHelper.accessor('id', { ...Columns.shortId, header: 'Delivery ID' }),
+  deliveryColHelper.accessor('alertId', { ...Columns.shortId, header: 'Event ID' }),
```
hawkw:
```suggestion
  deliveryColHelper.accessor('alertId', { ...Columns.shortId, header: 'Alert ID' }),
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+  deliveryColHelper.accessor('alertClass', {
+    header: 'Event class',
```
hawkw:
```suggestion
    header: 'Alert class',
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+        label: 'Resend',
+        disabled: delivery.trigger === 'probe' && 'Probes cannot be resent',
+        onActivate: () =>
+          confirmAction({
+            doAction: () =>
+              resendDelivery({
+                path: { alertId: delivery.alertId },
+                query: { receiver },
+              }),
+            errorTitle: 'Could not resend event',
```
hawkw:
```suggestion
            errorTitle: 'Could not resend alert',
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+            modalTitle: 'Confirm resend',
+            modalContent: (
+              <div className="space-y-4">
+                <p>
+                  Are you sure you want to resend this event? The dispatcher will attempt to
```
hawkw:
```suggestion
                  Are you sure you want to resend this alert? The dispatcher will attempt to
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+                  deliver it again.
+                </p>
+                <PropertiesTable>
+                  <PropertiesTable.Row label="Event class">
+                    <Badge color="neutral">{delivery.alertClass}</Badge>
+                  </PropertiesTable.Row>
+                  <PropertiesTable.IdRow id={delivery.alertId} label="Event ID" />
```
hawkw:
```suggestion
                  <PropertiesTable.Row label="Alert class">
                    <Badge color="neutral">{delivery.alertClass}</Badge>
                  </PropertiesTable.Row>
                  <PropertiesTable.IdRow id={delivery.alertId} label="Alert ID" />
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+    [resendDelivery, receiver]
+  )
+
+  const emptyState = (
+    <EmptyMessage
+      icon={<Webhooks24Icon />}
+      title="No deliveries"
+      body={
+        filter === 'all'
+          ? 'Events delivered to this webhook will show up here'
```
hawkw:
```suggestion
          ? 'Alerts delivered to this webhook receiver will show up here'
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+      subtitle={
+        <ResourceLabel>
+          <Webhooks16Icon /> {receiver}
+        </ResourceLabel>
+      }
+    >
+      <SideModal.Body>
+        <SideModal.Section>
+          <PropertiesTable>
+            <PropertiesTable.Row label="Event class">
```
hawkw:
```suggestion
            <PropertiesTable.Row label="Alert class">
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+              <Badge color="neutral">{delivery.alertClass}</Badge>
+            </PropertiesTable.Row>
+            <PropertiesTable.IdRow id={delivery.id} label="Delivery ID" />
+            <PropertiesTable.IdRow id={delivery.alertId} label="Event ID" />
```
hawkw:
```suggestion
            <PropertiesTable.IdRow id={delivery.alertId} label="Alert ID" />
```

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -0,0 +1,929 @@
...
+    </SideModal>
+  )
+}
+
+// The delivery request format is defined by RFD 538 and built in
+// https://github.com/oxidecomputer/omicron/blob/32615a35/nexus/src/app/webhook.rs#L395-L555
+// The API does not return the request that was sent, so we reconstruct it from
+// the delivery record. Alert data, the alert version, and the signature can't
+// be known from here, so they show up as angle-bracket placeholders.
+const payloadJson = (delivery: AlertDelivery, sentAt: string) => `{
```
hawkw: We could, perhaps, just fetch the alert by ID using the `alert_view` API endpoint that was added in https://github.com/oxidecomputer/omicron/pull/11072

app/pages/system/alerting/AlertReceiversTab.tsx
```diff
@@ -0,0 +1,159 @@
...
+export const handle = makeCrumb('Receivers', pb.alertReceivers())
+
+export default function AlertReceiversTab() {
+  const navigate = useNavigate()
+
+  const { mutateAsync: deleteReceiver } = useApiMutation(api.alertReceiverDelete, {
+    onSuccess(_data, variables) {
+      queryClient.invalidateEndpoint('alertReceiverList')
+      // prettier-ignore
+      addToast(<>Webhook <HL>{variables.path.receiver}</HL> deleted</>)
```
hawkw:
```suggestion
      addToast(<>Webhook receiver <HL>{variables.path.receiver}</HL> deleted</>)
```

app/pages/system/alerting/AlertsTab.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const handle = { crumb: 'Alerts' }
+
+export default function AlertsTab() {
+  return (
+    <TableEmptyBox>
+      <EmptyMessage
+        icon={<Monitoring24Icon />}
+        title="No alerts"
+        body="Alerts fired by the system will appear here"
```
hawkw:
```suggestion
        body="Alerts published by the system will appear here"
```

hawkw (changes requested):
I like the UI, it looks great! Let's try to keep the language consistent with the terminology we use in the documentation, API, and the CLI. 

In particular, we should always say "alert" and "alert class" rather than "event" or "event class". In the API and public documentation, the term "event" is only used in the abstract (i.e. "Alerts provide a notification when an event has occurred in the system"); it's not the name of an object in the API. I think the only place where we refer to them as "events" is in RFD 538, which...should really be updated to reflect the terminology used in the product. My bad.

Also, I'd generally prefer to always refer to an endpoint that receives webhooks for alerts as a "webhook receiver", not a "webhook". Elsewhere, I've seen the term "webhook" used to refer to both the request itself *and* the thing that receives that request, which I think is a bit confusing, so I'd rather we use the terminology consistently.

Sorry for being persnickety about names, the UI looks awesome!

charliepark:
Thank you for all of the copy suggestions, @hawkw!

hawkw:
I'm not sure if this is the right place to review the alert list that @fakemonster is working on, but I have some notes:

1. I feel like it might be a bit nicer to put the timestamp first in the table, followed by the class. Something a bit more like the [audit log](https://console-git-audit-log-oxidecomputer.vercel.app/system/audit-log).
2. We should probably also include the alert's UUID in the table, since that's how you would distinguish between two alerts with the same class, and you'd potentially want to be able to copy that etc.
3. Maybe we can also use the same styling as the audit log, rather than the "table of resources" styling? 
4. Also, I like that in the audit log, the entire row is clickable to display details on that entry,  rather than having to click on the three vertical dots at the side of the row.
5. At least on a large, high-resolution display, this feels a little bit...sparse. 
  image
  Maybe we should left align the columns and then use the rest of the space for an "all in one line" display of the payload JSON, which it's fine to truncate? if the viewer wants to see the full json they can click on the alert's details
6. It seems like in #2849, @david-crespo is highlighting the JSON, like this:

  image

  It would be nice to reuse that here.
7. I feel a bit unsure about whether we should be typesetting the alert classes (both in the alert list _and_ the webhook receiver editing UIs) in `UPPERCASE`, since the canonical strings are always `lowercase`, and if someone copies an `UPPERCASE.ALERT.CLASS` out of the UI and tries to use it as a subscription to that alert class in the API, it...won't work. This could be pretty unfortunate.

hawkw:
The alert receivers UI is looking awesome, thanks for applying my suggestions! One note: when sending a liveness probe to a webhook receiver, we might want to include a chekcbox for whether to resend failed deliveries if the prove succeeds in the popup?

david-crespo:
This is looking awesome. I had the same idea as @hawkw:

> Maybe we can also use the same styling as the audit log, rather than the "table of resources" styling?

Landing on the alerts page currently feels a little opaque and unwelcoming. It's not clear where alerts sit in the conceptual setup. There are no actions I can take — I can't create an alert or delete one. There are created and modified columns but I have no idea how it happens. The audit log styling would give it more of the feel of a timeline of immutable events. And clickable rows would be easier to discover than the details button in the row actions (which I in fact failed to discover until I saw Eliza mention it).

Separately, I'm also wondering if maybe receivers is a better first page since they're what you can actually configure and get your hands on.

image

app/pages/system/alerting/AlertReceiversTab.tsx
```diff
@@ -0,0 +1,159 @@
...
+)
+
+const colHelper = createColumnHelper<AlertReceiver>()
+
+const staticColumns = [
+  colHelper.accessor('name', {
+    cell: makeLinkCell((receiver) => pb.alertReceiver({ receiver })),
+  }),
+  colHelper.accessor('subscriptions', {
+    header: 'Alerts',
```
david-crespo: This should be called "subscriptions".

charliepark:
I like the evolving audit-log-ish view.
Screenshot 2026-09-02 at 1 49 07 PM
Screenshot 2026-09-02 at 1 49 19 PM

fakemonster:
> The audit log styling would give it more of the feel of a timeline of immutable events.

Agreed, this is where I had actually started, but since the audit log isn't merged yet, I opted to just do the basic thing here instead of straddling two branches. Once we've actually got the audit-log branch merged to main, we can/should hop over to it. Though perhaps as a separate PR? This seems awfully close to landing already, and if alerts is no longer the landing tab, it's not too much of a dead end. I can pick it up once I'm back in, I've already got a bit of a branch refactoring a common list component out of the audit log branch. Would be a good moment to introduce filtering by alert class

app/pages/system/alerting/AlertsTab.tsx
```diff
@@ -0,0 +1,200 @@
...
+        <PropertiesTable.Row
+          // TODO: explain this in the info bubble or column header. doc comment below
+          //
+          // Alert schemas are versioned on a per-alert-class basis. The schema version
+          // for a particular alert class does not correspond to an Oxide API version.
+          // Clients should expect to encounter earlier schema versions when retrieving
+          // alerts recorded by an earlier version of the system software.
+          label="Class version"
+        >
+          <Badge color="neutral">{alert.version}</Badge>
```
david-crespo: I lean against badge here.

app/forms/webhook-create.tsx
```diff
@@ -0,0 +1,222 @@
...
+    </>
+  )
+}
+
+const globCode = 'text-mono-sm bg-info-secondary text-info rounded-sm px-1'
+
+const SubscriptionsMessage = (
+  <>
+    Alert subscriptions may include simple globs to subscribe to multiple classes of alerts.
+    E.g. <code className={globCode}>hardware.**</code> or{' '}
```
fakemonster: what do you think of changing `hardware.**` to `hardware.sled.*`? i don't think we need to explain it but it'd be a helpful hint to see both a single and double asterisk example

mock-api/msw/handlers.ts
```diff
@@ -2661,24 +2808,221 @@ export const handlers = makeHandlers({
...
+    const receiver = lookup.alertReceiver(path)
+    db.alertReceivers = db.alertReceivers.filter((r) => r.id !== receiver.id)
+    db.alertDeliveries = db.alertDeliveries.filter((d) => d.receiver_id !== receiver.id)
+    return 204
+  },
+  alertDeliveryList({ path, query, cookies }) {
+    requireFleetViewer(cookies)
+    const receiver = lookup.alertReceiver(path)
+    retryPendingDeliveries(receiver)
+    // probe deliveries are stored like any other but never listed, matching
```
fakemonster, line 2720: this surprised me during testing. should we tell users "if you send a probe, don't expect it in the deliveries list?" maybe fine without and it's just my attitude

app/forms/webhook-create.tsx
```diff
@@ -0,0 +1,222 @@
...
+        rowKey={(secret) => secret}
+        onRemoveItem={(secret) => field.onChange(field.value.filter((s) => s !== secret))}
+        removeLabel={(secret) => `remove secret ${secret}`}
+      />
+      <ErrorMessage error={fieldState.error} label="Secrets" />
+    </>
+  )
+}
+
+const globCode = 'text-mono-sm bg-info-secondary text-info rounded-sm px-1'
```
fakemonster: could we `normal-case` here? would match the edit form (and the table view)

app/components/form/fields/SubscriptionsField.tsx
```diff
@@ -0,0 +1,531 @@
...
+                          )
+                        }
+                      >
+                        {row.description}
+                      </ItemLabel>
+                    </span>
+                    {label && (
+                      // mt-1 optically centers the 1rem mono label on the
+                      // 1.5rem name line
+                      <span className={cn('text-mono-xs mt-1', label.className)}>
```
fakemonster: this is an awkward one to suggest normal-case on... maybe it just stays weird?

david-crespo:
The option showing up on top of the `Matches 11 alert classes:` is a little janky, not sure what to do about it. You could move the matches thing down but then you have a weird gap. maybe it's part of the options popover but not a selectable option?

image

mock-api/msw/handlers.ts
david-crespo: I think we should say it somewhere, maybe right in the send probe modal?

there needs to be some charisma to carry this kind of move through

app/api/util.ts
```diff
@@ -40,6 +41,69 @@ export const INSTANCE_MAX_CPU = 254
...
+    .join('\\.')
+  return new RegExp(`^${pattern}$`)
+}
+
+/**
+ * IDs of the alerts a probe with `resend=true` would requeue: the receiver has
+ * a delivery for the alert and no non-probe delivery of that alert has left the
+ * failed state. Note this is per alert, not per delivery — delivery records are
+ * immutable history, so a failed one stays failed forever and a resend inserts
+ * a new record. The API has no endpoint for this, so we derive it from the
```
fakemonster, line 88: should it?

app/components/AlertClassBadge.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Badge for an alert class or subscription glob. Badges uppercase their text
+ * by default, but alert classes are lowercase in the API and someone who copies
+ * one out of the UI needs it to work as a subscription, so keep the case.
+ *
+ * Long classes truncate to fit their container, with the full name in a
+ * tooltip when that happens.
+ */
+export const AlertClassBadge = ({ children }: { children: string }) => {
```
fakemonster, line 23: this is something that's also come up in oxql land, which is the need for a style that looks an awful lot like a badge but is actually inline code (both in that we could have different line-break opinions long run, but more importantly that it must respect letter case). we shouldn't change anything about this now but i wanted to mention we might come back to this to solve it more generally

app/components/HighlightJSON.tsx
fakemonster, line 1: excited to delete this module when we get shiki in lol

app/components/SubscriptionMatchPreview.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * For a glob subscription pattern, show which alert classes it currently
+ * matches. Renders nothing for exact (non-glob) patterns.
+ * Note the match set is point-in-time: globs are re-evaluated by the control
+ * plane as alert classes are added.
+ */
+export function SubscriptionMatchPreview({ pattern }: { pattern: string }) {
+  // Same query as the class picker this sits under, so it's a cache hit rather
+  // than a fetch. Matching locally with `subscriptionRegex`, mirroring the
+  // control plane's glob compiler.
+  const { data } = useQuery(q(api.alertClassList, { query: { limit: ALL_ISH } }))
```
fakemonster: why not pass it as a dependency then? less fragile, simpler component

```diff
diff --git a/app/components/SubscriptionMatchPreview.tsx b/app/components/SubscriptionMatchPreview.tsx
index 40e9d57f..78c6843d 100644
--- a/app/components/SubscriptionMatchPreview.tsx
+++ b/app/components/SubscriptionMatchPreview.tsx
@@ -5,9 +5,7 @@
  *
  * Copyright Oxide Computer Company
  */
-import { useQuery } from '@tanstack/react-query'
-
-import { api, q } from '@oxide/api'
+import type { AlertClassResultsPage } from '@oxide/api'
 
 import {
   ALERT_SUBSCRIPTION_REGEX,
@@ -16,7 +14,6 @@ import {
   subscriptionRegex,
 } from '~/api/util'
 import { AlertClassBadge } from '~/components/AlertClassBadge'
-import { ALL_ISH } from '~/util/consts'
 
 /**
  * For a glob subscription pattern, show which alert classes it currently
@@ -24,12 +21,13 @@ import { ALL_ISH } from '~/util/consts'
  * Note the match set is point-in-time: globs are re-evaluated by the control
  * plane as alert classes are added.
  */
-export function SubscriptionMatchPreview({ pattern }: { pattern: string }) {
-  // Same query as the class picker this sits under, so it's a cache hit rather
-  // than a fetch. Matching locally with `subscriptionRegex`, mirroring the
-  // control plane's glob compiler.
-  const { data } = useQuery(q(api.alertClassList, { query: { limit: ALL_ISH } }))
-
+export function SubscriptionMatchPreview({
+  data,
+  pattern,
+}: {
+  data: AlertClassResultsPage | undefined
+  pattern: string
+}) {
   // validate before subscriptionRegex, which assumes a well-formed subscription
   const isValidGlob = isGlobPattern(pattern) && ALERT_SUBSCRIPTION_REGEX.test(pattern)
   if (!isValidGlob || !data) return null
diff --git a/app/pages/system/alerting/AlertReceiverPage.tsx b/app/pages/system/alerting/AlertReceiverPage.tsx
index 1747a5e4..a059d357 100644
--- a/app/pages/system/alerting/AlertReceiverPage.tsx
+++ b/app/pages/system/alerting/AlertReceiverPage.tsx
@@ -303,7 +303,7 @@ function AddSubscriptionModal({ onDismiss }: { onDismiss: () => void }) {
         required
         validate={(value) => validateSubscription(value, classNames)}
       />
-      <SubscriptionMatchPreview pattern={subscription} />
+      <SubscriptionMatchPreview data={classes.data} pattern={subscription} />
     </ModalForm>
   )
 }
```

app/forms/webhook-create.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  return (
+    <>
+      <PageHeader>
+        <PageTitle icon={<Webhooks24Icon />}>Create webhook receiver</PageTitle>
+      </PageHeader>
+      <FullPageForm
+        id="create-webhook-form"
+        form={form}
+        onSubmit={async ({ name, description, endpoint, secrets, subscriptions }) => {
```
fakemonster, line 192: i'm inclined to replace this with `async (body) => { await createWebhook.mutateAsync({ body }) }`, so if we extend the form we aren't surprised to find this function also needs updating. but commenting first in case i'm missing a good reason not to

app/pages/system/alerting/AlertReceiverDeliveries.tsx
```diff
@@ -0,0 +1,416 @@
...
+import { EmptyMessage } from '~/ui/lib/EmptyMessage'
+import { Listbox } from '~/ui/lib/Listbox'
+import { PropertiesTable } from '~/ui/lib/PropertiesTable'
+import { ResourceLabel, SideModal } from '~/ui/lib/SideModal'
+import { TableEmptyBox } from '~/ui/lib/Table'
+import { Tabs } from '~/ui/lib/Tabs'
+
+type StateFilter = 'all' | AlertDeliveryState
+
+const stateFilterParams = (filter: StateFilter) =>
```
fakemonster: this kind of function makes me feel good

charliepark:
> The option showing up on top of the `Matches 11 alert classes:` is a little janky, not sure what to do about it. You could move the matches thing down but then you have a weird gap. maybe it's part of the options popover but not a selectable option? [image]

I just pulled the new SubscriptionsField into this modal, to make subscription adding/removing more consistent between the overall create form and the edit modal.

app/pages/system/alerting/AlertReceiverPage.tsx
```diff
@@ -221,66 +214,68 @@ function SubscriptionsCard() {
...
+          path: receiverSelector,
+          body: { subscription },
+        })
+      }
+      for (const subscription of removed) {
+        await removeSubscription.mutateAsync({
+          path: { ...receiverSelector, subscription },
+        })
+      }
+    } catch {
```
david-crespo: I don't think this is a good idea — error handling here is kinda tough if one of them fails. If one fails for a good reason and you have to edit the input, it would be hard for a user to figure out what to edit. Since the API wants us to do it one request at a time, I think we should stick with that here for now. We could in theory still use the same UI but limit it to a single entry, but then we have to explain why it's limited and it doesn't feel right — might be better to keep the simpler single sub picker for now.

charliepark: Ah, good point; thank you. Will revert that.

app/pages/system/alerting/AlertReceiverDeliveries.tsx
```diff
@@ -0,0 +1,416 @@
...
+      <SideModal.Footer>
+        <Button variant="ghost" size="sm" onClick={onDismiss}>
+          Close
+        </Button>
+      </SideModal.Footer>
+    </SideModal>
+  )
+}
+
+// The delivery request format is defined by RFD 538 and built in
```
fakemonster: this is very nice, but all seems quite prone to drift from the API implementation. how important is it to us that we display "what we think was sent" instead of just showing the alert itself as a bit of convenience?

app/pages/system/alerting/AlertReceiverTesting.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+  return (
+    <PropertiesTable>
+      <PropertiesTable.Row label="Result">
+        {attemptResultBadge(attempt.result)}
+      </PropertiesTable.Row>
+      <PropertiesTable.Row label="Status">
+        {status ? (
+          <span className="flex items-center gap-1.5">
+            {attempt.result === 'succeeded' ? (
```
fakemonster: i think not matching here is probably ok...

app/pages/system/alerting/AlertReceiverTesting.tsx
```diff
@@ -0,0 +1,282 @@
...
+}
+
+const resendNote = (preview: ResendPreview) =>
+  match(preview)
+    .with(
+      { state: 'unknown' },
+      () => 'Alerts that never reached the endpoint are queued for another attempt.'
+    )
+    .with(
+      { state: 'known', count: 0 },
```
fakemonster: this came up in matrix, let's check truncated here and if it is, let them hit resend. i'm not sure what the text should be in that case, maybe "The last $ALL_ISH alerts have reached this endpoint; older ones may have failed." perhaps we drop the clarification on the end? "We checked the last $ALL_ISH alerts and they all reached this endpoint" sounds like a buzzfeed article

app/pages/system/alerting/AlertReceiverDeliveries.tsx
charliepark: Agree that representing the alert directly is clearer. Have done that in [e06dd46](https://github.com/oxidecomputer/console/pull/3320/commits/e06dd465279c2feb47d016684c1dfacd6d147e6f). I wonder if we should make the alerts table (and alert sidemodals) linkable, and just link to the relevant alert? For now, keeping the duplicated alert views, but that feels like it could maybe be cleaned up.

app/pages/system/alerting/AlertReceiverTesting.tsx
david-crespo: Failed Alerts Hate Him!

app/pages/system/alerting/AlertReceiverTesting.tsx
david-crespo: Eh I would do it for the exhaustive match in case another non-error case is added. Adding now.

app/forms/webhook-create.tsx
david-crespo: My worry is that we add a field to the form that we don't intend to go directly into the API response — TS won't error on that. The mock API's zod validators strip unknown fields on the way in, so it doesn't care, and the real API doesn't have `#[serde(deny_unknown_fields)]` on this post body, so it would be ignored there too. So strictly speaking it doesn't really matter, but I vaguely still prefer to avoid adding fields accidentally. If a new required field appears in the post body, we would get a type error so we'd know to handle it.

app/pages/system/alerting/AlertReceiverTesting.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  // throwOnError off because this only feeds a preview count: if it fails the
+  // modal falls back to describing the behavior without a number
+  const { data } = useQuery(
+    q(
+      api.alertDeliveryList,
+      { path: { receiver }, query: { limit: ALL_ISH } },
+      { throwOnError: false }
+    )
+  )
```
david-crespo, line 67: The API sorts by time and ID ascending by default, so this is fetching the oldest 1000 deliveries rather than the most recent. Fixing.

app/pages/system/alerting/AlertReceiverDeliveries.tsx
fakemonster: yeah let's definitely not do anything about it in this PR. that's a feature in and of itself

david-crespo:
### Typical case

If there are more than 1000 deliveries this will say "At least 2..."

Screenshot 2026-09-11 at 12 01 16 PM

### Hide checkbox when no alerts have ever been sent

Screenshot 2026-09-11 at 12 01 01 PM

### Disable with message when there are none to resend

Screenshot 2026-09-11 at 12 00 49 PM

david-crespo (approved):
WE ARE DOWN TO CLOWN (ALERT)

PR #3324, opened 2026-08-14
Support bundle detail modal

benjaminleonard:
Proposal for #3311 

CleanShot 2026-08-14 at 11 01 03

---

Adds a support bundle detail view, with metadata and a download button and merges the support bundle comment edit into here.

This adds a better path to download the bundle (rather than being tucked away in the action menu). I'm not opposed to getting the file tree in here, but something like `@pierre/trees` is probably overkill.


app/forms/support-bundle-create.tsx
```diff
@@ -62,7 +62,6 @@ export default function CreateSupportBundleSideModalForm() {
         as="textarea"
         name="userComment"
         label="Comment"
-        description="Note about why this bundle is being collected"
```
benjaminleonard, line 65: Perhaps unnecessary?

app/forms/support-bundle-edit.tsx
benjaminleonard, line 1: Moved into the detail modal

charliepark (approved):
Looks good! Thank you!

PR #3325, opened 2026-08-14
Width-aware truncation

### LINKED ISSUE #1536: Truncate that can fill space
Some combination of a css truncate plus this: http://jsfiddle.net/nsngdp7b/ to test if it's truncated and show the copy button.

### LINKED ISSUE #3261: Revisit logic on PropertiesTable IdRow, re: copyability, truncation
@fakemonster and I are chatting through the PropertiesTable.IdRow component, and he noted a few things:
1. Because of our using the Truncate component within it — the copy-to-clipboard button doesn't show up if the cell text isn't truncated, we only get the copy-to-clipboard affordance some of the time. It might be worth refactoring so that if `hasCopyButton` is true, the button shows up regardless of the truncation rules.
2. UUIDs are 36 characters (4 hyphens + 32 characters) and our default truncation length on the IdRow component is 32 characters, which means we lose 4 characters (and possibly more) no matter what, and if the cell is long enough, we should just be able to see the whole thing. So either we could show the whole thing, or show significantly fewer characters; as it is, it looks messy.
3. We might consider defaulting the UUID truncation, when applied, to the middle of the string, so users can visually compare either "first N" or "last N" characters, rather than just having the last characters chopped off.

benjaminleonard:
Closes #1536
Closes #3261

This has long been on my list to fix. The char-based truncation we have currently leaves a lot of unused space and is unreliable. In part because character widths (plus kerning) vary so much between two strings of identical number of characters.

`<Truncate>` now fills its container and truncates based on actual rendered width.

- `position="end"` is pure CSS (`text-overflow: ellipsis`). JS only checks `scrollWidth > clientWidth` on hover to decide whether to show the full-text tooltip
- `position="middle"` measures with canvas `measureText` (accounts for kerning and letter-spacing) and searches for the longest string that fits

Also standardizes ID display: IDs now middle-truncate everywhere, and adds truncation + copy buttons to places that had none. FileInput gets `contain: inline-size` on its label, fixing a bug where it forces the input wider. I _think_ this predates the truncation change.

Test changes to match.

From some local robot tests, the truncation measurements, including the binary-search are all very quick.

This also means we aren't truncating when there is space to display, before:

image


david-crespo:
I'm still seeing some truncation when there's space to spare. Looks awesome in motion, though.

https://github.com/user-attachments/assets/86118d16-1200-46c1-8a5e-7fb5ee0255e3

Probably not introduced here but I'm  noticing that the ellipsis plays poorly with the dashes, I wonder if there's room to be clever or we just have to live with it.

image

benjaminleonard:
Our right padding is deliberately a little bigger, it's a habit I've picked up since trailing spaces when identical to leading ones look a bit cramped. We do the same on cells, could potentially reduce though?

david-crespo:
I think it's too big if you can't get the whole UUID no matter how wide the window is. Probably only needs to be shrunk like 10%.

image

benjaminleonard:
Wait that's not right, that's a bug, the right margin is way smaller than that. Will take a look.

david-crespo:
Ah, this is a Firefox-only thing. Looks fine in Chrome and Safari.

benjaminleonard:
Looks like a sub-pixel discrepancy between the canvas measured text and the DOM measurement.

app/ui/lib/Truncate.tsx
```diff
@@ -52,6 +117,61 @@ export const Truncate = ({
...
+  // if we can't measure (jsdom) or the element isn't laid out yet, leave it alone
+  if (!ctx || el.clientWidth === 0) return text
+
+  const style = getComputedStyle(el)
+  // build the font shorthand from parts; `style.font` is empty in Firefox
+  ctx.font = `${style.fontStyle} ${style.fontWeight} ${style.fontSize} ${style.fontFamily}`
+  ctx.letterSpacing = style.letterSpacing === 'normal' ? '0px' : style.letterSpacing
+
+  const width = el.clientWidth
+  if (el.scrollWidth <= width) return text
```
david-crespo, line 145: GPT liked this better than comparing width to `measureText`.

app/ui/lib/Truncate.tsx
```diff
@@ -52,6 +117,61 @@ export const Truncate = ({
...
+
+  const graphemes = Array.from(graphemeSegmenter.segment(text), ({ segment }) => segment)
```
david-crespo, line 147: This prevents splitting unicode (e.g., emojis) in half — splits are only done between graphemes. See the test in `Truncate.spec.tsx`.

david-crespo (approved):
This looks great. I'm ready to merge, I think. I'm going to get #3287 done and then we can add browser tests for this, which can cover more edge cases.

app/ui/lib/Truncate.tsx
benjaminleonard: Yeah that's simpler

app/ui/lib/Truncate.tsx
benjaminleonard: Niche but nice!

PR #3326, opened 2026-08-14
Rich webhook event subscription input

benjaminleonard:
Rich event subscription input. Will require a bit of QA to check for robustness but the UX pattern feels satisfying. Good for both quickly adding classes and discoverability of the glob pattern matching.

Tag style input with chips for each of the subscription globs. Colour-coded to distinguish between exact and broad classes.

Dropdown opens on focus, showing the full catalog with name and description per row, subscribed classes sorted to the top (order is frozen while open so rows don't jump mid-picking). Rows are checkboxes: ticking commits an exact class without resetting the query, so multiple picks are cheap. Row states:

- *pending* — matches the glob being typed, labeled with the pattern (`HARDWARE.*.FAULT`)
- *near miss* — one segment too deep for the typed glob, labeled with the `**` widening that would cover it
- *covered* — matched by a committed glob; locked with a `via hardware.**` label so it can't be double-added
- Plain-text queries substring-filter with the match emphasized. Zero matches shows an empty state with a Clear button. An incomplete glob like `*.` shows the whole catalog rather than no results.

CleanShot 2026-08-14 at 12 02 44

Enter commits the query or toggles the highlighted row. Up/Down navigate rows, skipping covered ones. Backspace on an empty query arms the last chip; a second press deletes it. Left/Right move the armed selection so any chip can be deleted, and typing disarms.

Adding or removing chips can wrap the input to a new line, which would shift the dropdown; a ResizeObserver scrolls the page by the height delta so the panel and the line you're typing on stay put.

---

Also included:

`subscriptionRegex` moved from the mock handlers into `app/api/util.ts` (with tests) since the field now matches globs client side against a single catalog fetch.
- Mock catalog expanded with system-level classes (`hardware.sled.*`, `hardware.disk.*`, `system.update.*`, etc.) modeled on the existing PSU classes and RFD 538's FMA framing. Mock-only; Omicron still defines just three.
- Info message above the field explaining glob syntax, linking the webhooks guide and API docs. **Note:** the guide URL (`guides/operator/webhooks`) doesn't exist yet and needs to before release.
- E2E coverage for the full interaction set; the detail page's add-subscription modal is unchanged.


app/forms/webhook-create.tsx
```diff
@@ -38,97 +40,112 @@ export const validateEndpoint = (value: string) => {
...
-export const handle = titleCrumb('New webhook')
-
-export default function CreateWebhookSideModalForm() {
+const globCode = 'text-mono-sm bg-info-secondary text-info rounded-sm px-1'
+
+const SubscriptionsMessage = (
+  <>
+    Event subscriptions may include simple globs to subscribe to multiple categories of
+    events. E.g. <code className={globCode}>hardware.**</code> or{' '}
+    <code className={globCode}>**.fault</code>.{' '}
```
hawkw: Can we call these "Alerts" and "alert classes" rather than "events"? This is the language used in the API and in the documentation. 
```suggestion
    Alert subscriptions may include simple globs to subscribe to multiple  classes of
    alerts. E.g. <code className={globCode}>hardware.**</code> or{' '}
    <code className={globCode}>**.fault</code>.{' '}
```

app/forms/webhook-create.tsx
```diff
@@ -38,97 +40,112 @@ export const validateEndpoint = (value: string) => {
...
+    <a
+      href={links.webhooksGuide}
+      target="_blank"
+      rel="noreferrer"
+      className="mt-1 inline-block"
+    >
+      Read the Webhooks guide
```
hawkw, line 134: does this link to the section on globing in the alerts guide?

app/util/__snapshots__/path-builder.spec.ts.snap
```diff
@@ -71,6 +71,10 @@ exports[`breadcrumbs 2`] = `
       "label": "Alerts",
       "path": "/system/alerts",
     },
+    {
+      "label": "New webhook receiver",
+      "path": "/system/alerts-new",
```
hawkw, line 76: I don't think the name "alerts new" makes sense here. We aren't creating a new _alert_, we are creating a new _webhook receiver_, which is a kind of _alert receiver_.

app/util/links.ts
```diff
@@ -28,6 +28,9 @@ export const links = {
     'https://docs.oxide.computer/guides/configuring-guest-networking#_example_4_software_routing_tunnels',
   troubleshootingAccess:
     'https://docs.oxide.computer/guides/operator/faq#_how_do_i_fix_the_something_went_wrong_error',
+  // TODO: this guide does not exist yet; make sure it does before release
+  webhooksGuide: 'https://docs.oxide.computer/guides/operator/webhooks',
```
hawkw, line 32: I think this might be out of date: we could link to the top level alerts guide at https://docs.oxide.computer/guides/alerts/overview (which contains the explanation of alert classes and how glob patterns match them), and/or the webhooks guide at https://docs.oxide.computer/guides/alerts/webhooks (which discusses secrets and other details of how alerts are sent to webhooks receivers), as appropriate

app/api/util.spec.ts
```diff
@@ -7,7 +7,53 @@
...
+  genName,
+  instanceCan,
+  parsePortRange,
+  subscriptionRegex,
+  synthesizeData,
+} from './util'
+
+describe('subscriptionRegex', () => {
+  it('matches exact class names', () => {
+    expect(subscriptionRegex('probe').test('probe')).toBe(true)
```
hawkw, line 21: 'probe' is not a valid alert class to subscribe to, and the server will reject attempts to subscribe to the probe class with a 400 error. it's not treated as a "real" alert class. i think we shouldn't be testing with it if the API will reject it; we might want to explicitly include that rule in client side validation.

hawkw:
> Backspace on an empty query arms the last chip

what does this, uh. mean?

app/forms/webhook-create.tsx
david-crespo: Yep, we talked about this on a call yesterday after the FMA sync. also saying “subscriptions” as appropriate

hawkw: Great, just wanted to make sure we were all up to date since I think Benji had first started working on this several months ago, before we renamed basically everything.

david-crespo:
Backspace when the cursor is right next to a tag highlights the tag and then a second press deletes it. In this clip when the highlight ring is toggling on and off I'm pressing backspace, right, backspace, right. Then I press backspace twice to delete. Left/right also lets you move focus between the tags.

https://github.com/user-attachments/assets/1f4aa8d3-9b23-4c53-b3d4-dc06e286c8c2

david-crespo:
I kinda want `hardware.*` to turn into a chip as soon as I press space, though that might be hard to make work. Right now you have to press enter. You can figure it out pretty quickly, but it's not obvious you are supposed to press enter.

charliepark:
(For Eliza / anyone else following along, I'm planning to bring the comments and suggestions over to the original `webhooks` branch.)

app/forms/webhook-create.tsx
charliepark: I have updated the `webhooks` branch to link to the right spot

app/util/__snapshots__/path-builder.spec.ts.snap
charliepark: Updated in the `webhooks` branch to `/system/alerting/receivers-new`

app/util/links.ts
charliepark: Updated these over in the `webhooks` branch

app/api/util.spec.ts
charliepark: We are filtering it from the list now in the `webhooks` branch. Open to other conversation on it, but it was easy to drop it from the list.

PR #3330, opened 2026-08-18
Add quota editing sidemodal to system utilization page

### LINKED ISSUE #2406: Add 'Edit quotas' action to Utilization page
Backstory:
- In #2344 we'll be adding a "Copy ID" action to the table rows on the /utilization page.
- In #2369 we added an "edit quotas" side modal form to the Silo page.

It would be useful on the /utilization page, when looking at the quota / provisioned / available view to have the ability to edit the quota, as that's a spot where an admin might think "oof; let's bump that quota up a bit".

So this ticket has two parts:
1. Add an `Edit quotas` action to the table row, once a branch for #2344 is merged
2. Add the same edit form present on /system/silos/maze-war?tab=quotas to the page at /system/utilization. As the form doesn't depend on query params, I believe it should work just by passing a prop for the silo to the form.

charliepark:
This PR adds a sidemodal to the `/system/utilization` page, allowing the operator to edit quota values from that page. Before this change, quota updates have to be done from each silo's Quotas tab. (This PR doesn't remove that functionality; just adds it to the `/utilization` list page.)

Closes #2406

up-to-date screenshot:
Screenshot 2026-08-26 at 8 33 27 AM

### DIFF
--- a/app/forms/silo-quotas-edit.tsx
+++ b/app/forms/silo-quotas-edit.tsx
@@ -1,0 +1,119 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Copyright Oxide Computer Company
+ */
+import { useForm } from 'react-hook-form'
+import type { SetNonNullable } from 'type-fest'
+
+import {
+  api,
+  queryClient,
+  useApiMutation,
+  type SiloQuotasUpdate,
+  type VirtualResourceCounts,
+} from '@oxide/api'
+import { Cloud16Icon } from '@oxide/design-system/icons/react'
+
+import { NumberField } from '~/components/form/fields/NumberField'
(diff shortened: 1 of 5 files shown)


david-crespo:
This doesn't have to block the PR, but it might be worth a quick experiment: what if we showed the current provisioned amount in secondary text under each input? Otherwise you have to look back at the page under the modal. Which you can do but it's weird to have to. Plus if the window is narrow enough, the modal covers the rightmost columns:

image

charliepark:
I like the new microcopy, especially in support of the blue info box at the top.
Screenshot 2026-08-25 at 8 13 22 PM

david-crespo:
I’m into it. Could maybe do “provisioned:” alone, “currently” is implied I think. Ben might have something to say about where it goes — we could theoretically use the text hint we use for field descriptions, though I feel like as a user I am used to seeing those as static. Somehow below feels more live.

charliepark:
Screenshot 2026-08-26 at 8 26 16 AM

david-crespo:
Last commit puts the comma in the number here:

image

david-crespo (approved):
Love it!

PR #3332, opened 2026-08-18
Announce route changes for screen readers

david-crespo:
Similar deal to https://github.com/oxidecomputer/rfd-site/pull/258

## 🤖 summary

A client-side nav doesn't tell a screen reader anything: React Router swaps the DOM in place and focus stays on the link that was clicked, which usually isn't in the document anymore. Next.js ships a route announcer for this; RR leaves it to the app.

`useRouteAnnouncer` in `RootLayout` announces the crumbs deepest-first ("Instances, mock-project, Projects") through the react-aria live announcer we already use for toasts and field errors. Polite rather than assertive so it doesn't preempt the toast on flows that toast and then navigate. It also puts focus on the existing skip link target, but only when focus has fallen to `<body>` — if you clicked a sidebar link that's still sitting there, leave it alone. Use `preventScroll` to avoid conflicts with `useScrollRestoration` on back/forward.

Side modal forms get their own routes, so the announcer has to know that a form opening on top of a page isn't a page change. It keys off `titleOnly` crumbs, which already mark exactly those routes (had to fix `ip-pool-edit` and `subnet-pool-edit`, which were using `makeCrumb`). Detecting the open dialog in the DOM instead seemed simpler, but it ran into races.

### DIFF
--- a/app/hooks/use-crumbs.ts
+++ b/app/hooks/use-crumbs.ts
@@ -69,3 +69,14 @@ export const matchesToCrumbs = (matches: UIMatch[]) =>
     })
 
 export const useCrumbs = () => matchesToCrumbs(useMatches())
+
+/**
+ * Whether the current route is a side modal (form or detail panel) opening on
+ * top of a page. Keys off the `titleOnly` crumb flag because it exists for the
+ * same reason: the page underneath doesn't change.
+ *
+ * Not the same as `useIsInSideModal`, which asks whether the calling component
+ * is rendered inside a `SideModal`. That one is false on the page underneath;
+ * this one is true.
+ */
+export const useIsSideModalRoute = () => useCrumbs().some((c) => c.titleOnly)
--- a/app/hooks/use-route-announcer.ts
+++ b/app/hooks/use-route-announcer.ts
@@ -9,7 +9,7 @@ import { announce } from '@react-aria/live-announcer'
 import { useEffect, useRef } from 'react'
 import { useLocation } from 'react-router'
 
-import { useCrumbs } from './use-crumbs'
+import { useCrumbs, useIsSideModalRoute } from './use-crumbs'
 
 /**
  * A real page load tells a screen reader where it landed: the new page gets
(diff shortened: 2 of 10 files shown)


app/hooks/use-route-announcer.ts
```diff
@@ -0,0 +1,70 @@
...
+  // deepest crumb first, like the document title, so the specific page comes
+  // before its containers: "Instances, mock-project, Projects"
+  const pageName = crumbs
+    .map((c) => c.label)
+    .reverse()
+    .join(', ')
+
+  // `titleOnly` crumbs mark side modal forms, which have their own routes but
+  // open on top of a page rather than replacing it
+  const inSideModalForm = crumbs.some((c) => c.titleOnly)
```
david-crespo: Kind of a cute way to detect this but it is exactly what we want — we let the dialog announce itself, don't bother announcing the route.

benjaminleonard:
Have you confirmed with a screen reader?

app/hooks/use-crumbs.ts
```diff
@@ -69,3 +69,14 @@ export const matchesToCrumbs = (matches: UIMatch[]) =>
...
+/**
+ * Whether the current route is a side modal (form or detail panel) opening on
+ * top of a page. Keys off the `titleOnly` crumb flag because it exists for the
+ * same reason: the page underneath doesn't change.
+ *
+ * Not the same as `useIsInSideModal`, which asks whether the calling component
+ * is rendered inside a `SideModal`. That one is false on the page underneath;
+ * this one is true.
+ */
+export const useIsSideModalRoute = () => useCrumbs().some((c) => c.titleOnly)
```
david-crespo, line 82: Kind of a cute way to detect this but it is exactly what we want — we let the dialog announce itself, don't bother announcing the route. I extracted it to a common helper because I have another PR in progress that will use it.

david-crespo:
Yes, and it's not that great even though this PR does seem to improve things. Working on it.

PR #3334, opened 2026-08-18
Update IPv6 parsing

### LINKED ISSUE #1939: IPv6 regex doesn't exactly match `std::net::Ipv6Addr` parsing
So far I've only managed to find one case where our client-side IPv6 regex rejects an address that the API-side parsing by `std::net::Ipv6Addr` lets through. That means we will wrongly prevent the user from submitting the form, and if they really want it to go through, they have to use the CLI. I don't think this is a huge deal — some slippage here is inevitable — but it's good to track and would be nice to fix. A better fix might be to improve the server-side errors so we can make our client-side checking more minimal and lenient and therefore eliminate all false rejections.

```
2001:db8:3:4:5::192.0.2.33
```

I think it's because the `::` is next to the IPv4 bit, but I haven't looked into it in detail.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=493b3345b9f6c0b1c8ee91834e99ef7b

charliepark:
This updates our parsing / validating of IPv6 addresses, to make them more compliant with `std::net::Ipv6Addr` parsing. Instead of a regex, it uses the `new URL()` function as part of the evaluation.

Closes #1939

### DIFF
--- a/app/util/ip.ts
+++ b/app/util/ip.ts
@@ -27,34 +27,41 @@ const IPV4_REGEX =
 
 const IPV6_CHARS_REGEX = /^[0-9a-f:.]+$/i
 
-function isIpv6(ip: string): boolean {
-  if (!ip.includes(':') || !IPV6_CHARS_REGEX.test(ip)) return false
+/**
+ * Convert an IPv6 candidate to a form all browsers' URL parsers judge
+ * correctly, or return null if it's invalid in a way we can detect up front.
+ * The result is only structurally equivalent to the input (an embedded IPv4
+ * quad becomes placeholder groups) — use it for validity checking only, never
+ * as an address.
+ */
+export function toUrlCheckableIpv6(ip: string): string | null {
+  if (!ip.includes(':') || !IPV6_CHARS_REGEX.test(ip)) return null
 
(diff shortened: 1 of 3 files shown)


app/util/ip.ts
```diff
@@ -18,18 +18,30 @@ export const orderIps = (ips: ExternalIp[]) => R.sortBy(ips, (a) => IP_ORDER[a.k
...
+// IPv6 is more complex; use the WHATWG URL parser rather than a regex to determine validity.
+// Restrict the charset first so the parser only ever sees a bare host: `]`, `/`,
+// `%`, and whitespace would otherwise let non-addresses through as port, path, or zone.
+
+const IPV6_CHARS_REGEX = /^[0-9a-f:.]+$/i
+
+function isIpv6(ip: string): boolean {
+  if (!ip.includes(':') || !IPV6_CHARS_REGEX.test(ip)) return false
+  try {
+    new URL(`http://[${ip}]`)
```
david-crespo: WELL THAT'S CONVENIENT

david-crespo: Worth a comment here that the brackets are what tell it it has to be an IP and not, e.g., a domain.

david-crespo (approved):
nice

PR #3336, opened 2026-08-19
Don't scroll page to top when opening or closing a side modal

### LINKED ISSUE #3321: List pages scroll to top when you open detail side modal
Noticed while watching @sudomateo's excellent [k8s video](https://www.youtube.com/watch?v=lGeEPip-f1E). Scroll down, click a disk or image, and the page scrolls to the top underneath.

https://github.com/user-attachments/assets/e44513c8-2cbe-4382-8a44-71f7907758f9

https://github.com/user-attachments/assets/8c210a78-1676-44da-8605-68e7745b5887

david-crespo:
Closes #3321

Because side modals get their own routes, and the scroll height cache is keyed (in part) by route, opening a side modal counts as a new page with no saved scroll height, so it scrolls the document to top. The fix is to handle the modal scenario specially by opting out of resetting scroll when we switch to a new location that has the same underlying page path as the previous one. 

https://github.com/user-attachments/assets/af40dae5-7065-4599-be55-a9dd7ae7299e

### Example flows

Each of these scenarios is covered by e2e tests.

The underlying "page path" for a location is the path of the last non-`titleOnly` crumb, i.e., where the nav breadcrumbs point: a regular page's own path, or for a side modal, the path of the page under it. When a nav starts, we save the current scroll under the outgoing location's key. When it completes, if the page path changed we scroll to the new location's cached position (0 if none); if it didn't (side modal open/close), we leave the scroll alone and just record it under the new key so back/forward still works. 

#### Normal page → page nav (`…/disks` → `…/snapshots`, then back)

| Step                                                  | Location           | Page path     | Cache          | Scroll  |
| ----------------------------------------------------- | ------------------ | ------------- | -------------- | ------- |
| Scroll the disks page                                 | `…/disks` (k1)     | `…/disks`     |                | 143     |
| Click Snapshots: nav starts                           | `…/disks` (k1)     | `…/disks`     | k1: 143        | 143     |
| Nav completes: page changed → scroll to cached (none) | `…/snapshots` (k2) | `…/snapshots` | k1: 143        | 0       |
| Back: nav starts                                      | `…/snapshots` (k2) | `…/snapshots` | k1: 143, k2: 0 | 0       |
| Nav completes: page changed → scroll to cached        | `…/disks` (k1)     | `…/disks`     | k1: 143, k2: 0 | **143** |

#### Page → modal → page (open the New disk form, then dismiss it)

| Step                                                          | Location           | Page path | Cache                     | Scroll  |
| ------------------------------------------------------------- | ------------------ | --------- | ------------------------- | ------- |
| Scroll the disks page                                         | `…/disks` (k1)     | `…/disks` |                           | 143     |
| Click New disk: nav starts                                    | `…/disks` (k1)     | `…/disks` | k1: 143                   | 143     |
| Nav completes: same page → leave scroll, record under new key | `…/disks-new` (k2) | `…/disks` | k1: 143, k2: 143          | **143** |
| Dismiss: nav starts                                           | `…/disks-new` (k2) | `…/disks` | k1: 143, k2: 143          | 143     |
| Nav completes: same page → leave scroll, record under new key | `…/disks` (k3)     | `…/disks` | k1: 143, k2: 143, k3: 143 | **143** |

#### Modal submit → new page (create a VPC, land on VPC detail page)

| Step                                                  | Location             | Page path       | Cache            | Scroll |
| ----------------------------------------------------- | -------------------- | --------------- | ---------------- | ------ |
| Modal open over scrolled VPCs page                    | `…/vpcs-new` (k2)    | `…/vpcs`        | k1: 143, k2: 143 | 143    |
| Submit: nav starts                                    | `…/vpcs-new` (k2)    | `…/vpcs`        | k1: 143, k2: 143 | 143    |
| Nav completes: page changed → scroll to cached (none) | `…/vpcs/my-vpc` (k3) | `…/vpcs/my-vpc` | k1: 143, k2: 143 | **0**  |

### DIFF
--- a/app/hooks/use-scroll-restoration.ts
+++ b/app/hooks/use-scroll-restoration.ts
@@ -46,15 +46,22 @@ export function useScrollRestoration() {
   useEffect(() => {
     // opt out of native scroll restoration, it conflicts with ours
     window.history.scrollRestoration = 'manual'
+
+    // idle + new key = we just landed somewhere (or this is the initial render)
+    const landed = state === 'idle' && prev.current?.key !== key
+
+    // underlying page didn't change, which either means the nav was opening
+    // or closing a side modal, a query param tab change, or clicking the
+    // breadcrumb for the page you're already on
+    const samePage = prev.current?.page === page
+
     if (state === 'loading') {
       // nav in flight: save current scroll under the outgoing location's key
       setScrollPosition(key, window.scrollY)
-    } else if (state === 'idle' && prev.current?.key !== key) {
-      // we've landed on a page
-      if (prev.current?.page === page) {
-        // underlying page didn't change, which means the nav was opening or
-        // closing a side modal. leave scroll alone, record under the new key
-        // so it can be restored on forward/back
+    } else if (landed) {
+      if (samePage) {
+        // leave scroll alone, record under the new key so it can be restored
+        // on forward/back
         setScrollPosition(key, window.scrollY)
       } else {
         // landed on a new page: restore its saved scroll, or 0 if none saved
--- a/app/hooks/use-scroll-restoration.ts
+++ b/app/hooks/use-scroll-restoration.ts
@@ -5,8 +5,10 @@
  *
  * Copyright Oxide Computer Company
  */
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import { useLocation, useNavigation } from 'react-router'
+
+import { usePagePath } from './use-crumbs'
 
 function getScrollPosition(key: string) {
   const pos = window.sessionStorage.getItem(key)
@@ -33,15 +35,36 @@ function setScrollPosition(key: string, pos: number) {
 export function useScrollRestoration() {
   const key = `scroll-position-${useLocation().key}`
   const { state } = useNavigation()
+  // The page on screen at this location: a side modal route renders its parent
+  // page underneath, so it counts as that page. Navigating within the same
+  // page (opening or closing a side modal over it) shouldn't move the scroll.
+  const page = usePagePath()
+  // last committed location, tracked at idle. We can't track this in the
+  // loading state because a navigation whose loader data is already cached
+  // can complete without ever rendering with state === 'loading'.
+  const prev = useRef<{ key: string; page: string } | null>(null)
   useEffect(() => {
     // opt out of the browser's native scroll restoration so it doesn't jump
     // the still-visible old page to the new page's saved position on POP,
     // before the new route's loader resolves. We restore manually below.
     window.history.scrollRestoration = 'manual'
     if (state === 'loading') {
+      // during a navigation, location still reflects the old route
       setScrollPosition(key, window.scrollY)
-    } else if (state === 'idle') {
-      window.scrollTo(0, getScrollPosition(key))
+    } else if (state === 'idle' && prev.current?.key !== key) {
+      // both checks are needed: key changes on every nav, including same-page
+      // ones like opening a side modal, so key says we just landed on a new
+      // location (or this is the initial render) and page decides whether the
+      // nav stayed within the same page
+      if (prev.current?.page === page) {
+        // same page, new location (opened a side modal): leave the scroll alone
+        // and record it under the new location's key so back/forward to it
+        // restores correctly
+        setScrollPosition(key, window.scrollY)
+      } else {
+        window.scrollTo(0, getScrollPosition(key))
+      }
     }
-  }, [key, state])
+    if (state === 'idle') prev.current = { key, page }
+  }, [key, state, page])
 }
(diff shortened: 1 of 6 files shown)


app/hooks/use-scroll-restoration.ts
```diff
@@ -33,15 +35,36 @@ function setScrollPosition(key: string, pos: number) {
...
+    } else if (state === 'idle' && prev.current?.key !== key) {
+      // both checks are needed: key changes on every nav, including same-page
+      // ones like opening a side modal, so key says we just landed on a new
+      // location (or this is the initial render) and page decides whether the
+      // nav stayed within the same page
+      if (prev.current?.page === page) {
+        // same page, new location (opened a side modal): leave the scroll alone
+        // and record it under the new location's key so back/forward to it
+        // restores correctly
+        setScrollPosition(key, window.scrollY)
```
fakemonster, line 65: this branch isn't tested, strictly speaking. i could delete this line and pass

app/hooks/use-scroll-restoration.ts
```diff
@@ -33,15 +35,36 @@ function setScrollPosition(key: string, pos: number) {
...
+      // during a navigation, location still reflects the old route
       setScrollPosition(key, window.scrollY)
-    } else if (state === 'idle') {
-      window.scrollTo(0, getScrollPosition(key))
+    } else if (state === 'idle' && prev.current?.key !== key) {
+      // both checks are needed: key changes on every nav, including same-page
+      // ones like opening a side modal, so key says we just landed on a new
+      // location (or this is the initial render) and page decides whether the
+      // nav stayed within the same page
+      if (prev.current?.page === page) {
```
fakemonster: i had a good deal of trouble following the code paths here. mostly to do with reconciling the conditions and the comments around them, or rather, being certain this wasn't overlooking some unnamed conditions. i think we can handle that, add some clarity to the existing cases, _and_ prevent future spaghettification by using a custom type to put explicit names on what the conditions are observing:

```ts
/**
 * The type of navigation that occurred based on the navigation state and how
 * the new location compares to the previous one.
 */
type NavigationKind =
  /** A navigation is in flight; the old page is still on screen. */
  | 'loading'
  /** Landed on a new location, but the same page is underneath, e.g. a side
   *  modal opening/closing. */
  | 'movedToSamePage'
  /** Landed on a new location showing a different page.  */
  | 'movedToNewPage'
  /** Re-rendered at the location we're already on; nothing to do. */
  | 'noChange'

function getNavigationKind(
  state: Navigation['state'],
  key: string,
  page: string,
  prev: PrevLocation | null
): NavigationKind {
  return match(state)
    .with('loading', (): NavigationKind => 'loading')
    .with('submitting', (): NavigationKind => 'noChange')
    .with('idle', (): NavigationKind => {
      // same key means we re-rendered at the location we're already on
      if (prev?.key === key) return 'noChange'
      return prev?.page === page ? 'movedToSamePage' : 'movedToNewPage'
    })
    .exhaustive()
}
```

and then in place of the big `if` block:

```ts
    match(getNavigationKind(state, key, page, prev.current))
      .with('loading', () => setScrollPosition(key, window.scrollY))
      // leave the scroll alone and record it under the new location's key so
      // back/forward to it restores correctly
      .with('movedToSamePage', () => setScrollPosition(key, window.scrollY))
      .with('movedToNewPage', () => window.scrollTo(0, getScrollPosition(key)))
      .with('noChange', () => {})
      .exhaustive()
```

app/hooks/use-scroll-restoration.ts
david-crespo: Oh yeah. You have to open a modal, then go to another page, and then go back to test it. I’ll add it to the tests.

app/hooks/use-scroll-restoration.ts
david-crespo: I took more the spirit than the letter of your suggestion. Yours is very readable and I'd still consider it, but it rubs me the wrong way to essentially run the same match twice. What I took from that was that the problem was that the comments were bad — some too long, some missing — and improved those. I experimented with other arrangements of the if-else, like making it flatter, but they weren't really better. Well, there's also this middle ground, which uses variables to do what you were doing with the match union:

```ts
// idle + new key = we just landed somewhere (or this is the initial render)
const landed = state === 'idle' && prev.current?.key !== key

// underlying page didn't change, which means the nav was opening or
// closing a side modal
const modalNav = prev.current?.page === page

if (state === 'loading') {
  // nav in flight: save current scroll under the outgoing location's key
  setScrollPosition(key, window.scrollY)
} else if (landed) {
  if (modalNav) {
    // leave scroll alone, record under the new key so it can be restored
    // on forward/back
    setScrollPosition(key, window.scrollY)
  } else {
    // landed on a new page: restore its saved scroll, or 0 if none saved
    window.scrollTo(0, getScrollPosition(key))
  }
}
if (state === 'idle') prev.current = { key, page }
```

fakemonster: i like this one better than the one you've got at the moment, i'll take expressive variables over comments any day (i don't have to figure out whether the statement is about the line above or below, or both). but "modalNav" is overspecifying; there are other actions that cause "navigating to the same page", like switching tabs or clicking a self-referential breadcrumb

fakemonster (approved):
i might have quibbles about the specificity of "modals" here but functionally it's great

PR #3337, opened 2026-08-19
Require all final breadcrumbs to not point to their parents

fakemonster:
Inspired by a comment on [Charlie's webhooks PR](https://github.com/oxidecomputer/console/pull/3320/changes#diff-c8f1bef7c0f906ca5002103709d1125057a54691dbbe673d4978670628c61cbeR80). I was curious on whether there were terminal breadcrumbs that didn't abide by this reasoning, and inadvertently pointed to parent pages. And there were!

Breadcrumbs breaking this rule fell into two cases:

1. Pointing to a parent view that defaulted back to "where you just came from". Not a big deal, but a little silly
2. Pointing to a parent view that then defaulted to _some other place._ That's weird!

### DIFF
--- a/app/util/path-builder.spec.ts
+++ b/app/util/path-builder.spec.ts
@@ -186,3 +186,29 @@ test('breadcrumbs', async () => {
 
   expect(Object.fromEntries(pairs)).toMatchSnapshot()
 })
+
+// Some pages don't have self-referential breadcrumbs (like /instances-new). But
+// these pages also aren't _targeted_ by breadcrumbs! Any page that's targeted
+// by a breadcrumb should be able to point to itself.
+test('every page reachable by breadcrumb should have a self-referential breadcrumb', async () => {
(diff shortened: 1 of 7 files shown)


fakemonster:
Terrible commit name

app/util/path-builder.spec.ts
```diff
@@ -186,3 +186,29 @@ test('breadcrumbs', async () => {
...
+
+  for (const path of allPaths) {
+    const crumbs = matchesToCrumbs(await getMatches(path)).filter(
+      ({ titleOnly }) => !titleOnly
+    )
+    const last = R.last(crumbs)
+    if (last === undefined) expect.fail(`Found no breadcrumbs for ${path}`)
+    expect(dropFinalSlash(path)).toEqual(dropFinalSlash(last.path))
+  }
+})
```
david-crespo, line 214: love this kind of test

david-crespo (approved):

PR #3345, opened 2026-08-21
Don't reset submit button loading state right before nav

david-crespo:
Noticed while doing #3344. Form submit button loading state is true only when the mutation `isPending`. But there's a brief moment after the mutation completes but before the success nav happens where that's false and the button is on screen. What that looks like is the button flipping back to clickable right before the nav. That's stupid, so I gave every button where it makes sense `loading={isPending || isSuccess}`. There's some risk we could have a form where that is not appropriate (and there are a couple already — see `createNic` on the instance networking tab) but it's very easy for agents to get it right. I added a note to AGENTS.md about it.

### Before

https://github.com/user-attachments/assets/fc4aadf6-7ba2-4e68-af01-be2a4dda5e2f

### After

https://github.com/user-attachments/assets/7dd7d2de-5f34-48be-821b-58e2bb8f767a

### DIFF
--- a/app/pages/project/instances/NetworkingTab.tsx
+++ b/app/pages/project/instances/NetworkingTab.tsx
@@ -897,7 +897,7 @@ const AttachExternalSubnetModal = ({
       }}
       submitLabel="Attach"
       submitError={externalSubnetAttach.error}
-      loading={externalSubnetAttach.isPending}
+      loading={externalSubnetAttach.isPending || externalSubnetAttach.isSuccess}
       onDismiss={onDismiss}
     >
       <ListboxField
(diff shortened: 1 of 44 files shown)


app/pages/project/instances/NetworkingTab.tsx
```diff
@@ -788,8 +788,14 @@ export default function NetworkingTab() {
 
         {createModalOpen && (
           <CreateNetworkInterfaceForm
-            onDismiss={() => setCreateModalOpen(false)}
+            onDismiss={() => {
+              setCreateModalOpen(false)
+              createNic.reset() // clear stale error state
+            }}
```
david-crespo, line 794: Lack of `reset()` here was a bug, uncaught in a previous PR where I tried to fix these missing resets.

PR #3348, opened 2026-08-24
Metrics explorer page design

benjaminleonard:
WIP


app/components/oxql-autocomplete.ts
```diff
@@ -0,0 +1,144 @@
...
+})
+
+/**
+ * Complete based on which clause the cursor is in, determined with regexes
+ * rather than a real parser: OxQL clauses are short and always start with a
+ * table operation, so "text since the last pipe" is nearly always enough.
+ *
+ * Exported for tests; use {@link oxqlAutocomplete} in the editor.
+ */
+export const oxqlCompletionSource =
```
david-crespo, line 159: WOW lol

app/ui/styles/index.css
```diff
@@ -105,6 +108,19 @@
   margin-right: var(--content-gutter);
 }
 
+/* Inline code chip that takes its tint from the surrounding text color, so it
+   works on any background (e.g., inside an error message) */
+@utility inline-code {
+  font-size: 0.825em;
```
fakemonster: rare good application of `em`, love it

app/pages/system/OxqlPage.tsx
```diff
@@ -492,75 +561,194 @@ function LineChart({
...
       />
     </ChartContainer>
   )
 }
 
-function ChartEntry({ display, trim }: { display: ChartDisplay; trim: boolean }) {
+function ChartEntry({ display }: { display: ChartDisplay }) {
   return (
     <>
-      {display.showDivider ? (
```
fakemonster, line 525: `showDivider` is a dead field now. double-checking that's intentional before i delete it entirely @benjaminleonard

app/pages/system/MetricsExplorer.tsx
```diff
@@ -492,75 +561,194 @@ function LineChart({
...
+  const s =
+    v === null || v === undefined
+      ? ''
+      : typeof v === 'object'
+        ? JSON.stringify(v)
+        : String(v)
+  return /[",\n]/.test(s) ? `"${s.replaceAll('"', '""')}"` : s
+}
+
+const tablesToCsv = (tables: OxqlTable[]): string => {
```
fakemonster: I think CSVs aren't worth the effort. The `fields` data, which is arguably the most important, is pretty useless as a single column; any reasonable CSV would have an independent column for each field instead of concatenating them together. But that isn't always doable in our case, because queries may hit multiple tables and so have different fields across all data. The answer to that is to include a column for each field of each table, and leave the values empty when your row doesn't belong to that table. And given that two tables may share a field name but not a field _meaning,_ those names should be in some form like `Field (Table Name)`. All of which we could do, but I say it all mostly to illustrate that the JSON is much more pliable to an end-user, and it'll be tough for us to form a CSV with much utility without knowing someone's use case for it (as opposed to them generating it from the JSON themselves)

app/pages/system/MetricsExplorer.tsx
```diff
@@ -0,0 +1,984 @@
...
```
fakemonster, line 662: I think CSVs aren't worth the effort. The fields data, which is arguably the most important, is pretty useless as a single column; any reasonable CSV would have an independent column for each field instead of concatenating them together. But that isn't always doable in our case, because queries may hit multiple tables and so have different fields across all data. The answer to that is to include a column for each field of each table, and leave the values empty when your row doesn't belong to that table. And given that two tables may share a field name but not a field meaning, those names should be in some form like Field (Table Name). All of which we could do, but I say it all mostly to illustrate that the JSON is much more pliable to an end-user, and it'll be tough for us to form a CSV with much utility without knowing someone's use case for it (as opposed to them generating it from the JSON themselves)

david-crespo: nuke it

fakemonster: [d991f9d](https://github.com/oxidecomputer/console/pull/3348/commits/d991f9d46c50bbdea780d5b3beace68fedca561b)

PR #3350, opened 2026-08-24
tools: bundle size diff CLI

david-crespo:
No big deal, just something to check the effect of a given PR on the bundled output. Uses `fzf` for the PR picker and [`delta`](https://dandavison.github.io/delta/) for a nice diff.

Like `api-diff`, you can specify either a PR number (in this case a console PR rather than an omicron PR) or a base and a head rev. There are some heuristics in place to reduce diff noise: we have to replace the hashes in the filenames with placeholders and we ignore file size changes under 0.05 KB. We also sort the files by name rather than size (which Vite does by default) to avoid spurious changes in order due to files changing size slightly.

image

image

https://github.com/user-attachments/assets/2aa44d75-3ab5-4442-b9a3-3972a62c7cc3


david-crespo:
Not even a flake but some new horror.

image

PR #3352, opened 2026-08-25
Require npm 12

fakemonster:
Most importantly, this makes allowScripts default to off. npm@12 also requires node 22.22.2 or greater, so we just bump the specificity a bit. Technically >=22.22.2 is less strict than npm's real engine requirements, since it has a minimum version for 24 as well, but we assume usage of node 22 in general.

This also took a little extra footwork for the flake since no package currently offers npm 12, so we have to install it directly.


fakemonster:
Draftifying while I toy around with it

.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,40 @@
...
+
+inputs:
+  node-version:
+    description: Node version to install
+    default: '22'
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/setup-node@v6
```
david-crespo: setup-node is apparently on v7 now

fakemonster: i'm neutral on bumping. the changelog doesn't have any features that would affect what we're doing

fakemonster:
So: the install works, and the caching works. If we want to do this "everywhere" (and anywhere we want to enforce usage of npm@12, we would need this action), we need to do this same kind of change. The options I can think of:

1. we can copy this action to every npm-using repo, making whatever little changes are needed to make it compatible. I don't know the exact number of repos that'd be but I'd place it around "enough to be annoying".
2. we can have a single repo define an action and have other repos point _to_ the action. Perhaps we'd want this to be its own standalone repo since it may need to support more features than any one repo needs from the action, but github actions are perfectly happy being told to pull an action from any arbitrary public repo. Side blessing (or curse): we could standardize certain qualities of node setup in our CI.
3. we can avoid the setup-node issue if we use a different package manager (obligatory pnpm endorsement from me), but then the package manager version enforcement would need to come from corepack, which is annoying in its own right.
4. we can kick the can on enforcing this (in which case I'll throw out the changes other than the flake), and try to get a change merged to setup-node that's compatible with our goal here (there's already an [open issue](https://github.com/actions/setup-node/issues/1553)).

Of course, we could blend 4 with any of the other options.

david-crespo:
It's maybe 5 repos. docs, oxide-computer, rfd-site, careers, one or two more. Not a huge deal? I feel like making the standalone action repo is overkill if they're probably going to fix this in setup-node soon regardless.

david-crespo:
Ok, it doesn't look like they're going to fix that issue and it doesn't look like a Node version will ship with npm 12 as its default until 27, and we usually stick to LTSs, so I think I'm sold on the shared action in a repo.

https://github.com/oxidecomputer/actions

david-crespo:
https://github.com/oxidecomputer/actions/pull/6 makes it use the engines field in the `package.json` as the node version. It installs the newest version satisfying the range, so `>=24.15` doesn't work unless you're ok with 26. I put it on `^24.15` for now.

Edit: cool, it works

image

PR #3355, opened 2026-08-26
Drop jsdom and testing library by switching to vitest browser mode

david-crespo:
I'd been meaning to do this for a while. Nearly all the changes are to the tests.

### 🤖 summary

The unit suite is now pure logic running in node; anything that touches the DOM is a vitest browser mode test. That lets us remove jsdom, `@testing-library/*`, and msw/node entirely, along with the unit setup file, whose three stubs (canvas, matchMedia, ResizeObserver) existed only to fake DOM APIs jsdom lacks — those tests now run against the real things.

The non-mechanical bits:

* The API client tests run against an MSW worker in the browser (the same mechanism the dev server uses) instead of msw/node's patched fetch.
* `TimeSeriesChart` created a canvas at module scope, which made it un-importable without a DOM — this is what the setup file's canvas stub was for. The measuring context is now created lazily.
* The `TimeSeriesChart` spec no longer mocks anything: it renders real uPlot and spies on the instance's `redraw`/`setData` via a test-only `onCreate` prop. An earlier version of this PR kept the jsdom-era `vi.mock` of uplot-react, which flaked on webkit in CI — with a cold dep optimizer cache, the mock can attach to a different copy of the module than the one under test. Instance spying is the fallback because uPlot assigns its methods per instance; there's no prototype to patch. The old data-reference-stability test is reframed behaviorally: rerenders with equivalent data must not call `setData`, changed data must.
* The `loginUrl` unit test only exercised string concatenation, so it's replaced with an e2e test covering the actual 401 → `/login?redirect_uri=...` flow, using a new `error-401` sentinel project in the mock API. With that test gone, nothing mocks `nav-to-login` anymore, and mockability was the module's stated reason for existing, so it's folded into `client.ts`.

Note for #3312: `TimeSeriesChart.spec.tsx` moved to `TimeSeriesChart.browser.spec.tsx` and was rewritten around real uPlot instance spies, so the spec edits there will need porting.


package.json
```diff
@@ -103,7 +100,6 @@
     "eslint-plugin-react-hook-form": "^0.3.1",
     "husky": "^9.1.7",
     "ip-num": "^1.5.1",
-    "jsdom": "^29.1.1",
```
david-crespo, line 106: how could you say no to dropping 4 deps

app/api/client.ts
```diff
@@ -96,6 +95,18 @@ type ExpectedError = {
...
+  return opts.includeCurrent
+    ? // TODO: include query args too?
+      `/login?redirect_uri=${encodeURIComponent(pathname + search)}`
+    : '/login'
+}
+
+export function navToLogin(opts: { includeCurrent: boolean }) {
+  window.location.assign(loginUrl(opts))
+}
+
```
david-crespo, line 109: This was in its own file only so we could mock the module. We no longer mock it in the tests.

app/components/TimeSeriesChart.tsx
```diff
@@ -152,6 +153,13 @@ type TimeSeriesChartProps = {
   yAxisTickFormatter?: (val: number) => string
   hasError?: boolean
   loading: boolean
+  /**
+   * Test-only: exposes the underlying uPlot instance so specs can spy on its
+   * methods. uPlot assigns methods per-instance (no prototype), and module
+   * mocks are unreliable in Vitest browser mode, so this is the only reliable
+   * observation point.
+   */
+  onCreate?: (u: uPlot) => void
```
david-crespo, line 162: Kinda janky but also whatever? Definitely not more janky than mocking uPlot.

PR #3358, opened 2026-09-01
Mobile layout

david-crespo:
#3178 was meant to redo part of #2087 as a stepping stone to a proper mobile layout, but then we didn't follow up, which means the mobile layout actually got worse — it's really bad! This is an attempt at some quick fixes to make it tolerable without breaking anything for desktop users. As long as nothing on desktop sizes is affected, even if there are flaws it's hard for it to be worse than the status quo. I think it feels really good, and a thorough review by Fable 5.1 didn't turn up any regressions at desktop sizes.

## 🤖 Summary

Makes the console usable below 1000px, which is the existing `1000:` breakpoint from the design system. Above that width the layout is unchanged apart from the items called out at the end.

### Navigation

- **Sidebar becomes an overlay on small screens.** A hamburger toggle in the top bar opens it over the content with a scrim. It closes on scrim click and on any navigation, including navigations from breadcrumbs or quick actions. Open state lives in a small zustand store (`app/stores/mobile-nav.ts`) so the top bar and sidebar can share it. `PageSkeleton` renders the sidebar closed so there is no flash of an open drawer during load.
- **Top bar collapses to a single cell.** The home button cell only exists at desktop width. The shared class strings in `app/layouts/helpers.tsx` are used by both `TopBar` and `PageSkeleton` so the two can't drift.
- **Breadcrumbs collapse instead of overflowing.** When the path doesn't fit, leading crumbs are folded into a "…" menu listing the full path, keeping the silo picker and user menu in view. Widths are measured from an invisible copy of the crumbs so the longest complete suffix that fits can be chosen; the current crumb is the last to give way and truncates with an ellipsis as a final fallback. This applies at all widths, so long paths on desktop also collapse rather than pushing the right-side controls off screen.
- **User menu shows the username inside the menu** on small screens, where it is hidden from the trigger to save space.

### Components

- `Modal`, `SideModal`, `Toast`, and `ErrorPage` are capped to the viewport width with a gutter so they no longer run edge to edge or overflow.
- Tab lists scroll horizontally below the breakpoint instead of overflowing the page.
- `CardBlock.Header` actions, `DisksTableField` buttons, `Radio` cards, and `DateTimeRangePicker` stack or stretch to full width on small screens.
- Serial console terminal uses the full width and hides the scroll buttons on small screens; the footer height becomes content-driven.
- Side modal gutter shrinks to 1.5rem below the breakpoint.

### Changes that also affect desktop

- `--content-gutter` shrinks below 1000px rather than 768px, so windows between 768 and 999px get the narrower gutter. This aligns the gutter with the layout breakpoint used everywhere else.
- `overscroll-behavior-y: none` on `body` prevents pull-to-refresh and scroll chaining behind the sidebar overlay. On macOS this also removes the vertical rubber-band bounce.
- `DropdownMenu.Content` defaults `collisionPadding` to 12px (Base UI default is 5px) so menus keep a consistent margin from the viewport edge.

### Tests

- New `test/e2e/mobile-layout.e2e.ts` covers the sidebar overlay open/close/navigate flow, the breakpoint transition at 999/1000px, the username in the user menu, and modal/toast bounds at 320px.
- `test/e2e/breadcrumbs.e2e.ts` adds cases for collapsed breadcrumbs at 404, 650, 750, and 1024px, including the full-path menu.
- The serial console visual regression test is scoped to a role locator because `getByText` now also matches the hidden breadcrumb measurement copy.

### DIFF
--- a/app/components/DocsPopover.tsx
+++ b/app/components/DocsPopover.tsx
@@ -55,7 +55,7 @@ export const DocsPopover = ({ heading, icon, summary, links }: DocsPopoverProps)
       <PopoverPanel
         // popover-panel needed for enter animation
         className="popover-panel bg-raise light:bg-default shadow-menu z-10 w-96 rounded-lg"
-        anchor={{ to: 'bottom end', gap: 12 }}
+        anchor={{ to: 'bottom end', gap: 12, padding: 16 }}
       >
         <div className="px-4">
           <h2 className="text-sans-md mt-4 flex items-center gap-1">
--- a/app/components/ErrorPage.tsx
+++ b/app/components/ErrorPage.tsx
@@ -40,7 +40,7 @@ export function ErrorPage({ children }: Props) {
         </Link>
         <SignOutButton className="mt-4 mr-6" />
       </div>
-      <div className="bg-raise! shadow-modal absolute top-1/2 left-1/2 flex w-96 -translate-x-1/2 -translate-y-1/2 flex-col items-center justify-center space-y-4 rounded-lg p-8">
+      <div className="bg-raise! shadow-modal max-1000:w-[calc(100%-var(--content-gutter)*2)] absolute top-1/2 left-1/2 flex w-96 -translate-x-1/2 -translate-y-1/2 flex-col items-center justify-center space-y-4 rounded-lg p-8">
         <div className="my-2 flex h-12 w-12 items-center justify-center">
           <div className="bg-destructive-inverse absolute h-10 w-10 rounded-full opacity-40 motion-safe:animate-[ping_2s_cubic-bezier(0,0,0.2,1)_infinite]" />
           <Error12Icon className="text-error light:text-error-secondary relative h-8 w-8" />
--- a/app/components/PageSkeleton.tsx
+++ b/app/components/PageSkeleton.tsx
@@ -13,6 +13,7 @@ import {
   ContentPane,
   PageContainer,
   sidebarWrapperClass,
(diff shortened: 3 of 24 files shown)


david-crespo:
At mobile width the breadcrumbs can overflow and make the silo/system picker and user menu inaccessible.

image

Note however that this is not really more broken than the status quo, which lets you scroll sideways but completely borks the content.

image

david-crespo:
Quick fix for that. Ugly but not broken. It's a little unfortunate that the last crumb you're left with is generic "Routes" instead of including both the name of the router _and_ "Routes", but I don't think there's room for both.

Screenshot 2026-09-01 at 10 46 42 AM

david-crespo:
Responsive crumbs, improve profile button collapse behavior, put username inside mobile version of user menu since it disappears from the button.

https://github.com/user-attachments/assets/bed1feae-15af-491d-873c-42d626dfc7d1

david-crespo:
Looking really solid, only a couple of rough spots, probably mergable as-is.

https://github.com/user-attachments/assets/02d035dd-e79a-4c17-a42f-d12727f0f8c6

### Bad button

https://console-3o84wwqme-oxidecomputer.vercel.app/projects/mock-project/vpcs/default/routers/mock-system-router

image

### Can't see disk name in mini-table

The stacking buttons are a little sad but tolerable.

https://console-3o84wwqme-oxidecomputer.vercel.app/projects/mock-project/instances-new

image

PR #3365, opened 2026-09-03
Drop `instanceState` row hack from instance NIC and disk tables

david-crespo:
We had a hack in a couple of spots to prevent certain row actions componenents from unmounting and remounting, which would cause an open row actions menu to close. #3364 makes the hack unnecessary, so we remove it. 4dccaed5b67a0fc1d7d27085147040dd1ddeef4c adds e2e tests for those exact scenarios and shows they pass in CI. But those tests are too specific, so they're not worth keeping around. The next commit reverts them. The last one adds a vitest browser test that is a little more general, showing how an open menu behaves when its underlying state changes (in this case a button going from disabled to enabled).

### DIFF
--- a/app/pages/project/instances/NetworkingTab.tsx
+++ b/app/pages/project/instances/NetworkingTab.tsx
@@ -24,7 +24,6 @@ import {
   type ExternalIp,
   type ExternalSubnet,
   type InstanceNetworkInterface,
-  type InstanceState,
   type IpVersion,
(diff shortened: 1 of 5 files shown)


david-crespo:
Temporarily pushing a test that proves the changes doesn't break anything, but the tests aren't worth keeping so I'm going to revert them and put in a smaller vitest browser test.

Issue #3362, opened 2026-09-02
Audit/update links in /app/util/links.ts


charliepark:
I asked the robot to go through all of the links in /app/util/links.ts. I'm working on the alerting/webhooks branch at the moment, so don't want to take time out to follow up on these suggestions at the moment, but wanted to note this as something to come back to.

```
Checked all 49 unique URLs in app/util/links.ts (the license URL and the OxQL template excluded). Each page was fetched following redirects, and every fragment was checked against the id attributes on the final page.

42 resolve cleanly, including all five new alerts links: the overview, #_globbing, the webhooks guide, #_secrets, #_zero_downtime_webhook_secret_rotation, and the webhook_receiver_create API page. The timeseries-schemas base page for the OxQL template also exists.

7 need attention:

┌───────────────────────────┬─────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────┐
│            Key            │                             Problem                             │                        Suggested target                        │
├───────────────────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
│ docLinks.sleds            │ Page 404s: architecture/service-processors is gone              │ architecture/rack-components#_server_sled (anchor confirmed)   │
├───────────────────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
│ docLinks.deviceTokens     │ Redirects to cli/guides/introduction, which has no              │ guides/user-settings#_access_tokens (anchor exists on the      │
│                           │ _device_token_setup                                             │ user-settings page)                                            │
├───────────────────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
│ docLinks.oxql             │ Redirects to metrics/oxql-tutorial; anchor is                   │ guides/metrics/oxql-tutorial#_oxql_quick_start                 │
│                           │ _oxql_quick_start, not _oxql_quickstart                         │                                                                │
├───────────────────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
│ docLinks.scim             │ Page loads but _saml_authentication_scim_user_provisioning is   │ identity-providers#scim (an id scim exists)                    │
│                           │ missing                                                         │                                                                │
├───────────────────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
[...]

PR #3371, opened 2026-09-08
Update docs links

charliepark:
This updates a few links (both to docs.oxide.computer URLs and to a couple of external URLs). Nothing major.

Also one label change, where the "Metrics" link label now reads **"Utilization"**, as
    A. the page linking to it is now called "Utilization"
and
    B. the docs themselves are now titled "Calculating Utilization"


david-crespo (approved):
Looks great, does it close #3362?

charliepark:
> Looks great, does it close #3362?

Yes! Have now closed #3362.

PR #3373, opened 2026-09-08
Skip eager number updates that cause field rewrites

### LINKED ISSUE #3372: Instance: 'Disk size' clamping to 100 GiB for a 10 GiB image, but allowing 20 GiB if we toggle between images
### What happened

While creating an instance with the **almalinux10-pa-tests** image in Dogfood (which is supposed to have a size of 10GiB), the form won't let us set a Disk Size of 20 GiB and -instead- it tries to clamp it to 100.
However, if we select another image, set the 20 GiB and go back to choosing  **almalinux10-pa-tests** , then the 20 GiB is accepted and the VM boots fine.

[GiB dance](https://drive.google.com/file/d/16wPmvA8hPaq_lnQ-t37i0X51nPVqRlcC/view?usp=drive_link)

### Browser console output

```text
--
```

### Console version

https://github.com/oxidecomputer/console/commits/700ff8bf5ce55e426b69a5c0c67334fe3bad53d8

### What system are you on?

dogfood

### System version

23.0.0-0.ci+git7683e26f5b8

fakemonster:
When we resolved the number field committing in #3282 to not be so lazy, it became not lazy enough! In particular, min and max values would be applied while a user was still typing, which was far too intrusive. The solution is to check if react-aria would cause any changes to the input value on commit, and only force a commit when it's a true no-op on the input string.

Resolves https://github.com/oxidecomputer/console/issues/3372


app/ui/lib/NumberInput.browser.spec.tsx
```diff
@@ -57,26 +57,58 @@ test('fires onChange with NaN when the input is cleared', async () => {
...
+  await expect.element(input).toHaveValue('20')
+})
+
+test('does not step intermediate input while typing', async () => {
+  const screen = await render(<NumberInputHarness step={1} minValue={10} />)
+  const input = screen.getByRole('textbox', { name: 'Test number' })
+
+  await input.click()
+  await userEvent.type(input, '20')
+  await expect.element(input).toHaveValue('20')
```
fakemonster, line 104: pausing here to check that this is 20 is really just to avoid race conditions. locally, this would fail on chromium if i did `userEvent.type(input, '20.1')`. the end result would be "201". completely unreproducible irl

david-crespo (approved):

PR #3376, opened 2026-09-09
Consolidate number field min/max validation and revalidate when bounds change

david-crespo:
This is prep for upgrading React Aria, which will let us turn off clamping in the number input and make `NumberField` responsible for enforcing `min` and `max`. On main, `DiskSizeField` is the only number field that validates bounds for itself; the prefix length field on the external subnet form and every other `NumberField` with a `min` or `max` rely on the clamping. Once clamping is off, those fields need explicit validation.

So, in this PR we:

- **Move min/max validation into `NumberField` and delete `DiskSizeField`.** The validate rule that used to live in `DiskSizeField` now runs for every number field with a `min` or `max`, with the units appended to the message ("Can be at most 32", "Must be at least 10 GiB"). That left `DiskSizeField` as a wrapper setting `units="GiB"`, `required`, and `min={1}`. Its two callers now pass those directly.
- **Add `deps` to `ListboxField`** and use it on the external subnet form so changing the pool re-validates prefix length. Without this, a stale "Can be at most 32" would hang around after switching back to a v6 pool, since by default react-hook-form only re-runs validation on the field that changed.
- **Add `deps` to disk type radio in disk create**, so an existing size error updates (e.g., is cleared if appropriate) when switching between Local and Distributed instead of waiting for the next submit.

This also fixes a bug on main: the clamp is display-only. Entering 64 on a v6 pool and switching to v4 shows 32 in the box, but form state still holds 64 and submit sends `prefix_length: 64`. With this change, submit is blocked with a message instead.

The existing e2e specs for disks, external subnets, instance create, and subnet pools pass unchanged. The external subnet prefix length test gets a submit assertion covering the display-only clamp bug.


fakemonster (commented):
agreed that the faux-clamping behavior was a bug, but now the faux-clamping still happens WITH error messages, which i'm not sure is a huge improvement:

Image

david-crespo:
I would say blocking form submit and adding the bizarre message is at worst a lateral move. But I'll PR the react-aria upgrade that fixes it in a minute anyway.

PR #3382, opened 2026-09-11
Bump API

david-crespo:
Bumping to the version going to dogfood, then will bump again for https://github.com/oxidecomputer/omicron/pull/11281.


app/forms/vpc-create.tsx
```diff
@@ -24,7 +23,16 @@ import { validateVpcIpv6Prefix } from '~/util/ip'
 import { docLinks } from '~/util/links'
 import { pb } from '~/util/path-builder'
 
-const defaultValues: SetNonNullable<Required<VpcCreate>> = {
```
david-crespo, line 27: The API change did not like this because this turns `defaults` into a required field, so I changed this pattern which I am probably responsible for but which I don't really like anymore. defining the type directly is fine. The other files with similar changes didn't have to change but I figured they should match if we don't want to keep encouraging this pattern.

Issue #3317, opened 2026-08-10
Simplify CPU/memory selection on instance create


david-crespo:
As discussed at Aug 10 product eng sync ([transcript](https://docs.google.com/document/d/1ikc2t6SjLmMvv_DiT2YA-NbpLnjmzHLGdOPAGMqNDTU/edit?tab=t.fpp8f2tqazx3)), we want to get rid of the tabs here, instead having a single view with the direct inputs present by default, and then the presets are just buttons next to them that set particular values in the inputs. We want the presets to have ratios of 1 vCPU to 8 GiB RAM, and we want the lowest one to be 1 vCPU instead of 2.

<img width="804" height="332" alt="Image" src="https://github.com/user-attachments/assets/82163c67-486e-49e9-b81d-2a21315f033a" />

<img width="558" height="265" alt="Image" src="https://github.com/user-attachments/assets/16ddabd0-ff7e-4bec-9d67-5de533d4fa73" />

david-crespo:
We should do the ratios right now in case we don't get to the reworked UI.

PR #3384, opened 2026-09-14
Instance create: 1 vCPU to 8 GiB ratio, drop high CPU and high mem presets

david-crespo:
Half of #3317. I'm going to keep that open because the real work is the UI rework. This is just a stopgap in case we don't get to that for R23. I leave the preset UI in place, just changing the ratio in general purpose from 1:4 to 1:8 and dropping the other categories. It felt silly to change them to some other ratio, it's all very arbitrary. I tried other names for General Purpose (e.g., Presets) but they didn't feel right. This is going away soon enough anyway.

image


david-crespo:
Already chatted with @benjaminleonard about what I did with the tabs, so I'm just gonna merge.

